PR #7, opened 2020-03-27
Skeletoned out command sets

MattGal:
Big things to @mandel-macaque for helping me understand the challenges of Mono.options


src/XHarness/Android/AndroidTestCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (ShowHelp)
+            {
+                Options.WriteOptionDescriptions(Console.Out);
+                return 1;
+            }
+            if (extra.Count > 0)
+            {
+                Console.WriteLine($"Unknown arguments: {string.Join(" ", extra)}");
+                Options.WriteOptionDescriptions(Console.Out);
+                return 2;
```
mandel-macaque: I wonder if we can move this to some command function or something similar. Looks like it i a pattern that we are going to do a lot, or at least is several commands.

PR #11, opened 2020-03-30
Fix code style based on new .editorconfig

premun:


src/Microsoft.DotNet.XHarness.CLI/Android/AndroidGetStateCommand.cs
```diff
@@ -10,34 +10,37 @@ namespace Microsoft.DotNet.XHarness.CLI.Android
 {
     public class AndroidGetStateCommand : Command
     {
-        bool ShowHelp = false;
+        private bool _showHelp = false;
```
MattGal: pretty sure this doesn't actually follow the rest of .NET Core repos?

premun: Which part?

The code style is taken from the runtime repo and for private fields `_camelCase` is defined:
https://github.com/dotnet/xharness/blob/dfc0b9a8ec3e7fdfa0062a7b429d29486048677f/.editorconfig#L71-L78

Example file:
https://github.com/dotnet/runtime/blob/master/src/libraries/Microsoft.Extensions.Logging.Console/src/ConsoleLogger.cs

MattGal: yeah you're right.  Ignore me :)

PR #13, opened 2020-04-01
Unify common argument parsing for test/package commands

premun:
Common arguments and their validation is now handled in one place for Android and iOS platforms.

This should help make sure our Android and iOS CLI won't digress much and should help reduce the need to write validation boilerplate twice. It also allows to write async implementation of the commands which was cumbersome to do before.


premun:
Needs to be done for the package command still..

PR #14, opened 2020-04-02
Force LF EOLs in shell scripts

premun:
Repo was initialized with arcade scripts having LF and they cannot be executed on Linux


premun:
Or maybe just my checkout is wrong..

premun:
Well, this will be useful anyway to have to enable running them in docker on windows

PR #26, opened 2020-04-14
Remove explicit permission granting (doesn't seem necessary)

MattGal:


MattGal:
(I'm going to just merge this when it's green, it's just deleting unnecessary code)

PR #27, opened 2020-04-14
Download and unzip ADB from android tools

MattGal:
This should probably be in Microsoft.DotNet.XHarness.Android but I couldn't get it to end up in the Nupkg without putting it in the CLI project

I asked a couple folks, also tagging @rainersigwald  in case he has ideas.  Basically I'd like to have this ADB stuff be in the referenced Android project and not the CLI, and I failed to make that happen so I just stuck it in the CLI project for now.  Is this fixable?

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -16,7 +16,7 @@ namespace Microsoft.DotNet.XHarness.Android
         #region Constructor and state variables
 
         // When packaged as a DotNet CLI tool, this should always be a minimal adb.exe
-        private const string DefaultAdbExePath = @"..\..\..\content\adb.exe";
+        private const string DefaultAdbExePath = @"..\..\..\content\adb\adb.exe";
         private const string AdbEnvironmentVariableName = "ADB_EXE_PATH";
 
         private readonly string _absoluteAdbExePath;
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -15,14 +15,14 @@ namespace Microsoft.DotNet.XHarness.Android
     {
         #region Constructor and state variables
 
-
-
+        // When packaged as a DotNet CLI tool, this should always be a minimal adb.exe
+        private const string DefaultAdbExePath = @"..\..\..\content\adb.exe";
         private const string AdbEnvironmentVariableName = "ADB_EXE_PATH";
 
         private readonly string _absoluteAdbExePath;
         private readonly ILogger _log;
 
-        public AdbRunner(ILogger log, string adbExePath = ".\\adb.exe")
+        public AdbRunner(ILogger log, string adbExePath = DefaultAdbExePath)
         {
             _log = log ?? throw new ArgumentNullException(nameof(log));
 
--- a/src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
+++ b/src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
@@ -1,6 +1,27 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <!-- When updating, avoid using 'latest' url as this can make the same commit build differently on different days -->
+    <WindowsAndroidSdkUrl>https://dl.google.com/android/repository/platform-tools_r29.0.6-windows.zip</WindowsAndroidSdkUrl>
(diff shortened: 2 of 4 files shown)


src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
```diff
@@ -1,5 +1,26 @@
...
+    <!-- When updating, avoid using 'latest' url as this can make the same commit build differently on different days -->
+    <WindowsAndroidSdkUrl>https://dl.google.com/android/repository/platform-tools_r29.0.6-windows.zip</WindowsAndroidSdkUrl>
+    <WindowsAndroidSdkFileName>windows-android-platform-tools.zip</WindowsAndroidSdkFileName>
+  
+    <!-- TODO: Figure out minimal set of binaries for linux, then copy them.  Until then this is unsued. -->
+    <LinuxAndroidSdkUrl>https://dl.google.com/android/repository/platform-tools_r29.0.6-linux.zip</LinuxAndroidSdkUrl>
+    <LinuxAndroidSdkFileName>linux-android-platform-tools.zip</LinuxAndroidSdkFileName>
+  </PropertyGroup>
+
+  <Target Name="DownloadAndroidSdks" BeforeTargets="Build" >
```
premun, line 13: Just asking because I don't know MSBuild much - will this download on every Build whenever I want to run in VS or how does it know it shouldn't run this target?

Or should there be a condition `if adb.exe exists`?

src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
```diff
@@ -1,5 +1,26 @@
...
+    <Unzip SourceFiles="$(IntermediateOutputPath)/android-tools/$(WindowsAndroidSdkFileName)" DestinationFolder="$(IntermediateOutputPath)/android-tools-unzipped/windows" OverwriteReadOnlyFiles="true" />
+    <ItemGroup>
+      <WindowsAdbFiles Include="$(IntermediateOutputPath)/android-tools-unzipped/windows/platform-tools/adb*" />
+      <Content Include="@(WindowsAdbFiles)">
+        <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      </Content>
+    </ItemGroup>
+  </Target>
+
   <PropertyGroup>
```
premun, line 24: nit: The `PropertyGroup` with target framework, assembly name and so on would be maybe nice to keep on top of the file

src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
MattGal: I can make it optional, good point.  It's only 7 MB but I wasn't really thinking about the local build experience especially if someone had slow internet :)

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -15,14 +15,14 @@ public class AdbRunner
     {
         #region Constructor and state variables
 
-
-
+        // When packaged as a DotNet CLI tool, this should always be a minimal adb.exe
+        private const string DefaultAdbExePath = @"..\..\..\content\adb.exe";
```
premun, line 19: I think you need `..\..\..\content\adb\adb.exe`

MattGal: yup

premun: Also would be great to use `Path.Join` so that we can support the unix world

MattGal:
OK I think it's ready to merge now but will make sure builds succeed before clicking merge.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
MattGal: will add this when I add linux adb.

MattGal: (also for most things you can just use forward slashes in Windows and it figures it out :) )

premun:
I will merge this as I was on top of your branch so that I can open my PR

MattGal:
I was actually holding off on this to try to figure out if I could suppress (I can't even repro it locally) the error / warning:

```
 error NU5100: The assembly 'content\adb\AdbWinApi.dll' is not inside the 'lib' folder and hence it won't be added as a reference when the package is installed into a project. Move it into the 'lib' folder if it needs to be referenced.
```
... but I can work on that today as it doesn't seem to be anything more than an annoyance.

premun:
Oh sorry! I saw you suppressed it for Debug builds, so I thought you're good based on your comment.

PR #31, opened 2020-04-15
[iOS] Migrate code from xamarin/xamarin-macios to the repo.

mandel-macaque:
The following code use to live under
https://github.com/xamarin/xamarin-macios but has moved here. The code
is what allows to:

* Create a facade applicaton for testing dlls.
* Execute tests in simulator/devices.

This code has been worked on for years and is been tested. There are a
number of unit tests that are currently in xamarin-macios that will be
moved in a different PR to ensure that the tests work correctly in the
new xharness CI. There have been just two cosmetic changes:

* Code formatting: Adapted to the dotnet core style.
* OSS header added.


src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
```diff
@@ -0,0 +1,625 @@
...
+                else if (line.Contains("[FAIL]"))
+                {
+                    Console.WriteLine(line);
+                    failed = true;
+                }
+            }
+            return (ResultLine: resultLine, Failed: failed);
+        }
+
+        async Task<(string resultLine, bool failed, bool crashed)> ParseResultFile(AppBundleInformation appInfo, string test_log_path, bool timed_out)
```
akoeplinger, line 320: can you provide a way to turn off the xml parsing magic? I don't think we want to use it in Helix.

src/Microsoft.DotNet.XHarness.iOS.Shared/CrashSnapshotReporter.cs
```diff
@@ -0,0 +1,202 @@
...
+                    // (if we put them in /tmp, they'd never be deleted).
+                    crashReports = newCrashFiles
+                        .Select(async crash => await ProcessCrash(crash))
+                        .Select(t => t.Result)
+                        .Where(c => c != null);
+                }
+
+                foreach (var cp in crashReports)
+                {
+                    WrenchLog.WriteLine("AddFile: {0}", cp.FullPath);
```
akoeplinger, line 113: might be a good idea to get rid of all the references to Wrench, it moved to the CI graveyard a few years ago 😄

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
mandel-macaque: Is not used unless specified by the apprunner, but yes, we will provide a way not to parse it. The apprunner that @premun will do will not call it.

src/Microsoft.DotNet.XHarness.iOS.Shared/CrashSnapshotReporter.cs
mandel-macaque: trust me, I want to, I need to find out how :/

premun:
Please provide commit SHA from which you copied the code from in the description and then in the merged commit message. I somehow feel it might be useful one day :)

src/Microsoft.DotNet.XHarness.iOS.Shared/CrashSnapshotReporter.cs
premun: There's not many and from what I have seen the logging is turned off there as well, so we could be safe to remove it. Can happen later

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
premun: So we just copy the outputs back and we're done?

mandel-macaque: @premun yes, in the apprunner, just do not add the generation of the html, afaik is done there in xamarin-macios, then vsts/azure will pick the resulting xml. Makes the runner easier :)

Issue #5, opened 2020-03-27
Migrate the Microsoft.DotNet.XHarness.iOS.Tests project to XUnit


premun:
Currently is NUnit, we might want to move to XUnit

PR #32, opened 2020-04-15
[iOS] Migrate tests from xamarin-macios to the repo.

mandel-macaque:
Bring all tests but the ProcessManager tests to the repo. Only the
following cosmetic tests have been done:

* Follow the code style of dotnet.
* Add the OSS header.


mandel-macaque:
Added as a draft PR to simplify the code review.

mandel-macaque:
@MattGal did I add the tests in the expected location?

premun:
@mandel-macaque test called `FormatArgumentsTest` will fail on Windows and will break the pipeline. I am not aware of any out-of-the-box solution from arcade that would enable you to skip based on platform but maybe if it's one test only, we can also mark it as skipped?

mandel-macaque:
@premun yes, make sense, is not nice, but a valid point.

premun:
@MattGal just FYI the tests are using NUnit: https://github.com/dotnet/xharness/issues/5

Maybe that is why there's no test results found? I see them run and it does find some XML.

I have started the migration of these to XUnit and am almost finished, so I can open PR to this branch then..

premun:
I am finished. Maybe we want to merge the migration to XUnit first to this branch?

https://github.com/dotnet/xharness/pull/48

MattGal:
> @MattGal just FYI the tests are using NUnit: #5
> 
> Maybe that is why there's no test results found? I see them run and it does find some XML...

Yes, I was aware of this;  I just made enough changes to get the tests to try to report to AzDO, they should actually automatically get run once they're xunit.

tests/Microsoft.DotNet.XHarness.iOS.Shared.Tests/CrashSnapshotReporterTests.cs
```diff
@@ -0,0 +1,137 @@
...
+using Moq;
+using NUnit.Framework;
+using Microsoft.DotNet.XHarness.iOS.Shared.Execution;
+using Microsoft.DotNet.XHarness.iOS.Shared.Execution.Mlaunch;
+using Microsoft.DotNet.XHarness.iOS.Shared.Logging;
+using Microsoft.DotNet.XHarness.iOS.Shared.Utilities;
+
+namespace Microsoft.DotNet.XHarness.iOS.Shared.Tests
+{
+    [TestFixture]
```
MattGal, line 19: am confused, isn't this supposed to be converted to Xunit now?

premun: When we close #48 it will

PR #33, opened 2020-04-16
Include mlaunch in the CLI NuGet package

premun:
Adds MSBuild targets that download the `mlaunch` tool from the `xamarin/macios-binaries` repo and includes them in the resulting NuGet package.

The binaries repo is cached in temp dir and mlaunch is downloaded only when the desired version of mlaunch changes.

Both Windows and Unix builds are supported by the csproj target.

Unnecessary files are trimmed from mlaunch (mdb, pdb, MacOS..) so the whole CLI NuGet with both ADB and mlaunch is now 5.9 MB.


src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
```diff
@@ -8,6 +8,22 @@
     <ToolCommandName>xharness</ToolCommandName>
     <!-- Mono.Options is apparently not strong-name signed -->
     <NoWarn>CS8002;</NoWarn>
+
+    <IsWindows Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'true'">true</IsWindows>
```
MattGal, line 12: Suggestion: could more simply use the `$(OS)` property here or `[MSBuild]::IsOSPlatform()`

src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
```diff
@@ -8,6 +8,22 @@
...
+    <IsOSX Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::OSX)))' == 'true'">true</IsOSX>
+    <IsLinux Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))' == 'true'">true</IsLinux>
+
+    <!-- This is a git revision of the xamarin/macios-binaries repo where we get the mlaunch tool from
+         If you change the version, please run /eng/download-mlaunch.sh to get this version into your environment -->
+    <MlaunchVersion>3fbdcdb97459ca2c699d47e33028a106b95a7f1f</MlaunchVersion>
+
+    <!-- When updating, avoid using 'latest' url as this can make the same commit build differently on different days -->
+    <WindowsAndroidSdkUrl>https://dl.google.com/android/repository/platform-tools_r29.0.6-windows.zip</WindowsAndroidSdkUrl>
```
MattGal, line 21: confused, isn't this already added?

src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
```diff
@@ -46,5 +52,17 @@
       </Content>
     </ItemGroup>
   </Target>
-  
+
+  <Target Name="DownloadMlaunch" BeforeTargets="Build;Pack">
+    <Exec Condition="'$(IsWindows)' != 'true'" Command="chmod 0755 ../../eng/download-mlaunch.sh &amp;&amp; ../../eng/download-mlaunch.sh --commit $(MlaunchVersion) --target-dir $(IntermediateOutputPath)mlaunch" />
```
MattGal, line 57: I think you could just get rid of the OS logic above and do `Condition=" '$(OS)' == 'Windows_NT' "` and its opposite.

src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
premun: I was resolving conflicts when you moved this part around and just left all configuration together in one place. I can put property groups next to their respective targets that use them but somehow this felt better?

MattGal: seems fine to me.

premun:
@MattGal I had a really long fight here with PowerShell and git where git sends stuff to stderr even though it succeeds and PowerShell then fails. So just in case you are calling git from PS, there's an env variable that makes git use stdout. It didn't fail on my local, just in AzDO. I guess different git.. Really nasty...

Anyway.. can you please check the build failures here?
1. I do the same like you did with ADB and put mlaunch in `obj/` (this intermediary variable)
2. Microbuild is trying to sign everything and fails

Things I don't know:
1. Are we supposed to sing in `obj/`? Shouldn't we sign in `bin/`? Or this is the normal way before things are copied to `bin`/`publish`/`package` or wherever?
2. Can we make Microbuild sign everything but the `mlaunch` dir? I had a quick look at the YAML but didn't see any singning configuration. I guess it's deeper in arcade. If you don't have experience with this, I will investigate. If you have done this before, please share an advice

MattGal:
> @MattGal I had a really long fight here with PowerShell and git where git sends stuff to stderr even though it succeeds and PowerShell then fails. So just in case you are calling git from PS, there's an env variable that makes git use stdout. It didn't fail on my local, just in AzDO.. Really nasty...
> 
> Anyway.. can you please check the build failures here?
> 
> 1. I do the same like you did with ADB and put mlaunch in `obj/` (this intermediary variable)
> 2. Microbuild is trying to sign everything and fails
> 
> Can we make Microbuild sign everything but the `mlaunch` dir or should I put it somewhere else and reference from there? I had a quick look at the YAML but didn't see any signing configuration. I guess it's somehwere deeper in arcade. If you don't have experience with this, I will investigate. If you have done this before, please share an advice

Discussed on IM, basically just need entries for any managed binaries we don't own like this: https://github.com/dotnet/runtime/blob/b93829cf38f079ccb895708bb7e938ab9541bcca/eng/Signing.props#L27 to leave their signing alone.  ADB doesn't have this problem I think because it's entirely native

EgorBo:
Is there an ETA for this to land?

premun:
@EgorBo trying to land it right now - I only need to ignore mlaunch from signing, run some tests and we're good to go.

premun:
@EgorBo looks like the CI is really slow tonight (500 builds in queue) so might not make it today.. Apologies.

EgorBo:
> @EgorBo looks like the CI is really slow tonight (500 builds in queue) so might not make it today.. Apologies.

No worries 🙂

eng/Signing.props
```diff
@@ -0,0 +1,24 @@
...
+    <FileSignInfo Include="mlaunch.exe" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="System.Net.Http.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="System.Numerics.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="System.Xml.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="Xamarin.Hosting.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="Xamarin.Mac.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="Mono.Security.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="mscorlib.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="System.Core.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="System.dll" CertificateName="3PartySHA2" />
```
premun, line 13: @MattGal I didn't find any other to list files using some path pattern matching. I checked [other dotnet repos](https://github.com/search?q=org%3Adotnet+filename%3ASigning.props&type=Code) and it seemed they all also do this. I tried to do `<ItemsToSign Remove=...` as well but no luck there either.

However, I am not sure it is 100% safe to have `System.dll` here..

PR #34, opened 2020-04-16
Do not build Release for PRs

premun:
I was thinking we could only do Debug for PRs and in Debug skip things like downloading mlaunch and ADB which we don't really need?

@MattGal I don't have much experience in whether it will become hard for us to maintain green internal pipeline if we do what I propose here but maybe it is the right to do will the decreased capacity we are seeing these days? Let me know what you think.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,7 +45,7 @@ stages:
       enablePublishUsingPipelines: true
       enablePublishBuildAssets: true
       helixRepo: dotnet/xharness
-      
+
       jobs:
       - job: Windows_NT
         pool:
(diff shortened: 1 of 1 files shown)


premun:
On another look - mlaunch on my local takes 4 minutes to download - that was my main motivation - but in AzDO it took 5 seconds so I think we might be good with everything everywhere.

MattGal:
I'd prefer we not do this, it is indeed possible to have conditional logic that only runs on release builds and while uncommon, could be broken only for the release flavor.

premun:
Alright, let's not do this then

PR #35, opened 2020-04-16
Fix argument validation

premun:
There were several bugs in the mess that I caused when I wrote the original validation where we for example ignored the bool coming out of the base validation.

This quick fix should make things more simple before we start implementing the whole iOS part.


premun:
I will merge, even though we cannot build on Windows, MacOS should be enough for this kind of change

Issue #39, opened 2020-04-16
[Runners] Provide Android application entry point implementation.


mandel-macaque:
Provide the required implementation for the android entry point. Code should allow callers to:

* Provided a text writer to be used as a logger.
* Specify a location used to store the results.

PR #40, opened 2020-04-16
[Runners] Implement the android application entry point.

mandel-macaque:
Add the android implementation (much simpler than the iOS one). Clients
just need to provide a text writer for the logs and the final location
where tests should be written.

fixes: https://github.com/dotnet/xharness/issues/39


src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
```diff
@@ -0,0 +1,78 @@
+﻿
+// Licensed to the .NET Foundation under one or more agreements.
```
akoeplinger, line 2: remove extra newline above

src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
```diff
@@ -0,0 +1,78 @@
...
+            var logger = (Logger == null || options.EnableXml) ? new LogWriter(Device) : new LogWriter(Device, Logger);
+            logger.MinimumLogLevel = MinimumLogLevel.Info;
+            var testAssemblies = GetTestAssemblies();
+
+            var runner = await CreateRunner(logger);
+
+            // if we have ignore files, ignore those tests
+            await runner.Run(testAssemblies).ConfigureAwait(false);
+
+            TestRunner.Jargon jargon = Core.TestRunner.Jargon.NUnitV3;
```
akoeplinger, line 45: we definitely don't want to use NUnit xml on Helix but xunit

mandel-macaque: @akoeplinger you can pass the jargon to use in the app arguments. One question, does the helix test uploaded take the attachments in xunit xml? That might mean I have to add some code for the xml generation, not hard but we should check it

MattGal: Helix is fine with XUnit or NUnit to report to Azure Devops.  For Kusto tracking (legacy) it's got to be XUnit.

src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
```diff
@@ -0,0 +1,77 @@
...
+    public abstract class AndroidApplicationEntryPoint : ApplicationEntryPoint
+    {
+        /// <summary>
+        /// Implementators should provide a text writter than will be used to
+        /// write the logging of the test that are executed.
+        /// </summary>
+        public abstract TextWriter Logger { get; }
+
+        /// <summary>
+        /// Implementators should provide a full path in which the final
```
MattGal, line 22: nit: Implementors

src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Threading.Tasks;
+using Microsoft.DotNet.XHarness.Tests.Runners.Core;
+
+namespace Microsoft.DotNet.XHarness.Tests.Runners
+{
+    public abstract class AndroidApplicationEntryPoint : ApplicationEntryPoint
+    {
+        /// <summary>
+        /// Implementators should provide a text writter than will be used to
+        /// write the logging of the test that are executed.
```
MattGal, line 17: nit: tests

PR #41, opened 2020-04-16
Move ADB.exe to try to suppress NU5100

MattGal:
Should fix build errors.  I locally verified that the lookup path behavior does not work on Windows too. 

Was meant to be part of https://github.com/dotnet/xharness/pull/27

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.InteropServices;
 using System.Text;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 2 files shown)


MattGal:
going to merge here as the queue backup struggle is real, and the build error is gone on the OSX leg.

premun:
@MattGal I found in the docs that the `runtimes` folder is for multi-targetting and you put there architecture-specific implementations that are then used on runtime instead of your regular implementation.

I also found there is some `tools` folder [apparently meant for content files and scripts](https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks#content-files-and-powershell-scripts). Maybe we want that one?

Or you have found some other source that says that content should go to `runtimes`?

MattGal:
> @MattGal I found in the docs that the `runtimes` folder is for multi-targetting and you put there architecture-specific implementations that are then used on runtime instead of your regular implementation.
> 
> I also found there is some `tools` folder [apparently meant for content files and scripts](https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks#content-files-and-powershell-scripts). Maybe we want that one?
> 
> Or you have found some other source that says that content should go to `runtimes`?
> 
> > going to merge here as the queue backup struggle is real, and the build error is gone on the OSX leg.
> 
> The mlaunch part is not checked in yet so there wouldn't be any error.

Nope that was just a suggestion from a NuGet SME I chatted with, can move it after discussing.

PR #43, opened 2020-04-17
[iOS] Add app un/installer and runner and call them from the CLI test command

premun:
Adds the iOS project and tests for it

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Common/LogLevelExtensions.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Common/LogLevelExtensions.cs
@@ -1,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DotNet.XHarness.CLI.Common
+{
+    internal static class LogLevelExtensions
+    {
+        public static int ToInt(this LogLevel level) => level switch
+        {
+            LogLevel.Trace => 6,
+            LogLevel.Debug => 5,
+            LogLevel.Information => 4,
+            LogLevel.Warning => 3,
+            LogLevel.Error => 2,
+            LogLevel.Critical => 1,
+            LogLevel.None => 0,
+            _ => throw new ArgumentOutOfRangeException(nameof(level)),
+        };
+    }
+}
--- a/src/Microsoft.DotNet.XHarness.CLI/iOS/iOSTestCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/iOS/iOSTestCommand.cs
@@ -171,9 +171,9 @@ namespace Microsoft.DotNet.XHarness.CLI.iOS
                     {
                         _log.LogError($"Failed to uninstall the app bundle with exit code: {uninstallResult.ExitCode}");
                     }
-                }
 
-                _log.LogInformation($"Application '{appBundleInfo.AppName}' was uninstalled successfully");
+                    _log.LogInformation($"Application '{appBundleInfo.AppName}' was uninstalled successfully");
+                }
             }
         }
     }
--- a/src/Microsoft.DotNet.XHarness.CLI/iOS/iOSTestCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/iOS/iOSTestCommand.cs
@@ -46,7 +46,7 @@ namespace Microsoft.DotNet.XHarness.CLI.iOS
             }
         }
 
-        protected override async Task<int> InvokeInternal()
+        protected override async Task<ExitCode> InvokeInternal()
         {
             var processManager = new ProcessManager(_arguments.XcodeRoot, _arguments.MlaunchPath);
             var deviceLoader = new HardwareDeviceLoader(processManager);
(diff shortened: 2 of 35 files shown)


src/Microsoft.DotNet.XHarness.CLI/Common/LogLevelExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DotNet.XHarness.CLI.Common
+{
+    internal static class LogLevelExtensions
+    {
+        public static int ToInt(this LogLevel level) => level switch
+        {
+            LogLevel.Trace => 6,
```
premun, line 14: @mandel-macaque this corresponds to how many `-v` we send to mlaunch

src/Microsoft.DotNet.XHarness.CLI/iOS/iOSTestCommand.cs
```diff
@@ -19,10 +29,15 @@ internal class iOSTestCommand : TestCommand
 
         public iOSTestCommand() : base()
         {
-            Options = new OptionSet() {
+            Options = new OptionSet()
+            {
                 "usage: ios test [OPTIONS]",
                 "",
                 "Packaging command that will create a iOS/tvOS/watchOS or macOS application that can be used to run NUnit or XUnit-based test dlls",
+                { "xcode=", "Path where Xcode is installed", v => _arguments.XcodeRoot = v},
```
premun, line 37: @mandel-macaque some of these could share an interface between the package and test commands.

premun: Also, not sure, but I probably changed the name from xcode-root and mlaunch-path. We can revisit that as well. But should be similar to whatever Android has. Or maybe later we could just revisit everything once we are stabilized so that we are unified across the CLI

mandel-macaque: agreed, lets land as it is, since the API is not closed yet and this is a preview, we are allowed to change it.

src/Microsoft.DotNet.XHarness.CLI/iOS/iOSTestCommandArguments.cs
```diff
@@ -4,36 +4,96 @@
...
+using Microsoft.DotNet.XHarness.iOS.Shared.Utilities;
 
 namespace Microsoft.DotNet.XHarness.CLI.iOS
 {
     internal class iOSTestCommandArguments : TestCommandArguments
     {
+        /// <summary>
+        /// Path to where Xcode is located.
+        /// </summary>
+        public string XcodeRoot { get; set; }
```
mandel-macaque, line 20: We should add the same we added in xamarin-macions, a lazy version, if the path is not provided, use xcode-select, but I'll create an issue for this and we can move fwd.

src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformationParser.cs
```diff
@@ -2,20 +2,35 @@
...
 
     public class AppBundleInformationParser : IAppBundleInformationParser
     {
+        private const string PlistBuddyPath = "/usr/libexec/PlistBuddy";
+
+        private readonly IProcessManager _processManager;
+
+        public AppBundleInformationParser(IProcessManager processManager)
+        {
+            _processManager = processManager ?? throw new System.ArgumentNullException(nameof(processManager));
```
mandel-macaque, line 31: We already have using System. So no new for new System.ArgumentException.

src/Microsoft.DotNet.XHarness.CLI/iOS/iOSTestCommandArguments.cs
premun: This already happens inside `ProcessManager` I tihnk. Should we move it out of there?

mandel-macaque: yes, I created an issue so that we do not forget: https://github.com/dotnet/xharness/issues/49

Feel free to state it there, then I think we only need to change the mono options to use ':'

src/Microsoft.DotNet.XHarness.iOS/AppRunner.cs
```diff
@@ -0,0 +1,359 @@
...
+            _snapshotReporterFactory = snapshotReporterFactory ?? throw new ArgumentNullException(nameof(snapshotReporterFactory));
+            _captureLogFactory = captureLogFactory ?? throw new ArgumentNullException(nameof(captureLogFactory));
+            _deviceLogCapturerFactory = deviceLogCapturerFactory ?? throw new ArgumentNullException(nameof(deviceLogCapturerFactory));
+            _testReporterFactory = reporterFactory ?? throw new ArgumentNullException(nameof(_testReporterFactory));
+            _mainLog = mainLog ?? throw new ArgumentNullException(nameof(mainLog));
+            _logs = logs ?? throw new ArgumentNullException(nameof(logs));
+            _helpers = helpers ?? throw new ArgumentNullException(nameof(helpers));
+        }
+
+        public async Task<(string deviceName, int exitCode)> RunApp(
```
MattGal, line 55: Suggestion:  This one function is 302 lines long.  Seems like some of it could be factored into helper functions for readability (and perhaps in the future, test-ability)

premun: I totally agree. This code is not my invention unfortunately and was migrated from Xamarin:
https://github.com/xamarin/xamarin-macios/blob/master/tests/xharness/AppRunner.cs

I didn't change it in this iteration because I have unit tests that I have it running against and was making sure I didn't break stuff.

I will definitely split this later into device/simulator part, if there's time..

MattGal: Makes sense, thanks for explaining.

Issue #36, opened 2020-04-16
[Runners] Create events to notify that the test execution has been completed.


mandel-macaque:
Add events in the runners to notify the callers that the tests started and completed so that they can react to them. This follows the comment here: https://github.com/dotnet/runtime/pull/34976#discussion_r409418192

PR #46, opened 2020-04-17
[Runners] Add events to notify that a test run started and completed.

mandel-macaque:
Add two events that notify when the runner started and when it
completed. The completed event will provide information of the results
of the run.

Later we can look at adding event for each of the tests starting and
completing, which is possible, but we are not going to add events that
we have not been asked to add. KISS

fixes: https://github.com/dotnet/xharness/issues/36


EgorBo:
Can this be merged today?

Issue #50, opened 2020-04-20
[Runners] Provide events per test start and test end.


mandel-macaque:
Provide an event in the runners that will notify client. 

* Start event: Should contain the full name of the test that started.
* Completed event: Should contain the full name of the test that completed and the result.

PR #51, opened 2020-04-20
[Runners] Provide events to notify when a test has started or completed.

mandel-macaque:
Add event to the base class and each of the runner implementations will
call it using the provided protected methods.

Expose the events via the ApplicationEntryPoint base class that will
connect to the runner and fwd the events.

fixes: https://github.com/dotnet/xharness/issues/50

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitTestRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitTestRunner.cs
@@ -262,6 +262,10 @@ namespace Microsoft.DotNet.XHarness.Tests.Runners.Xunit
             });
             OnInfo($"\t[FAIL] {args.Message.Test.TestCase.DisplayName}");
             OnInfo(sb.ToString());
+            OnTestCompleted((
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitTestRunner.cs
```diff
@@ -153,6 +158,11 @@ void HandleTestPassed(MessageHandlerArgs<ITestPassed> args)
             LogTestDetails(args.Message.Test, log: OnDebug);
             LogTestOutput(args.Message, log: OnDiagnostic);
             ReportTestCases("   Associated", args.Message.TestCases, log: OnDiagnostic);
+            // notify the completion of the test
+            OnTestCompleted((
+				TestName: args.Message.Test.DisplayName,
+                TestResult: TestResult.Passed
+            ));
```
EgorBo, line 165: I don't see where it's fired with `TestResult.Failed`

mandel-macaque: Good catch.

PR #56, opened 2020-04-21
Fix mlaunch path in NuGet when mlaunch is not from cache

premun:
There are two places where the files are copied from the checked out repo and one was copying the whole folder instead of files inside.


premun:
I am going to merge this since I tested it locally

Issue #55, opened 2020-04-21
[iOS] Do not generate the human readable html by default.


mandel-macaque:
The parsing of the xml to a human readable version is done for the xamarin/xamarin-macios team. This does not make any sense in the terms of the xharness tool as a CLI, set it to not be generated by default and let the xamarin/xamarin-macios set it.

premun:
Ah, sorry, I forgot to create this one

mandel-macaque:
@premun I'm on it, no worries :)

PR #61, opened 2020-04-21
[iOS] Do not generate the html report unless it is requested.

mandel-macaque:
Do not generate the hml per default. This is a feature for
xamarin-macios and has to be explicitly enabled.

fixes: https://github.com/dotnet/xharness/issues/55


src/Microsoft.DotNet.XHarness.iOS.Shared/IResultParser.cs
```diff
@@ -32,7 +32,7 @@ public interface IResultParser
 
         // parses the xml of the given jargon, create a human readable result and returns a result line with the summary of what was
         // parsed.
-        (string resultLine, bool failed) GenerateHumanReadableResults(string source, string destination, XmlResultJargon xmlType);
+        (string resultLine, bool failed) GenerateHumanReadableResults(string source, string destination, XmlResultJargon xmlType, bool generateHumanResults);
```
premun, line 35: I think this API is a bit confusing - I don't want to GenerateHumanReadableResults, so I just don't call this? 
Or why would I call it and then set false to the boolean? What would I expect to happen?

mandel-macaque: just pass false to the boolean and it will not be generated, which is the default, maybe changing the name is a good idea

premun: What I am trying to say is:
if I am user of our library and I see `GenerateHumanReadableResults` I would just not call it at all. But that will cause problems, right?
I wouldn't call `GenerateHumanReadableResults(generateHumanResults: false)`

Also I was thinking this bool could be on a class level passed to the ctor as a behavior of the whole runner

premun:
I have changed 
```
(string resultLine, bool failed) GenerateHumanReadableResults(
    string source, string destination, XmlResultJargon xmlType, bool generateHumanResults);
```
to
```
(string resultLine, bool failed) ParseResults(
    string source, XmlResultJargon xmlType, string humanReadableReportDestination = null);
```

I will test now to catch the NRE

premun:
@mandel-macaque I will merge this one as the other bug fixes will be on top of this

PR #62, opened 2020-04-22
Add WebAssembly test runner.

vargaz:
From:
https://github.com/mono/mono/tree/master/sdks/wasm/xunit-runner


mandel-macaque:
@vargaz @marek-safar so if I understand correctly, this is the runner used in wasm, and we want to move it. I think that we can move the cmd line parsing to the main entry point of the CLI, add a new verb 'wasm' and then that way we can reuse the work that @MattGal is doing with the arcade work.

Ideally with that approach, we have:

* iOS: xharness ios test
* Android: xharness android test
* Wasm: xharness wasm test

Each of the verbs take their own specific extra agrs and do the appropriate magic. 

Regarding the runner, I think there is some small overlap between the provided code and the current one, but I think that should be an after though. We have to think how feasible is to share code in that space and I'm sure that @vargaz has A LOT more knowledge about that than me.

MaximLipnin:
Since I wasn't able to push to Zoltan's branch, I opened https://github.com/dotnet/xharness/pull/171

Issue #58, opened 2020-04-21
Stop running tests in official builds


akoeplinger:
I noticed that you're running tests in the official builds on https://dev.azure.com/dnceng/internal/_build?definitionId=795&_a=summary, that doesn't happen in other official builds and causes the official build to take longer than necessary.

@premun @MattGal

premun:
Thanks for letting us know. I will turn them of.

PR #63, opened 2020-04-22
Stop running tests in official builds

premun:
We shouldn't be running unit tests in official builds.

I created 4 scripts for clarity (PR/internalbuild.sh/cmd) and in case we would like to customize more in the future.

Fixes: https://github.com/dotnet/xharness/issues/58


azure-pipelines.yml
```diff
@@ -63,23 +63,33 @@ stages:
...
-        - task: PublishTestResults@2
-          displayName: 'Publish Unit Test Results'
-          inputs:
-            testResultsFormat: xUnit
-            testResultsFiles: '$(Build.SourcesDirectory)/artifacts/TestResults/**/*.xml'
-            mergeTestResults: true
-            searchFolder: $(system.defaultworkingdirectory)
-            testRunTitle: XHarness unit tests - $(Agent.JobName)
-          condition: succeededOrFailed()
+        - ${{ if eq(variables._RunAsPublic, True) }}:
```
MattGal, line 83: This doesn't need to be conditional, there just won't be test results if you don't make them.

azure-pipelines.yml
```diff
@@ -63,23 +63,33 @@ stages:
...
                 _BuildConfig: Debug
         steps:
-        - script: eng\common\cibuild.cmd
-            -configuration $(_BuildConfig)
-            -prepareMachine
-            $(_InternalBuildArgs)
-          name: Build
-          displayName: Build
-          condition: succeeded()
+        - ${{ if eq(variables._RunAsPublic, False) }}:
```
MattGal, line 66: I strongly prefer not having duplicate-except-for-running-tests scripts.  Why not just change the value of a variable like $(_InternalBuildArgs) conditionally?

azure-pipelines.yml
premun: It gives warnings though:
image

Not sure we want those on every build, so I added these guards

MattGal: sure that's fine.  Less pressing than the "don't duplicate the whole script to remove 1 arg" issue :)

azure-pipelines.yml
premun: Good point, will do

premun:
>The whole point of the %* and $@ characters in these scripts is to allow external arguments to come in, such as from Yaml. Please refactor this to call the same script with different args; this is too much duplication.

@MattGal can you be more specific? My problem was that `CIBuild.cmd` from Arcade commons is already calling with `-test`:
```
powershell -ExecutionPolicy ByPass -NoProfile -command "& """%~dp0Build.ps1""" -restore -build -test -sign -pack -publish -ci %*"
```

So whatever I append through `%*` won't help. Or there is a way? That is why I created the new cmd/sh.

MattGal:
> > The whole point of the %* and $@ characters in these scripts is to allow external arguments to come in, such as from Yaml. Please refactor this to call the same script with different args; this is too much duplication.
> 
> @MattGal can you be more specific? My problem was that `CIBuild.cmd` from Arcade commons is already calling with `-test`:
> 
> ```
> powershell -ExecutionPolicy ByPass -NoProfile -command "& """%~dp0Build.ps1""" -restore -build -test -sign -pack -publish -ci %*"
> ```
> 
> So whatever I append through `%*` won't help. Or there is a way? That is why I created the new cmd/sh.
> 
> I guess I can bypass it and call the build.cmd directly..

No you can just trump it via Msbuild properties.  Example: https://github.com/dotnet/arcade/blob/592ec36c7fcfbd4c3a6025c61456e351aa2c9057/azure-pipelines.yml#L64

azure-pipelines.yml
```diff
@@ -63,23 +63,33 @@ stages:
...
-            mergeTestResults: true
-            searchFolder: $(system.defaultworkingdirectory)
-            testRunTitle: XHarness unit tests - $(Agent.JobName)
-          condition: succeededOrFailed()
+        - ${{ if eq(variables._RunAsPublic, True) }}:
+          - task: PublishTestResults@2
+            displayName: 'Publish Unit Test Results'
+            inputs:
+              testResultsFormat: xUnit
+              testResultsFiles: '$(Build.SourcesDirectory)/artifacts/TestResults/**/*.xml'
```
mandel-macaque, line 88: AFAIK, we have NUnit and xUnit tests atm. We should move https://github.com/dotnet/xharness/blob/master/tests/Microsoft.DotNet.XHarness.iOS.Tests/AppInstallerTests.cs to xUnit. I have created an issue for that: https://github.com/dotnet/xharness/issues/65

azure-pipelines.yml
premun: I clumped the `Build` and then `Build and run tests`+`Publish tests` together and used the override for the Test MSBuild param. I tested both the internal and public runs to verify everything

azure-pipelines.yml
premun: Good catch, forgot to migrate those as well. Fixed in #68

Issue #73, opened 2020-04-23
[iOS] The `ios test` command hangs indefinitely


premun:
The root cause is this call:
https://github.com/dotnet/xharness/blob/9d7c784d6edd84f600405e6852359f837154f66b/src/Microsoft.DotNet.XHarness.iOS.Shared/Listeners/SimpleTcpListener.cs#L26

Somehow this hangs (might be this: https://github.com/dotnet/runtime/issues/24513) when we call `listener.Cancel()` from `AppRunner`:
https://github.com/dotnet/xharness/blob/9d7c784d6edd84f600405e6852359f837154f66b/src/Microsoft.DotNet.XHarness.iOS/AppRunner.cs#L350

PR #74, opened 2020-04-23
Prevent hang of XHarness when Listener doesn't initialize fully

premun:
This is not a final fix but rather unblock for others.

More information about the hang here:
https://github.com/dotnet/xharness/issues/73

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS/AppRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS/AppRunner.cs
@@ -353,7 +353,8 @@ namespace Microsoft.DotNet.XHarness.iOS
                 }
             }
 
-            listener.Cancel();
+            // TODO: https://github.com/dotnet/xharness/issues/73
+            // listener.Cancel();
             listener.Dispose();
 
             // check the final status, copy all the required data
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.XHarness.iOS/AppRunner.cs
```diff
@@ -347,7 +353,8 @@ public class AppRunner
                 }
             }
 
-            listener.Cancel();
+            // TODO: https://github.com/dotnet/xharness/issues/73
+            // listener.Cancel();
```
mandel-macaque, line 357: Hmm that is annoying. Crazy question, have you tried something int the listener.Cancel to instead of calling the Cancel of the TcpListener to access the Socket and close it? Something like:
```csharp
server.Server.Close();
```
Is a little brute but might work.

premun: Thanks, will give it a go

premun:
@mandel-macaque I had something like this:
https://github.com/premun/xharness/commit/572007ccd21ec873f98dbfb6d07a5dceee0caab0

But then I realized the timeout actually works, just the hang is after. I am also thinking of not setting any timeout inside the AppRunner and controlling it through cancellationToken from the CLI

premun:
I merged this so that Egor has something to work with

PR #81, opened 2020-04-24
Create Command/Argument namespace and make various small fixes and improvements

premun:
Bit of a cleanup that clumped in my branches. Will help a great deal for the ios package command and later for nulability eventually.
No functional changes:
- Move Arguments and Commands to separate namespaces
- Remove PackageCommand abstract level (Android doesn't have package command)
- Remove WorkingDirectory argument that is not used by iOS or Android, so we don't have to keep supplying it
- Add missing headers to some files
- Fix unit test namespaces
- Disable paralelization for tests that require serial execution


premun:
Thanks for the merge, @mandel-macaque!

Issue #87, opened 2020-04-25
[iOS] Paths are created with quotes.


mandel-macaque:
Executing the following:

```bash
dotnet artifacts/bin/Microsoft.DotNet.XHarness.CLI/Debug/netcoreapp3.1/Microsoft.DotNet.XHarness.CLI.dll ios test  --working-directory=“workdir” --output-directory=“outdir” --app=/Users/mandel/Xamarin/xamarin-macios/master/xamarin-macios/tests/xharness/tmp-test-dir/monotouch-test177/bin/iPhone/Debug64-unified/monotouchtest.app --targets=ios-device --mlaunch="/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/bin/mlaunch" -v
```

The created dirs contain quotes.
```
“workdir”
“outdir”
```

premun:
I think if you do regular quotes this doesn't happen. But the one in your example are not `"` but `”`. I've been using quotes and Mono.Options takes care of them for you normally.

mandel-macaque:
@premun is right, closing this guy.

PR #89, opened 2020-04-25
[iOS] Trim the " in a path added from the cli.

mandel-macaque:
If the user added " to add paths with spaces, trim them, else we will
create paths that contain the ".

fixes: https://github.com/dotnet/xharness/issues/87


mandel-macaque:
no need, looks like a zsh plugin I used was misbehaving.

Issue #91, opened 2020-04-27
[iOS] App completes successfully but XHarness tool fails


premun:
Egor has built me an app from these tests:
```
    public class DummyTests
    {
        [Fact]
        public void PassingTest()
        {
            Assert.Equal(1, 1);
        }

        [Fact]
        public void FailingTest()
        {
            Assert.Equal(1, 2);
        }

        [Fact(Skip = "Skipped reason")]
        public void SkippedTest()
        {
            Assert.Equal(1, 1);
        }
    }
```

The app runs fine, it terminates but we fail to parse the results:
```
Could not parse xml result file: System.Xml.XmlException: Data at the root level is invalid. Line 1, position 1.
   at System.Xml.XmlTextReaderImpl.Throw(Exception e)
   at System.Xml.XmlTextReaderImpl.Throw(String res, String arg)
   at System.Xml.XmlTextReaderImpl.ParseRootLevelWhitespace()
   at System.Xml.XmlTextReaderImpl.ParseDocumentContent()
   at System.Xml.XmlTextReaderImpl.Read()
   at Microsoft.DotNet.XHarness.iOS.Shared.XmlResultParser.ParseNUnitXml(StreamReader stream, StreamWriter writer) in /Users/prvysoky/Desktop/xharness/src/Microsoft.DotNet.XHarness.iOS.Shared/XmlResultParser.cs:line 188
   at Microsoft.DotNet.XHarness.iOS.Shared.XmlResultParser.GenerateHumanReadableResults(String source, String destination, XmlResultJargon xmlType) in /Users/prvysoky/Desktop/xharness/src/Microsoft.DotNet.XHarness.iOS.Shared/XmlResultParser.cs:line 370
   at Microsoft.DotNet.XHarness.iOS.Shared.TestReporter.ParseResultFile(AppBundleInformation appInfo, String test_log_path, Boolean timed_out) in /Users/prvysoky/Desktop/xharness/src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs:line 374
File data is:
##########
01:02:43.8921440 [Local Date/Time:     04/27/2020 01:02:43]
[Remote Address:       127.0.0.1:52126]
<?xml version="1.0" encoding="utf-8"?>
<test-results name="Test results" errors="0" inconclusive="0" ignored="0" invalid="0" not-run="0" date="2020-04-27" time="01:02:45" total="3" failures="1" skipped="1">
[...]

PR #93, opened 2020-04-27
Clean-up SimpleListeners and fix invalid XML header

premun:
As mentioned here: https://github.com/dotnet/xharness/issues/91#issuecomment-619859563, we include some garbage in the results XML

### DIFF
--- a/tests/Microsoft.DotNet.XHarness.iOS.Tests/AppRunnerTests.cs
+++ b/tests/Microsoft.DotNet.XHarness.iOS.Tests/AppRunnerTests.cs
@@ -274,6 +274,7 @@ namespace Microsoft.DotNet.XHarness.iOS.Tests
             _listener.Verify(x => x.Initialize(), Times.AtLeastOnce);
             _listener.Verify(x => x.StartAsync(), Times.AtLeastOnce);
             // _listener.Verify(x => x.Cancel(), Times.AtLeastOnce); // TODO: https://github.com/dotnet/xharness/issues/73
+            _listener.Verify(x => x.Dispose(), Times.AtLeastOnce);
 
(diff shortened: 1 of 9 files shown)


tests/Microsoft.DotNet.XHarness.iOS.Tests/AppRunnerTests.cs
```diff
@@ -274,7 +274,6 @@ public async Task RunOnSimulatorSuccessfullyTest()
             _listener.Verify(x => x.Initialize(), Times.AtLeastOnce);
             _listener.Verify(x => x.StartAsync(), Times.AtLeastOnce);
             // _listener.Verify(x => x.Cancel(), Times.AtLeastOnce); // TODO: https://github.com/dotnet/xharness/issues/73
-            _listener.Verify(x => x.Dispose(), Times.AtLeastOnce);
```
mandel-macaque, line 277: Why was this removed? should the app runner be disposing the listener?

tests/Microsoft.DotNet.XHarness.iOS.Tests/AppRunnerTests.cs
```diff
@@ -402,7 +401,6 @@ public async Task RunOnDeviceSuccessfullyTest(bool useTunnel)
...
```
mandel-macaque, line 405: Same, why did we remove this?

tests/Microsoft.DotNet.XHarness.iOS.Tests/AppRunnerTests.cs
premun: A mistake - bad revert, will fix

Issue #90, opened 2020-04-25
xharness packages are no longer published


marek-safar:
the last available NuGet package is from 21/4/2020 (version number .20221.3)

premun:
@marek-safar, thanks - we know about this, we were broken by arcade and a fix should be on the way. I coordinated with Egor the first day it happened saying I can provide a manually built package in case it becomes a blocker but he said he is working on Android now, so I decided to just wait.

premun:
There was an update from Arcade today but the issue remains unresolved. I will talk to Matt/Ricardo today to get it fixed.

https://dev.azure.com/dnceng/internal/_build/results?buildId=619068&view=logs&j=0a6c679f-72be-5867-422e-acb741a068a3

premun:
@marek-safar I unblocked our pipeline with a workaround, once this finishes:
https://dev.azure.com/dnceng/internal/_build/results?buildId=619174&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76&j=8d2248bb-cd55-5169-11d5-7c5b1f0b2b92

we should have new versions of XHarness

premun:
We have a new package:
https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet-eng&package=Microsoft.DotNet.XHarness.CLI&protocolType=NuGet&version=1.0.0-prerelease.20227.4&view=overview

PR #99, opened 2020-04-27
Disable signing validation to unblock the pipeline

premun:
We were broken by Arcade so here's a workaround before they fix it

More info: https://github.com/dotnet/arcade/issues/5328


premun:
Fixes: #90

premun:
Egor is waiting for this, so I will close it ASAP

PR #100, opened 2020-04-27
[iOS][Tests] Add test to make sure that the skipped count is correct.

mandel-macaque:
Add a test to make sure that when there are skipped tests those are
correctly reported.


tests/Microsoft.DotNet.XHarness.iOS.Shared.Tests/Microsoft.DotNet.XHarness.iOS.Shared.Tests.csproj
```diff
@@ -51,6 +51,8 @@
     <EmbeddedResource Include="Samples\SystemXunit.csproj" />
     <EmbeddedResource Include="Samples\TouchUnitSample.xml" />
     <EmbeddedResource Include="Samples\xUnitSample.xml" />
+    <None Remove="Samples\Issue95.xml" />
```
premun, line 54: nit: I think this line is extra

PR #101, opened 2020-04-27
[iOS] Fix hangs when stopping the TCP listener

premun:
Fixes: #73

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS.Shared/Listeners/SimpleTcpListener.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS.Shared/Listeners/SimpleTcpListener.cs
@@ -41,7 +41,10 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared.Listeners
         {
             _client?.Close();
             _client?.Dispose();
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.XHarness.iOS.Shared/Listeners/SimpleTcpListener.cs
```diff
@@ -41,7 +41,10 @@ protected override void Stop()
         {
             _client?.Close();
             _client?.Dispose();
-            _server?.Stop();
+
+            // _server?.Stop(); was causing hangs
+            // https://github.com/dotnet/xharness/issues/73
+            _server?.Server?.Shutdown(SocketShutdown.Both);
```
mandel-macaque, line 47: Brute force, but it works :)

premun: Yeah, it's not 100% but also no difference to not closing the connection properly either :)

premun:
Build is green, just GH doesn't know about it

Issue #98, opened 2020-04-27
[iOS] Generate XML test results in xUnit format by default


mandel-macaque:
xUnit is the basic framework used by mono, lets move to always use xUnit and let other projects specify the desired output. This means adding a new option to the test command that will allow to specify the xml jargon to use.

Issue #77, opened 2020-04-23
[iOS] Xml results are not written in the oputdir.


mandel-macaque:
Probably we are missing the xml reporter of (works case) we have tcp network issues. That will be solved when we land https://github.com/dotnet/xharness/pull/66 and move to use tunnels.

premun:
I think this can be closed now

PR #102, opened 2020-04-27
[iOS] Ensure that we do generate xml when we call the CLI.

mandel-macaque:
During the refactor we removed the env vars that are passed to the
runners to decide if xml is used and the jargon to use. The env
paramenters have been added and tests updated. Now the runners will get
notified that xml is used (always) and the jargon (default to xunit).

fixes: https://github.com/dotnet/xharness/issues/98
fixes: https://github.com/dotnet/xharness/issues/77

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/iOS/iOSTestCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/iOS/iOSTestCommandArguments.cs
@@ -37,6 +37,11 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.iOS
         /// </summary>
         public TimeSpan LaunchTimeout { get; set; } = TimeSpan.FromMinutes(5);
 
+        /// <summary>
+        /// Allows to specify the xml format to be used in the result files.
+        /// </summary>
+        public XmlResultJargon XmlResultJargon { get; set; } = XmlResultJargon.xUnit; // default by mono
+
         public IReadOnlyCollection<TestTarget> TestTargets { get; set; }
 
         public override IList<string> GetValidationErrors()
@@ -64,7 +69,7 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.iOS
                         // let the user know that the target is not known
                         // and all the available ones.
                         var sb = new StringBuilder();
-                        sb.AppendLine($"Failed to parse test target '{targetName}'. Avaliable targets are:");
+                        sb.AppendLine($"Failed to parse test target '{targetName}'. Available targets are:");
                         sb.AppendLine();
                         foreach (var val in Enum.GetValues(typeof(TestTarget)))
                         {
@@ -99,6 +104,17 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.iOS
                 errors.Add($"Failed to find Xcode root at {XcodeRoot}");
             }
 
+            if (XmlResultJargon == XmlResultJargon.Missing)
+            {
+                var sb = new StringBuilder();
+                sb.AppendLine($"Failed to parse xml jargon. Available targets are:");
+                sb.AppendLine();
+                foreach (var val in new [] {XmlResultJargon.NUnitV2, XmlResultJargon.NUnitV3, XmlResultJargon.TouchUnit, XmlResultJargon.xUnit})
+                {
+                    sb.AppendLine($"\t* {val.ToString()}");
+                }
+                errors.Add(sb.ToString());
+            }
             return errors;
         }
 
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/iOS/iOSTestCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/iOS/iOSTestCommand.cs
@@ -178,7 +178,7 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.iOS
                     mainLog,
                     logs,
                     new Helpers(),
-                    true, // the cli ALWAYS will get the output as xml
+                    useXmlOutput: true, // the cli ALWAYS will get the output as xml
                     useTcpTunnel: _channel == CommunicationChannel.UsbTunnel);
 
                 (deviceName, success) = await appRunner.RunApp(appBundleInfo,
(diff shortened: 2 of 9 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/iOS/iOSTestCommand.cs
```diff
@@ -172,6 +178,7 @@ protected override async Task<ExitCode> InvokeInternal()
                     mainLog,
                     logs,
                     new Helpers(),
+                    true, // the cli ALWAYS will get the output as xml
```
premun, line 181: nit: would be nice to have the param name

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/iOS/iOSTestCommandArguments.cs
```diff
@@ -99,6 +104,17 @@ public override IList<string> GetValidationErrors()
                 errors.Add($"Failed to find Xcode root at {XcodeRoot}");
             }
 
+            if (XmlResultJargon == XmlResultJargon.Missing)
+            {
+                var sb = new StringBuilder();
+                sb.AppendLine($"Failed to parse xml jargon. Available targets are:");
+                sb.AppendLine();
+                foreach (var val in new [] {XmlResultJargon.NUnitV2, XmlResultJargon.NUnitV3, XmlResultJargon.TouchUnit, XmlResultJargon.xUnit})
```
premun, line 112: We are doing this quite a lot around and could use some unified way of converting String->Enum->List of values. 
https://github.com/dotnet/xharness/blob/7ff61e57d730e9bd4da7849cece139ae92e5bbb1/src/Microsoft.DotNet.XHarness.CLI/Commands/iOS/iOSPackageCommand.cs#L276-L283

Leave it here as is, if we decide to do nullability, we will revisit these places anyway. But the future one would be nice to be case insensitive because I am quite struggling to type xUnit and NUnit correctly :)

src/Microsoft.DotNet.XHarness.iOS/AppRunner.cs
```diff
@@ -133,6 +136,14 @@ public class AppRunner
                 args.Add(new SetEnvVariableArgument("USE_TCP_TUNNEL", true));
             }
 
+            if (_useXmlOutput)
```
premun, line 139: Sorry for having this slipped through the refactor ☺

premun:
Doesn't this also require changes in the TestRunner to enable XUnit there? I noticed for example that the jargon enum used for parsing the input options didn't even have xUnit value

premun:
Yeah, it will. I tried to run this and got following:
```
Unhandled Exception:
System.TypeInitializationException: The type initializer for 'Microsoft.DotNet.XHarness.Tests.Runners.Core.ApplicationOptions' threw an exception.
 ---> System.ArgumentException: Requested value 'xUnit' was not found.
   at System.Enum.TryParseByName(RuntimeType enumType, String originalValueString, ReadOnlySpan`1 value, Boolean ignoreCase, Boolean throwOnFailure, UInt64& result) in /Users/egorbo/prj/runtime-3/src/libraries/System.Private.CoreLib/src/System/Enum.cs:line 722
   at System.Enum.TryParseInt32Enum(RuntimeType enumType, String originalValueString, ReadOnlySpan`1 value, Int32 minInclusive, Int32 maxInclusive, Boolean ignoreCase, Boolean throwOnFailure, TypeCode type, Int32& result) in /Users/egorbo/prj/runtime-3/src/libraries/System.Private.CoreLib/src/System/Enum.cs:line 479
   at System.Enum.TryParse(Type enumType, String value, Boolean ignoreCase, Boolean throwOnFailure, Object& result) in /Users/egorbo/prj/runtime-3/src/libraries/System.Private.CoreLib/src/System/Enum.cs:line 328
   at System.Enum.Parse(Type enumType, String value, Boolean ignoreCase) in /Users/egorbo/prj/runtime-3/src/libraries/System.Private.CoreLib/src/System/Enum.cs:line 271
   at Microsoft.DotNet.XHarness.Tests.Runners.Core.ApplicationOptions..ctor()
   at Microsoft.DotNet.XHarness.Tests.Runners.Core.ApplicationOptions..cctor()
   --- End of inner exception stack trace ---
```

mandel-macaque:
hmm @premun let me check

src/Microsoft.DotNet.XHarness.iOS.Shared/TestImporter/Templates/Managed/Resources/src/tvOSApp/ViewController.cs
```diff
@@ -105,13 +105,14 @@ public async override void ViewDidLoad()
             Xamarin.iOS.UnitTests.TestRunner.Jargon jargon = Xamarin.iOS.UnitTests.TestRunner.Jargon.NUnitV3;
```
premun, line 105: Is there a different default here?

src/Microsoft.DotNet.XHarness.iOS.Shared/TestImporter/Templates/Managed/Resources/src/watchOS/Extension/InterfaceController.cs
```diff
@@ -140,6 +140,9 @@ void RunTests ()
 					case XmlVersion.NUnitV3:
```
premun, line 140: Again, I think here is a NUnitV3 default?

mandel-macaque: It is already the default in https://github.com/dotnet/xharness/pull/102/files#diff-dc254c28cb157a76f159304a4405239bR134 but if we get a diff from the options we respect it.

src/Microsoft.DotNet.XHarness.iOS.Shared/TestImporter/Templates/Managed/Resources/src/tvOSApp/ViewController.cs
mandel-macaque: No, is NUnitV3, but we are passing via the args xUnit, should not matter for the manage packaged apps.

PR #106, opened 2020-04-27
[iOS] Unify the env vars to be shared between the libs.

mandel-macaque:
Add the env vars as variables and use the same static class in the
runners project and in the cli to make sure that we do not have small
errors due to typos.


src/Microsoft.DotNet.XHarness.iOS.Shared/Execution/Mlaunch/EnviromentVariables.cs
```diff
@@ -0,0 +1,73 @@
+namespace Microsoft.DotNet.XHarness.iOS.Shared.Execution.Mlaunch
+{
+    public static class EnviromentVariables
+    {
+
+        /// <summary>
+        /// Env var that will tell the runner to start the execution of the tests automatically.
+        /// </summary>
+        public static string AutoStart = "NUNIT_AUTOSTART";
```
premun, line 9: Make them `const`

tests/Microsoft.DotNet.XHarness.iOS.Tests/AppRunnerTests.cs
```diff
@@ -264,11 +264,11 @@ public async Task RunOnSimulatorSuccessfullyTest(bool useTunnel, bool useXml)
             var xmlParam = useXml ? "-setenv=NUNIT_ENABLE_XML_OUTPUT=true -setenv=NUNIT_ENABLE_XML_MODE=wrapped -setenv=NUNIT_XML_VERSION=xUnit " : "";
 
             var expectedArgs = $"-argument=-connection-mode -argument=none -argument=-app-arg:-autostart " +
-                $"-setenv=NUNIT_AUTOSTART=true -argument=-app-arg:-autoexit -setenv=NUNIT_AUTOEXIT=true " +
-                $"-argument=-app-arg:-enablenetwork -setenv=NUNIT_ENABLE_NETWORK=true -setenv=DISABLE_SYSTEM_PERMISSION_TESTS=1 -v -v " +
-                $"-argument=-app-arg:-hostname:127.0.0.1 -setenv=NUNIT_HOSTNAME=127.0.0.1 -argument=-app-arg:-transport:Tcp " +
-                $"-setenv=NUNIT_TRANSPORT=TCP -argument=-app-arg:-hostport:{_listener.Object.Port} " +
-                $"-setenv=NUNIT_HOSTPORT={_listener.Object.Port} {xmlParam}--launchsim {StringUtils.FormatArguments(s_appPath)} " +
+                $"-setenv={EnviromentVariables.AutoStart}=true -argument=-app-arg:-autoexit -setenv={EnviromentVariables.AutoExit}=true " +
```
premun, line 267: I think we might want to keep strings here in tests. Current way, if we make a mistake in the variable name, we will never catch it. Ideally you would change to constants and tests don't change and you know you didn't change anything

mandel-macaque: Well, if we update the constants, we will get failure.. I'm not sure about that. Is not that the constant are going to change a lot..

premun: Well, if they change, we changed behavior of the program and tests should change.
But for example now I cannot be sure you didn't make a mistake in any of those and it's hard to validate by eye. So we are not testing much if we reference them from tests, we just wrote almost the same code twice?

mandel-macaque: I can revert that, not a big issue

PR #109, opened 2020-04-28
[iOS] Improve the TunnelBore API.

mandel-macaque:
Remove the need to use a bool. Allow the factory to take a null tunnel
bore (less memory used) and add a property that will tell use if the
factory will use a tunnel bore.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/iOS/iOSTestCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/iOS/iOSTestCommand.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DotNet.XHarness.CLI.CommandArguments;
@@ -76,7 +77,7 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.iOS
             var processManager = new ProcessManager(_arguments.XcodeRoot, _arguments.MlaunchPath);
             var deviceLoader = new HardwareDeviceLoader(processManager);
             var simulatorLoader = new SimulatorLoader(processManager);
-            var tunnelBore = new TunnelBore(processManager);
+            var tunnelBore = (_channel == CommunicationChannel.UsbTunnel) ? new TunnelBore(processManager) : null;
 
             var logs = new Logs(_arguments.OutputDirectory);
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.XHarness.iOS.Shared/Listeners/SimpleListenerFactory.cs
```diff
@@ -26,17 +26,18 @@ public interface ISimpleListenerFactory
...
 
         public ITunnelBore TunnelBore { get; private set; }
 
-        public SimpleListenerFactory(ITunnelBore tunnelBore)
-        {
-            TunnelBore = tunnelBore ?? throw new ArgumentNullException(nameof(tunnelBore));
-        }
+        public bool UseTunnel => TunnelBore != null;
+
+        public SimpleListenerFactory(ITunnelBore tunnelBore) =>
```
premun, line 39:
```suggestion
        public SimpleListenerFactory(ITunnelBore tunnelBore = null) =>
```

src/Microsoft.DotNet.XHarness.CLI/Commands/iOS/iOSTestCommand.cs
```diff
@@ -76,7 +77,7 @@ protected override async Task<ExitCode> InvokeInternal()
             var processManager = new ProcessManager(_arguments.XcodeRoot, _arguments.MlaunchPath);
             var deviceLoader = new HardwareDeviceLoader(processManager);
             var simulatorLoader = new SimulatorLoader(processManager);
-            var tunnelBore = new TunnelBore(processManager);
+            var tunnelBore = (_channel == CommunicationChannel.UsbTunnel) ? new TunnelBore(processManager) : null;
```
premun, line 80: I think it would be better to have `Channel` in `_arguments` so that it is together with the rest.

I think later the whole `OptionSet` will actually move to `Arguments` (when we do nullability). It also makes more sense responsibility wise. So we would forget something behind like this.

src/Microsoft.DotNet.XHarness.iOS/AppRunner.cs
```diff
@@ -131,7 +128,7 @@ public class AppRunner
             args.Add(new SetAppArgumentArgument($"-hostport:{listener.Port}", true));
             args.Add(new SetEnvVariableArgument("NUNIT_HOSTPORT", listener.Port));
 
-            if (_useTcpTunnel && !isSimulator) // simulators do not support tunnels
+            if (_listenerFactory.UseTunnel && !isSimulator) // simulators do not support tunnels
```
premun, line 131: I think it's somehow a bit weird we are asking Factory for things. Maybe this could be another return value from `Create`?

mandel-macaque: not a good idea for now, I'll merge will create a pr that fixes things and we can take a second look.

Issue #88, opened 2020-04-25
[iOS] If application cannot deploy to device to wrong arch notify it to the user


mandel-macaque:
We can tell that the app could not be installed because the target device has a diff arch by parsing the installation logs. Rather than telling the user to look at the logs, parse the installation logs and notify the error in a nicer way.

PR #122, opened 2020-04-28
[iOS] If the user tries to install the app in the wrong arch, let him know.

mandel-macaque:
If a user tries to execute the test command with an application and
the wrong device arch, do notify him since we do know is an expected
error.

Sample output:
```
info: test[0]
      Starting test for ios-device targeting XQAiPodTouch5f..
info: test[0]
      Installing application 'MonoTouchTest' on  on device 'XQAiPodTouch5f'
fail: test[0]
      Failed to install the app bundle (exit code=1): IncorrectArchitecture: Failed to find matching arch for 64-bit Mach-O input file /private/var/installd/Library/Caches/com.apple.mobile.installd.staging/temp.Z91Igw/extracted/monotouchtest.app/monotouchtest
XHarness exit code: -43
```

fixes: https://github.com/dotnet/xharness/issues/88


mandel-macaque:
If we have this issues in CI it will make it easier to know the reason.

Issue #126, opened 2020-04-29
[iOS] If targets contain simulators the tunnel is used.


mandel-macaque:
xharness tries to use the tunnel when there are simulators in the targets. If there are any targets that are simulators, this should not happen since a tunnel cannot be created to a simulator.

PR #127, opened 2020-04-29
[iOS] If we use simulators do not use a tunnel.

mandel-macaque:
Check what targets are passed to the CLI and do not use tunnels in that
case.

fixes: https://github.com/dotnet/xharness/issues/126


src/Microsoft.DotNet.XHarness.CLI/Commands/iOS/iOSTestCommand.cs
```diff
@@ -78,7 +78,8 @@ protected override async Task<ExitCode> InvokeInternal()
             var processManager = new ProcessManager(_arguments.XcodeRoot, _arguments.MlaunchPath);
             var deviceLoader = new HardwareDeviceLoader(processManager);
             var simulatorLoader = new SimulatorLoader(processManager);
-            var tunnelBore = (_channel == CommunicationChannel.UsbTunnel) ? new TunnelBore(processManager) : null;
+            var hasSimulators = _arguments.TestTargets.Where(t => t.IsSimulator()).Any();
```
premun, line 81: Can you please decide this lower in the foreach and do tunnels for devices and network for sims?

PR #129, opened 2020-04-29
Adb linux support

MattGal:
I am still testing, but making sure including the binary works right with SignTool stuff


MattGal:
builds finished and this is a tiny change, I'm just merging.

PR #133, opened 2020-04-30
Report launch failure as crash

premun:


src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
```diff
@@ -580,7 +580,7 @@ public async Task<(TestExecutingResult ExecutingResult, string ResultMessage)> P
             {
                 result.ExecutingResult = TestExecutingResult.TimedOut;
             }
-            else if (crashed)
+            else if (crashed || launchFailure)
```
mandel-macaque, line 583: Approving, but maybe changing from TestExecuting.Crashed to LauchFailure would make more sense.

premun:
Merging in favor of https://github.com/dotnet/xharness/issues/143 since this might be a bit more complicated with the result flags

PR #134, opened 2020-04-30
Make the Runners.Tests project a friendly assembly

premun:
Allows testing of internals


mandel-macaque:
Merging so that it does not add conflicts with current work I have.

Issue #137, opened 2020-04-30
Fix ActiveIssue change


akoeplinger:
The change in https://github.com/dotnet/xharness/pull/130 is wrong/unnecessary as it skips test methods with the attribute rather than skipping the `category=failing` trait generated by the [ActiveIssueDiscoverer](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.XUnitExtensions/src/Discoverers/ActiveIssueDiscoverer.cs) in the arcade xunit extensions.

We should instead just pass the -notrait in.

akoeplinger:
@mandel-macaque

EgorBo:
Tried the latest xharness (20230.1) and crashes
```
Unhandled Exception:
System.ArgumentNullException: Value cannot be null. (Parameter 'key')
   at System.Collections.Concurrent.ConcurrentDictionary`2[[System.Type, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.AttributeUsageAttribute, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].ThrowKeyNullException()
   at System.Collections.Concurrent.ConcurrentDictionary`2[[System.Type, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.AttributeUsageAttribute, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].GetOrAdd(Type key, Func`2 valueFactory)
   at Xunit.Sdk.ReflectionAttributeInfo.GetAttributeUsage(Type attributeType) in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Reflection\ReflectionAttributeInfo.cs:line 57
   at Xunit.Sdk.ReflectionMethodInfo.GetCustomAttributes(MethodInfo method, String assemblyQualifiedAttributeTypeName) in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Reflection\ReflectionMethodInfo.cs:line 87
   at Xunit.Sdk.ReflectionMethodInfo.GetCustomAttributes(String assemblyQualifiedAttributeTypeName) in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Reflection\ReflectionMethodInfo.cs:line 80
[...]

PR #138, opened 2020-04-30
[Runners] Do not filter per Attr, make client use traits.

mandel-macaque:
We support traits from the very first release, those have to be passed in
a file that is provided by the child of the entry point:

IgnoreFilesDirectory.

That directory can contain two diff files:

* xunit-excludes.txt
* nunit-excludes.txt

I have updated the docs to describe what should be done since is the
root of the misscommunication.

fixes: https://github.com/dotnet/xharness/issues/137

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Tests.Runners/IgnoreFileParser.cs
+++ b/src/Microsoft.DotNet.XHarness.Tests.Runners/IgnoreFileParser.cs
@@ -68,6 +68,9 @@ namespace Microsoft.DotNet.XHarness.Tests.Runners.Core
 
         public static async Task<IEnumerable<string>> ParseContentFilesAsync(string contentDir)
         {
+            if (string.IsNullOrEmpty(contentDir))
+                return Array.Empty<string>();
+
             var ignoredTests = new List<string>();
             foreach (var f in Directory.GetFiles(contentDir, "*.ignore"))
             {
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.XHarness.Tests.Runners/IgnoreFileParser.cs
```diff
@@ -80,23 +83,26 @@ public static async Task<IEnumerable<string>> ParseContentFilesAsync(string cont
...
+            while ((line = await reader.ReadLineAsync()) != null)
             {
-                string line;
-                while ((line = await reader.ReadLineAsync()) != null)
-                {
-                    if (string.IsNullOrEmpty(line))
-                        continue;
-                    ignoredTraits.Add(line);
-                }
+                if (string.IsNullOrEmpty(line))
```
premun, line 93: Should this also test for `StartsWith('#')`?

mandel-macaque: I did not add support for comments in the traits files, we could do so @akoeplinger @EgorBo ?

EgorBo: up to you, I don't think we need that, I doubt we'll ever touch that file by hands.

PR #139, opened 2020-04-30
[iOS] Add support to parse the arch of a csproj or a plist.

mandel-macaque:
Add code that will allow to parse the arch of a bundled app and a
project this will later help to use the correct device when the device
name is not provided.

We yet do not select the correct arch, this is the first step to do so.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformationParser.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformationParser.cs
@@ -87,7 +87,7 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared
                 ? Directory.GetDirectories(Path.Combine(appPackagePath, "Watch"), "*.app")[0]
                 : appPackagePath;
 
-            return new AppBundleInformation(appName, appPackagePath, bundleIdentifier, launchAppPath, supports32.Contains(Armv7), extension: null);
+            return new AppBundleInformation(appName, appPackagePath, bundleIdentifier, launchAppPath, supports32.Contains(Armv7, StringComparison.InvariantCultureIgnoreCase), extension: null);
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformationParser.cs
```diff
@@ -56,14 +57,17 @@ public AppBundleInformation ParseFromProject(string projectFilePath, TestTarget
                 csproj.GetOutputPath(platform, buildConfiguration).Replace('\\', Path.DirectorySeparatorChar),
                 appName + (extension != null ? ".appex" : ".app"));
 
+            string arch = csproj.GetMtouchArch(platform, buildConfiguration);
+            bool supports32 = arch.Contains("ARMv7") || arch.Contains("i386");
```
premun, line 61: Do you want `StringComparison.InvariantCultureIgnoreCase` or these are always capitalized? Just asking because the other Arm const is in a different case

mandel-macaque: @premun better to ignore the case, good point.

src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformationParser.cs
```diff
@@ -58,7 +58,7 @@ public AppBundleInformation ParseFromProject(string projectFilePath, TestTarget
                 appName + (extension != null ? ".appex" : ".app"));
 
             string arch = csproj.GetMtouchArch(platform, buildConfiguration);
-            bool supports32 = arch.Contains("ARMv7") || arch.Contains("i386");
+            bool supports32 = arch.Contains("ARMv7", StringComparison.InvariantCultureIgnoreCase) || arch.Contains("i386", StringComparison.InvariantCultureIgnoreCase);
```
premun, line 61: Also line 90

PR #140, opened 2020-04-30
MacOS Support - include native adb binary

MattGal:
A parallel change was all that linux needed.  I'm still working on getting this to a mac to try it.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -57,10 +57,15 @@ namespace Microsoft.DotNet.XHarness.Android
             {
                 return Path.Join(currentAssemblyDirectory, @"..\..\..\runtimes\any\native\adb\windows\adb.exe");
             }
-            else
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -57,10 +57,15 @@ private static string GetCliAdbExePath()
...
-            else
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
             {
                 return Path.Join(currentAssemblyDirectory, @"../../../runtimes/any/native/adb/linux/adb");
             }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return Path.Join(currentAssemblyDirectory, @"../../../runtimes/any/native/adb/macos/adb");
+            }
+            throw new NotSupportedException("Cannot determine OS platform being used, thus we can not select ADB executable.");
```
MattGal, line 68: (This would be only hittable on FreeBSD)

MattGal:
I don't have a mac near me or with an android emu on it but given linux works with the same change and basic get-state works with this change, I'm going to merge:

image

Issue #141, opened 2020-04-30
[Runners] Expose LogExcludedTests so that we can log excluded tests.


mandel-macaque:
Allow runners to set the LogExcludedTests to true to see which tests are ignored.

PR #142, opened 2020-04-30
[Runners] Expose a property to allow to log the skipped tests.

mandel-macaque:
Expose the property so that we can show which of the tests got skipped.

fixes: https://github.com/dotnet/xharness/issues/141


src/Microsoft.DotNet.XHarness.Tests.Runners/ApplicationEntryPoint.cs
```diff
@@ -107,6 +107,12 @@ public abstract class ApplicationEntryPoint
         /// </summary>
         protected virtual string IgnoredTraitsFilePath => null;
 
+        /// <summary>
+        /// States if the skipped tests should be logger. Helpful to determine why some tests are ran and others
```
MattGal, line 111: "should be logged", "are run" or maybe "are executed"

Issue #16, opened 2020-04-06
Enable nullable references


premun:
We can consider turning on nullable references which would totally make sense at least for the CLI project where we really care for arguments to be set properly. Other projects could follow.
For iOS.Shared it might be a problem to enable it for the whole project at once since it's quite large but we can definitely try it and see. It would also make sense there because we really care about proper flow of arguments from the CLI to mlaunch and similar.

premun:
@MattGal we have spoken about this with @mandel-macaque briefly on Friday and we agree that we would be up for trying this out. Your thoughts?

premun:
I had a first stab over the weekend at this and it gets really tricky with the MonoOptions parsing.

MattGal:
> @MattGal we have spoken about this with @mandel-macaque briefly on Friday and we agree that we would be up for trying this out. Your thoughts?

I think it's fancy technology for fancy technology's sake, and maybe not the best use of your time.  But if you're doing it, go for it!

premun:
That's why I did it on my own time since it is mostly my curiosity. But I disagree it's for fancy technology's sake. There is a lot of value in it that prevents a really nasty class of runtime errors. I think for us it might be even more important since this is not a service and we do not collect logs from the usage of the tool so if it crashes for people, it might be hard to repro.

After turning it on it already showed me bugs in the CLI that would result in crashes.

MattGal:
Sure.  It sounds like the kind of thing that's cheaper to do now than later...

PR #145, opened 2020-05-03
Enable C# 8.0 nullability for the CLI, Android and iOS projects, improve help command behaviour

premun:
This enables the nullability and resolves all warnings connected to that in the projects.

There were a couple of changes I had to do (arguments re-work) in order to achieve this and some I did to improve working with the code:
- Moved option definition to Argument classes. This makes the code easier to work with because the option definitions are together with the properties that they fill and their validation.
- Improved the help command so that you can run `help [command] [subcommand]` which was not possible before. Now it's easier to explore the tool using calling `help xy`. It now also works properly even when you mess up the subcommand name.
- Moved the Darwin class to the iOS project and kept the CLI project strictly CLI.
- Dedicate the `ios test t=` argument to the `targets` option. It was originally assigned to `timeout` but `timeout` is optional and `targets` not. This change allows to call `ios test` with all required args using shortcuts (`-o -a -t`)

Fixes: #16


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/XHarnessCommandArguments.cs
```diff
@@ -23,5 +39,67 @@ protected static string RootPath(string path)
...
+            if (invalidValues != null)
+            {
+                values = values.Where(v => !invalidValues.Contains(v));
+            }
+
+            var separator = Environment.NewLine + "\t- ";
+            IEnumerable<string?> allowedValued = values.Select(t => display?.Invoke(t) ?? t.ToString());
+
+            return separator + string.Join(separator, allowedValued);
+        }
```
mandel-macaque, line 103: I like this class :)

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/XHarnessCommandArguments.cs
```diff
@@ -23,5 +39,67 @@ protected static string RootPath(string path)
...
+        /// <summary>
+        /// Helper method that enables parsing of enums from string.
+        /// When an invalid value is supplied, available values are printed.
+        /// </summary>
+        /// <typeparam name="TEnum">Enum type</typeparam>
+        /// <param name="argumentName">Name of the argument that is being parsed, strictly for help printing purposes</param>
+        /// <param name="value">Value of the arg to be parsed</param>
+        /// <param name="invalidValues">List of values that should not be available to set</param>
+        /// <returns>Parsed enum value</returns>
+        protected static TEnum ParseArgument<TEnum>(string argumentName, string? value, params TEnum[]? invalidValues) where TEnum : struct, IConvertible
```
mandel-macaque, line 52: The IConvertible is a nice trick.

premun:
What is missing is the Tests.Runners project

I tried turning it on for the test projects and 2 of them were fine, iOS.Shared.Tests was not. But not sure we want that for tests

MattGal:
Change is huge! Did you test any of the android side before making the PR?

premun:
@MattGal I did test every command and I tested calling help for every suite/command as well

Issue #53, opened 2020-04-20
Xunit should respect `[ActiveIssue]` attribute


EgorBo:
We don't use "rsp/ignore list" approach in `dotnet/runtime` anymore, thus `xharness` needs to support
`[ActiveIssue]`, e.g. https://github.com/dotnet/runtime/blob/110282c71b3f7e1f91ea339953f4a0eba362a62c/src/libraries/System.ComponentModel.Composition/tests/System/Integration/CompositionContainerAttributedModelCycleTests.cs#L33

we've marked several hundreds of `dotnet/runtime` tests with it.

https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.XUnitExtensions/src/Discoverers/ActiveIssueDiscoverer.cs

EgorBo:
Turns out there are ~2000 ActiveIssues (half of them for mono) so we can't run the tests until this issue if fixed.

premun:
Apparently this didn't work and the tests are still being executed

premun:
Ok, the issue was - there was a misunderstanding and a **category** named like that was ignored, not the attribute

marek-safar:
@premun @mandel-macaque is there any ETA when this will be resolved?

premun:
@mandel-macaque I will take this and do it today

premun:
@mandel-macaque is taking over this after all. ETA hopefully today, Manuel?

PR #146, opened 2020-05-04
[Runners] Refactor filters and fix a number of issues in them.

mandel-macaque:
Filter were not working properly and not well tested. Added a few
changes to make things better:

* Added the filtering logic to the filter away from the test runner.
* Added the logic of grouping filters to a class.
* Added tests that cover the filtering done by the different types of filters.
* Added tests that cover the filtering done by groups of filters.

This commit should fix issue
https://github.com/dotnet/xharness/issues/53 once and for all.


src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitFilter.cs
```diff
@@ -2,23 +2,28 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#nullable enable
```
premun, line 5: \o/

src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitFilter.cs
```diff
@@ -71,29 +80,176 @@ public static XUnitFilter CreateClassFilter(string className, bool exclude, stri
...
+                return log(Exclude);
+            }
+
+            string fileName = Path.GetFileName(assembly.FullPath);
+            if (string.Equals(fileName, AssemblyName, StringComparison.Ordinal))
+            {
+                return log (Exclude);
+            }
+
+            // nor path of name matched the filter, therefore return the oposite of the Exclude value
```
premun, line 191:
```suggestion
            // No path of the name matched the filter, therefore return the opposite of the Exclude value
```

src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitFilter.cs
```diff
@@ -71,29 +80,176 @@ public static XUnitFilter CreateClassFilter(string className, bool exclude, stri
...
+                return excluded;
+
+            var selector = FilterType == XUnitFilterType.Trait ?
+                $"'{SelectorName}':'{SelectorValue}'" : $"'{SelectorValue}'";
+
+            log($"[FILTER] {(excluded? excludedText : includedText)} test (filtered by {FilterType}; {selector}): {testCase.DisplayName}");
+            return excluded;
+        }
+
+        bool ReportFilteredAssembly(TestAssemblyInfo assemblyInfo, bool excluded, Action<string>? log = null)
```
premun, line 225: I don't understand why this returns `excluded` back? For example in the usage above:
```csharp
return log(Excluded);
```

Is is quite unclear what is going on

src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitFilter.cs
```diff
@@ -71,29 +80,176 @@ public static XUnitFilter CreateClassFilter(string className, bool exclude, stri
...
+        bool ApplyTraitFilter(ITestCase testCase, Func<bool, bool>? reportFilteredTest = null)
+        {
+            Func<bool, bool> log = (result) => reportFilteredTest?.Invoke(result) ?? result;
+
+            if (!testCase.HasTraits())
+            {
+                return log (!Exclude);
+            }
+
+            if (testCase.TryGetTrait(SelectorName!, out var values))
```
premun, line 107: Should this `!` be handled?

mandel-macaque: Good point, will add that, was added because we do not know if the client will have the nullability enabled.

src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitFilter.cs
mandel-macaque: Is a pattern that I inherited.. is not nice, but I did not want to make more changes, is like doing:

```chsarp
if (logFilteredTests)
{
    log.WriteLine ("My log msg");
}
return excluded;
```

mainly, so that we do not need to write that if all the time.

src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/ITestCaseExtenstions.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.XHarness.Tests.Runners.Xunit
+{
+    /// <summary>
+    /// Useful extensions that make working with the ITestCase interface nicer within the runner.
+    /// </summary>
+    public static class ITestCaseExtenstions
```
akoeplinger, line 16: typo in class+filename

mandel-macaque: he, I'm the master of typos.. even when I pay attention, thx

src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/TestCaseExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Microsoft.DotNet.XHarness.Tests.Runners.Xunit
+{
+    /// <summary>
+    /// Useful extensions that make working with the ITestCase interface nicer within the runner.
+    /// </summary>
+    public static class TestCaseExtensions
+    {
+        /// <summary>
+        /// Returns if the test case does have traits.
```
MattGal, line 19: Suggest "Returns boolean indicating whether..."... it returns every time.

src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/TestCaseExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+                                       [NotNullWhen(true)] out List<string>? values,
+                                       StringComparison comparer = StringComparison.InvariantCultureIgnoreCase)
+        {
+            if (trait == null)
+            {
+                values = null;
+                return false;
+            }
+
+            // there is no guarantee that the dict created by xunit is case insensitive, therefor, trygetvalue might
```
MattGal, line 37: therefore, TryGetValue()...

tests/Microsoft.DotNet.XHarness.Tests.Runners.Tests/XUnitFiltersCollectionTests.cs
```diff
@@ -0,0 +1,238 @@
...
+                    testCase.Setup(t => t.DisplayName).Returns(testDisplayName);
+
+                    yield return new object[]
+                    {
+                        collection,
+                        testCase.Object,
+                        true,
+                    };
+
+                    // single filter that includes
```
MattGal, line 43: includes ... ?

src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/TestCaseExtensions.cs
premun: @MattGal GitHub (AzDO now as well actually) has these handy suggestions that could make fixing these faster ^^

```suggestion
            // there is no guarantee that the dict created by xunit is case insensitive, therefore, trygetvalue might
```

premun: image

PR #147, opened 2020-05-04
Revert "Disable signing validation to unblock the pipeline (#99)"

premun:
This reverts commit 08b0461a25241aa77124185d58cc935f2a9f66bd.

Looks like it was fixed already: https://github.com/dotnet/arcade/issues/5328

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -153,9 +153,6 @@ stages:
 - ${{ if eq(variables._RunAsInternal, True) }}:
   - template: eng\common\templates\post-build\post-build.yml
     parameters:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -153,9 +153,6 @@ stages:
 - ${{ if eq(variables._RunAsInternal, True) }}:
   - template: eng\common\templates\post-build\post-build.yml
     parameters:
-      # Disabling signing validation until we can get the issues sorted. This should be re-enabled
-      # after https://github.com/dotnet/arcade/issues/5328 is resolved.
-      enableSigningValidation: false
```
premun, line 158: Should we also enable the source link validation below?

PR #149, opened 2020-05-04
[iOS] Fix getting information about 32 bit support

premun:
I was able to build the app bundle inside the runtime repo and it was missing the `UIRequiredDeviceCapabilities` property which made XHarness crash


premun:
@EgorBo please do not update XHarness CLI in runtime before this gets in. It will break because there is a missing property in the `Info.plist` files you put in the bundles.

I am not sure whether it should be added by you or is not necessary. @mandel-macaque might know what's the right thing to do?

mandel-macaque:
AFAIK it should be present so that we know the arch, but is good to say, if not present, we do not support 32b, which is the old arch and is going to be remove at some point

PR #157, opened 2020-05-06
Allow the verbosity to be passed without a value

premun:
Fixes: https://github.com/dotnet/xharness/issues/154

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/XHarnessCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/XHarnessCommandArguments.cs
@@ -68,6 +68,19 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments
         /// <returns>Parsed enum value</returns>
         protected static TEnum ParseArgument<TEnum>(string argumentName, string? value, params TEnum[]? invalidValues) where TEnum : struct, IConvertible
         {
+            if (value == null)
+            {
+                throw new ArgumentNullException(message: $"Empty value supplied to {argumentName}", null);
+            }
+
+            if (value.All(c => char.IsDigit(c)))
+            {
+                // Any into would parse into enum successfully, so we forbid that
+                throw new ArgumentException(
+                    $"Invalid value '{value}' supplied for {argumentName}. " +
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/XHarnessCommandArguments.cs
```diff
@@ -66,6 +68,19 @@ protected static string RootPath(string path)
...
         protected static TEnum ParseArgument<TEnum>(string argumentName, string? value, params TEnum[]? invalidValues) where TEnum : struct, IConvertible
         {
+            if (value == null)
+            {
+                throw new ArgumentNullException(message: $"Empty value supplied to {argumentName}", null);
+            }
+
+            if (value.All(c => char.IsDigit(c)))
+            {
+                // Any into would parse into enum successfully, so we forbid that
```
akoeplinger, line 78:
```suggestion
                // Any int would parse into enum successfully, so we forbid that
```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/XHarnessCommandArguments.cs
```diff
@@ -28,7 +28,9 @@ public OptionSet GetOptions()
         {
             var options = GetCommandOptions();
 
-            options.Add("verbosity=|v=", "Verbosity level (1-6) where higher means less logging. (default = 2 / Information)", v => Verbosity = ParseArgument<LogLevel>("verbosity", v));
+            options.Add("verbosity:|v:", "Verbosity level - defaults to 'Information'. If passed without value, 'Debug' is assumed (highest)",
```
akoeplinger, line 31: The wording around defaults is confusing to me, I assume it means that without -v it uses Info, but with -v it uses Debug (unless you pass an explicit value)?

How about this instead:
```suggestion
            options.Add("verbosity:|v:", "Verbosity level - defaults to 'Information' if not specified. If passed without value, 'Debug' is assumed (highest)",
```

premun: Yes, you're completely right. Thanks for the feedback!

Issue #156, opened 2020-05-06
[Runners] Remove ActiveIssue workaround.


mandel-macaque:
The active issue was not ignored due to a bug in its discoverer, a fix was landed in arcade and a workaround in the runners. The workaround should be remove as soon as we have the correct package bump.

PR #158, opened 2020-05-06
[Runners] Workaround to discover all the active issues.

mandel-macaque:
The ActiveIssue discoverer does have a bug in which it does not add the
traits on Android or iOS. The fix was landed here:
https://github.com/dotnet/arcade/pull/5425 but this commit adds a
workaround. Issue https://github.com/dotnet/xharness/issues/156 to
track the workaround and make sure it is removed.

fixes: https://github.com/dotnet/xharness/issues/53

PS: This time it really fixes the issue ;)

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/TestCaseExtensions.cs
+++ b/src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/TestCaseExtensions.cs
@@ -24,12 +24,14 @@ namespace Microsoft.DotNet.XHarness.Tests.Runners.Xunit
             if (testCase.Traits == null)
                 return;
 
-            // the following is a hack to work around an issue in the extensions that
+            // the following code is a work around an issue in the extensions that
             // was fixed in https://github.com/dotnet/arcade/pull/5425 but until we get the
-            // arcade update, we do the following, check for the discoverer and the add
+            // arcade update, we do the following, check for the discoverer and then add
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/TestCaseExtensions.cs
```diff
@@ -15,13 +19,49 @@ namespace Microsoft.DotNet.XHarness.Tests.Runners.Xunit
     /// </summary>
     public static class TestCaseExtensions
     {
+        static void DiscoverAllTraits(this ITestCase testCase)
+        {
+            if (testCase.Traits == null)
+                return;
+
+            // the following is a hack to work around an issue in the extensions that
```
MattGal, line 27: Don't say hack, say workaround, or policheck will flag this.

src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/TestCaseExtensions.cs
```diff
@@ -15,13 +19,49 @@ namespace Microsoft.DotNet.XHarness.Tests.Runners.Xunit
...
+            // was fixed in https://github.com/dotnet/arcade/pull/5425 but until we get the
+            // arcade update, we do the following, check for the discoverer and the add
```
MattGal, line 29: we do the following:

PR #160, opened 2020-05-06
Changes needed to deal with arch issues

MattGal:
Today I learned that our emulators are set up x86 only while Runtime team has been primarily been building for x86-x64.  Once we start using real devices we'll need arm64-v8a apps to match.

This change makes it so either the user can specify the arch to run on (--device-arch) and get an error if no such device is available, _or_ leave this out and have it be guessed from the contents of the /lib/ folder of the apk itself.  

Also handles the slightly different non-error failure message from uninstalling an app that isn't there on an emulator versus real device.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -307,7 +307,7 @@ namespace Microsoft.DotNet.XHarness.Android
 
         public Dictionary<string, string?> GetAttachedDevicesAndArchitectures()
         {
-            Dictionary<string, string?> devicesAndArchitectures = new Dictionary<string, string>();
+            Dictionary<string, string?> devicesAndArchitectures = new Dictionary<string, string?>();
 
             var result = RunAdbCommand("devices -l");
             string standardOutput = result.StandardOutput;
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -51,7 +51,7 @@ namespace Microsoft.DotNet.XHarness.Android
             }
         }
 
-        public void SetActiveDevice(string deviceSerialNumber)
+        public void SetActiveDevice(string? deviceSerialNumber)
         {
             _currentDevice = deviceSerialNumber;
         }
@@ -305,9 +305,9 @@ namespace Microsoft.DotNet.XHarness.Android
             _log.LogDebug($"Copied {filesToCopy.Length} files");
         }
 
-        public Dictionary<string, string> GetAttachedDevicesAndArchitectures()
+        public Dictionary<string, string?> GetAttachedDevicesAndArchitectures()
         {
-            Dictionary<string, string> devicesAndArchitectures = new Dictionary<string, string>();
+            Dictionary<string, string?> devicesAndArchitectures = new Dictionary<string, string>();
 
             var result = RunAdbCommand("devices -l");
             string standardOutput = result.StandardOutput;
@@ -316,22 +316,23 @@ namespace Microsoft.DotNet.XHarness.Android
             int retriesLeft = 5;
             while (retriesLeft-- > 0 && result.ExitCode == (int)AdbExitCodes.SUCCESS && standardOutput == "")
             {
-                result = RunAdbCommand("devices -l");
+                result = RunAdbCommand("devices -l", TimeSpan.FromSeconds(30));
                 standardOutput = result.StandardOutput;
             }
 
             if (result.ExitCode == (int)AdbExitCodes.SUCCESS)
             {
                 string[] lines = standardOutput.Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries);
-                
+
+                // Start at 1 to skip first line, which is always 'List of devices attached'
                 for (int lineNumber = 1; lineNumber < lines.Length; lineNumber++)
                 {
                     _log.LogDebug($"Evaluating line: {lines[lineNumber]}");
                     var lineParts = lines[lineNumber].Split(' ');
-                    if (lineParts.Length > 0)
+                    if (!string.IsNullOrEmpty(lineParts[0]))
                     {
-                        var shellArchitecture = RunAdbCommand($"-s {lineParts[0]} shell getprop ro.product.cpu.abi");
                         var deviceSerial = lineParts[0];
+                        var shellArchitecture = RunAdbCommand($"-s {deviceSerial} shell getprop ro.product.cpu.abi");
 
                         if (shellArchitecture.ExitCode == (int)AdbExitCodes.SUCCESS)
                         {
@@ -340,14 +341,14 @@ namespace Microsoft.DotNet.XHarness.Android
                         else
                         {
                             _log.LogError($"Error trying to get device architecture: {shellArchitecture.StandardError}");
-                            devicesAndArchitectures.Add(deviceSerial, "unknown");
+                            devicesAndArchitectures.Add(deviceSerial, null);
                         }
                     }
                 }
             }
             else
             {
-                _log.LogError($"Error: {result.StandardError}");
+                _log.LogError($"Error: listing attached devices / emulators: {result.StandardError}");
             }
             return devicesAndArchitectures;
         }
@@ -427,7 +428,7 @@ namespace Microsoft.DotNet.XHarness.Android
                 throw new FileNotFoundException($"Provided path for adb.exe was not valid ('{_absoluteAdbExePath}')");
             }
 
-            string deviceSerialArgs = string.IsNullOrEmpty(_currentDevice) ? "" : $"-s {_currentDevice}";
+            string deviceSerialArgs = string.IsNullOrEmpty(_currentDevice) ? string.Empty : $"-s {_currentDevice}";
 
             _log.LogDebug($"Executing command: '{_absoluteAdbExePath} {deviceSerialArgs} {command}'");
 
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -20,6 +20,7 @@ namespace Microsoft.DotNet.XHarness.Android
 
         private readonly string _absoluteAdbExePath;
         private readonly ILogger _log;
+        private string? _currentDevice;
 
         public AdbRunner(ILogger log, string adbExePath = "")
         {
@@ -50,6 +51,11 @@ namespace Microsoft.DotNet.XHarness.Android
             }
         }
 
+        public void SetActiveDevice(string deviceSerialNumber)
+        {
+            _currentDevice = deviceSerialNumber;
+        }
+
         private static string GetCliAdbExePath()
         {
             var currentAssemblyDirectory = Path.GetDirectoryName(typeof(AdbRunner).Assembly.Location);
@@ -65,7 +71,7 @@ namespace Microsoft.DotNet.XHarness.Android
             {
                 return Path.Join(currentAssemblyDirectory, @"../../../runtimes/any/native/adb/macos/adb");
             }
-            throw new NotSupportedException("Cannot determine OS platform being used, thus we can not select ADB executable.");
+            throw new NotSupportedException("Cannot determine OS platform being used, thus we can not select an ADB executable.");
         }
 
         #endregion
@@ -164,7 +170,8 @@ namespace Microsoft.DotNet.XHarness.Android
             {
                 _log.LogInformation($"Successfully uninstalled {apkName}.");
             }
-            else if (result.ExitCode == (int)AdbExitCodes.ADB_UNINSTALL_APP_NOT_ON_DEVICE)
+            else if (result.ExitCode == (int)AdbExitCodes.ADB_UNINSTALL_APP_NOT_ON_DEVICE ||
+                     result.ExitCode == (int)AdbExitCodes.ADB_UNINSTALL_APP_NOT_ON_EMULATOR)
             {
                 _log.LogInformation($"APK '{apkName}' not on device.");
             }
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -298,19 +305,62 @@ public void PushFiles(string localDirectory, string devicePath, bool removeIfPre
...
+            Dictionary<string, string> devicesAndArchitectures = new Dictionary<string, string>();
+
+            var result = RunAdbCommand("devices -l");
+            string standardOutput = result.StandardOutput;
+
+            // If the adb server is dead, this can return 0 exit code and no output; retry a few times if so)
+            int retriesLeft = 5;
+            while (retriesLeft-- > 0 && result.ExitCode == (int)AdbExitCodes.SUCCESS && standardOutput == "")
+            {
+                result = RunAdbCommand("devices -l");
```
premun, line 319: Should there be a timeout? Or is there one inside `RunAdbCommand`?

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -298,19 +305,62 @@ public void PushFiles(string localDirectory, string devicePath, bool removeIfPre
...
+                        {
+                            _log.LogError($"Error trying to get device architecture: {shellArchitecture.StandardError}");
+                            devicesAndArchitectures.Add(deviceSerial, "unknown");
+                        }
+                    }
+                }
+            }
+            else
+            {
+                _log.LogError($"Error: {result.StandardError}");
```
premun, line 350:
```suggestion
                _log.LogError($"Error when listing attached devices: {result.StandardError}");

```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -298,19 +305,62 @@ public void PushFiles(string localDirectory, string devicePath, bool removeIfPre
...
+            {
+                result = RunAdbCommand("devices -l");
+                standardOutput = result.StandardOutput;
+            }
+
+            if (result.ExitCode == (int)AdbExitCodes.SUCCESS)
+            {
+                string[] lines = standardOutput.Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries);
+                
+                for (int lineNumber = 1; lineNumber < lines.Length; lineNumber++)
```
premun, line 327: Can you put a comment why we skip the first line? Is it a table header or something? Or not intentional? :)

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -298,19 +305,62 @@ public void PushFiles(string localDirectory, string devicePath, bool removeIfPre
...
+                {
+                    _log.LogDebug($"Evaluating line: {lines[lineNumber]}");
+                    var lineParts = lines[lineNumber].Split(' ');
+                    if (lineParts.Length > 0)
```
premun, line 331: Will this not be true every time?

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -298,19 +305,62 @@ public void PushFiles(string localDirectory, string devicePath, bool removeIfPre
...
+                    {
+                        var shellArchitecture = RunAdbCommand($"-s {lineParts[0]} shell getprop ro.product.cpu.abi");
+                        var deviceSerial = lineParts[0];
```
premun, line 334:
```suggestion
                        var deviceSerial = lineParts[0];
                        var shellArchitecture = RunAdbCommand($"-s {deviceSerial} shell getprop ro.product.cpu.abi");

```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -298,19 +305,62 @@ public void PushFiles(string localDirectory, string devicePath, bool removeIfPre
...
+
+                        if (shellArchitecture.ExitCode == (int)AdbExitCodes.SUCCESS)
+                        {
+                            devicesAndArchitectures.Add(deviceSerial, shellArchitecture.StandardOutput.Trim());
+                        }
+                        else
+                        {
+                            _log.LogError($"Error trying to get device architecture: {shellArchitecture.StandardError}");
+                            devicesAndArchitectures.Add(deviceSerial, "unknown");
```
premun, line 343: It might be better to have a `Dictionary<string, string?>` and put null where it's unknown and the nullability will later force you to handle that case somewhere in the code rather than having to remember to handle the "unknown" string.

src/Microsoft.DotNet.XHarness.Android/ApkHelper.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+            if (!Path.GetExtension(apkPath).Equals(".apk", StringComparison.OrdinalIgnoreCase))
+            {
+                throw new InvalidOperationException("Only know how to open APK files.");
+            }
+
+            using (ZipArchive archive = ZipFile.Open(apkPath, ZipArchiveMode.Read))
+            {
+                // Enumerate all folders under /lib inside the zip
+                var allLibFolders = archive.Entries.Where(e => e.FullName.StartsWith("lib/"))
```
premun, line 25: I guess this enumerates files in `lib/` as well? Can that be bad somehow?

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -38,6 +39,20 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
             var runner = new AdbRunner(logger);
 
+            // Assumption: APKs we test will only have one arch for now
+            string apkRequiredArchitecture = string.IsNullOrEmpty(_arguments.DeviceArchitecture) ?
+                ApkHelper.GetApkSupportedArchitectures(_arguments.AppPackagePath).FirstOrDefault() :
+                _arguments.DeviceArchitecture;
+
+            if (!string.IsNullOrEmpty(_arguments.DeviceArchitecture))
+            {
+                logger.LogInformation($"Will attempt to run device on specified architecture: '{apkRequiredArchitecture}'");
```
premun, line 49: nit: if we have this `if/else` already, would maybe make sense to fill `apkRequiredArchitecture` in here too. Now we have two different conditional blocks for the same thing

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -377,7 +427,9 @@ private string FormatProcessOutputs((string StandardOutput, string StandardError
                 throw new FileNotFoundException($"Provided path for adb.exe was not valid ('{_absoluteAdbExePath}')");
             }
 
-            _log.LogDebug($"Executing command: '{_absoluteAdbExePath} {command}'");
+            string deviceSerialArgs = string.IsNullOrEmpty(_currentDevice) ? "" : $"-s {_currentDevice}";
```
premun, line 430:
```suggestion
            string deviceSerialArgs = string.IsNullOrEmpty(_currentDevice) ? string.Empty : $"-s {_currentDevice}";

```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -120,6 +143,36 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+        {
+            var allDevicesAndTheirArchitectures = runner.GetAttachedDevicesAndArchitectures();
+            if (allDevicesAndTheirArchitectures.Count == 1)
+            {
+                // There's only one device, so -s argument isn't needed but still check
+                if (!allDevicesAndTheirArchitectures.Values.Contains(apkRequiredArchitecture, StringComparer.InvariantCultureIgnoreCase))
+                {
+                    logger.LogError($"Single device available has architecture '{allDevicesAndTheirArchitectures.Values.First()}', != '{apkRequiredArchitecture}'. Package installation will likely fail, but we'll try anyways."); 
+                }
+                return ""; // Empty = Use default device
```
premun, line 156:
```suggestion
                return string.Empty; // Empty = Use default device
```

premun: But again, I think it would be better to utilize the compiler, allow `string?` on return and return `null` and handle that later. The API of the method then also becomes more intuitive since you can probably guess it returns `null` when no device found and does not for example throw an exception.

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -120,6 +143,36 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+                    return firstAvailableCompatible.Key;
+                }
+                else
+                {
+                    logger.LogError($"No devices found with architecture '{apkRequiredArchitecture}'.  Just returning first available device; installation will likely fail, but we'll try anyways.");
+                    return allDevicesAndTheirArchitectures.Keys.First();
+                }
+            }
+            logger.LogError("Unable to detect an attached device.");
+            return "";
```
premun, line 173: Same as above - `null` would be better

```suggestion
            return string.Empty;
```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -120,6 +143,36 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+                    logger.LogDebug($"Using first-found compatible device - serial: '{firstAvailableCompatible.Key}' - Arch: {firstAvailableCompatible.Value}");
+                    return firstAvailableCompatible.Key;
+                }
+                else
+                {
+                    logger.LogError($"No devices found with architecture '{apkRequiredArchitecture}'.  Just returning first available device; installation will likely fail, but we'll try anyways.");
+                    return allDevicesAndTheirArchitectures.Keys.First();
+                }
+            }
+            logger.LogError("Unable to detect an attached device.");
```
premun, line 172: This hints that there was a failure trying to find a device - maybe there's just none connected so we should say that?

```suggestion
            logger.LogError("No attached device detected");
```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -120,6 +143,36 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
             return Task.FromResult(ExitCode.GENERAL_FAILURE);
         }
 
+        private string GetDeviceToUse(ILogger logger, AdbRunner runner, string apkRequiredArchitecture)
+        {
+            var allDevicesAndTheirArchitectures = runner.GetAttachedDevicesAndArchitectures();
+            if (allDevicesAndTheirArchitectures.Count == 1)
```
premun, line 149: I quite don't understand what is the difference for `1` and `n` devices. We do the same for both - just return the first attached and warn

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -120,6 +143,36 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+            {
+                // There's only one device, so -s argument isn't needed but still check
+                if (!allDevicesAndTheirArchitectures.Values.Contains(apkRequiredArchitecture, StringComparer.InvariantCultureIgnoreCase))
+                {
+                    logger.LogError($"Single device available has architecture '{allDevicesAndTheirArchitectures.Values.First()}', != '{apkRequiredArchitecture}'. Package installation will likely fail, but we'll try anyways.");
```
premun, line 154:
```suggestion
                    logger.LogWarning($"Single device available has architecture '{allDevicesAndTheirArchitectures.Values.First()}' != '{apkRequiredArchitecture}'. Package installation will likely fail, but we'll try anyways."); 
```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -120,6 +143,36 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+                    logger.LogError($"Single device available has architecture '{allDevicesAndTheirArchitectures.Values.First()}', != '{apkRequiredArchitecture}'. Package installation will likely fail, but we'll try anyways."); 
+                }
+                return ""; // Empty = Use default device
+            }
+            else if (allDevicesAndTheirArchitectures.Count > 1)
+            {
+                if (allDevicesAndTheirArchitectures.Any(kvp => kvp.Value.Equals(apkRequiredArchitecture, StringComparison.OrdinalIgnoreCase)))
+                { 
+                    var firstAvailableCompatible = allDevicesAndTheirArchitectures.Where(kvp => kvp.Value.Equals(apkRequiredArchitecture, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
+                    logger.LogDebug($"Using first-found compatible device - serial: '{firstAvailableCompatible.Key}' - Arch: {firstAvailableCompatible.Value}");
```
premun, line 163: I think it might be helpful to say how many other devices were there. I can see how that can help debugging in the future when you have three devices connected, one of them dead and it only detects two but you will never know or something like that.
(just what I also observed as a custom in Xamarin so there might be value in it)

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -120,6 +143,36 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+                    return firstAvailableCompatible.Key;
+                }
+                else
+                {
+                    logger.LogError($"No devices found with architecture '{apkRequiredArchitecture}'.  Just returning first available device; installation will likely fail, but we'll try anyways.");
```
premun, line 168:
```suggestion
                    logger.LogWarning($"No devices found with architecture '{apkRequiredArchitecture}'.  Just returning first available device; installation will likely fail, but we'll try anyways.");
```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -50,6 +51,11 @@ public AdbRunner(ILogger log, string adbExePath = "")
             }
         }
 
+        public void SetActiveDevice(string deviceSerialNumber)
+        {
+            _currentDevice = deviceSerialNumber;
```
mandel-macaque, line 56: check for null?

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
mandel-macaque: We might want to also add a timeout here just in case, similar to what we do in iOS, you never know.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
mandel-macaque: Correct, even with an empty string you will get an array with a single element. Probably > 1?

src/Microsoft.DotNet.XHarness.Android/ApkHelper.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+
+namespace Microsoft.DotNet.XHarness.Android
+{
+    public static class ApkHelper
+    {
+        public static List<string> GetApkSupportedArchitectures(string apkPath)
+        {
```
mandel-macaque, line 12: Check for null.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
premun: Yeah, true, we get a lot of indefinite hangs.

So one timeout in-between the attempts, one for the call itself

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
MattGal: Null is fine, the only place where this is used always calls string.IsNullOrEmpty.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
MattGal: there is a timeout already when not providing an arg, but I can lower it for this one as it should be faster than some.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
MattGal: wow man, I am hurt you'd think I didn't know which indexes are 0 based and not :P  
First line is always `List of devices attached` so skipping it

src/Microsoft.DotNet.XHarness.Android/ApkHelper.cs
MattGal: if you put a file in there there (I didn't find any like this) instead of an arch-named folder, I'd say that was a supported architecture.  I think it's not worth worrying about now;  this is a fallback and ideally the users will tell us what it is before it gets there.

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
MattGal: mostly just because if something went wrong, default has a better chance of just working.  If you send -s <non existent device> somehow you timeout.  I guess you're right.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
premun: But I know it was intentional, just it helps to explain that in the code because it is not obvious

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
MattGal: hrm.  OK.

premun: I think these shouldn't be errors because if we manage to run everything on a "wrong" device, we might still fail the build for someone by outputting errors

PR #161, opened 2020-05-07
[iOS] Show run log in stdout in verbose mode

premun:
This will start logging whatever is in the run log (`run-Simulator_iOS64.log`) to the stdout to get more details upfront.

Egor was asking for this behavior and it is a good point that this would be expected.


premun:
@mandel-macaque what do you think about this?

Also please don't merge, has to be tested that no `FullPath` will crash everything

premun:
Closing and re-opening after I changed the pipeline

PR #163, opened 2020-05-07
[iOS] Add iOS integration tests build stage

premun:
This adds a new stage that runs for public builds and tests our tool E2E on an iPhone Simulator.

1. We publish the nupkg with the XHarness tool as an artifact (compiled in Debug mode)
2. The integration tests stage sends a Helix job to the staging `OSX.1015.Amd64.iOS.Open` queue 
3. The job downloads the nupkg, installs the dotnet tool locally and a gets a test app from Azure storage account, then runs it on a 64 bit Simulator

There are few things where we could improve things but I think we need this asap to have some sort of a validation


premun:
Note to self: if we update projects to `net5.0` we will need to change the channel to `release/5.0.1xx-preview4` or something similar (probably not `master` though). Overview of values is here: https://github.com/dotnet/sdk/branches/all?query=release

premun:
@MattGal do you think you could have a look at this?

azure-pipelines.yml
```diff
@@ -150,12 +178,44 @@ stages:
...
+    jobs:
+    - job: integration_tests
+      displayName: iOS Integration Tests
+      pool:
+        name: Hosted macOS
+      steps:
+      - download: current
+        artifact: Microsoft.DotNet.XHarness.CLI.OSX.IntegrationTests.Debug
+
+      - template: /eng/common/templates/steps/send-to-helix.yml
```
MattGal, line 194: This is good for now, but we should get this into the Arcade Repo and consume the Helix SDK XHarness support (if you give me a functioning app and the args for it I can drive that)

tests/integration-tests/helix-payload/osx-helix-payload.sh
```diff
@@ -0,0 +1,46 @@
...
+    rm -rf "$path/Microsoft.DotNet.XHarness.CLI/$version"
+done <<< "$cache_dirs"
+
+set -x
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# TODO - Package the app bundle here using `dotnet xharness ios package ..`
+# For now we download a pre-bundled app:
+curl -L --output $here/app.zip "https://netcorenativeassets.blob.core.windows.net/resource-packages/external/macos/test-ios-app/System.Numerics.Vectors.Tests.app.zip?sp=r&st=2020-05-04T13:23:20Z&se=2028-05-04T21:23:20Z&spr=https&sv=2019-10-10&sr=b&sig=7sKBDMZrlk%2FA58zlbaUYptb98kK7EacQpmJ9RxlLLrE%3D"
```
MattGal, line 22: SAS token is unnecessary on this URL (resource-packages is a publicly readable container)

tests/integration-tests/helix-payload/osx-helix-payload.sh
```diff
@@ -0,0 +1,46 @@
...
+
+set -e
+
+version='1.0.0-ci'
+
+# Clean the NuGet cache from the previous 1.0.0-ci version of the tool
+# TODO: This might have a better solution: https://github.com/dotnet/xharness/issues/123
+echo "Cleaning the NuGet cache from the previous version of the tool..."
+cache_dirs=`dotnet nuget locals All -l | cut -d':' -f 2 | tr -d ' '`
+while IFS= read -r path; do
```
MattGal, line 11: Remind me; did you try `dotnet nuget locals all --clear` and it didn't work?  Seems much safer to let nuget do the deleting.

azure-pipelines.yml
premun: I agree. As you say, for now it's ok since the value I am after is to verify that the machines I have are running XHarness fine

tests/integration-tests/helix-payload/osx-helix-payload.sh
premun: I didn't want to do that because we will share OSX machines with other workloads and we would be purging the cache for everyone so that might be undesirable

premun:
@MattGal I made this work with .net 5.0. There were few breaking changes in the SDK behavior but nothing major - only downside is that in `azure-pipelines.yaml` we have to supply the same version as is set in `global.json` by Arcade when calling `send-to-helix` but this is just temporary and should go away when we switch to Helix SDK.

There is another issue though - I renamed the xunit XML to `testResults.xml` and the tests didn't appear in AzDO:
https://dev.azure.com/dnceng/public/_build/results?buildId=640111&view=results
and logs for that:
https://helix.int-dot.net/api/2019-06-17/jobs/ba3efbd9-165a-4e68-bc46-65367244dc4a/workitems/WorkItem/

But again, I don't think it's a big deal because this will change again. However, you made a good point that we should verify there are at least some results and I added a check for that so I guess we are good for the time being

PR #165, opened 2020-05-07
Adds support for skipping tests when SkipTestException is raised

steveisok:
If a test fails and the exception type is SkipTestException, then the runner will flag that as a skipped test.  This is to support tests that depend on something in the xunit pipeline that may not be supported on mobile.  We don't want to have to go into every test and mark ConditionalFact/Theory.


mandel-macaque:
test failure looks unrelated, lets re-run to see if it is flacky.

PR #169, opened 2020-05-11
Make XmlVersion.xUnit default

MattGal:
Currently Runtime teams and ADO reporting all uses XUnit, so make this a default.  Eventually we'll clean the duplicated enums up and support this from cmd line


MattGal:
if this builds on mac it'll build on Windows, just merging here...

PR #175, opened 2020-05-12
Move WaitForDevice() above GetDeviceToUse()

MattGal:
…  otherwise we may not be able to get a value for architecture used.


MattGal:
Change has nothing to do with Helix issue on Integration tests, so merging.

PR #178, opened 2020-05-13
Exit HandleTestFailed when SkipTestException is handled

safern:
cc: @steveisok @akoeplinger

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitTestRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitTestRunner.cs
@@ -225,6 +225,7 @@ namespace Microsoft.DotNet.XHarness.Tests.Runners.Xunit
             if (exceptionType == "Microsoft.DotNet.XUnitExtensions.SkipTestException")
             {
                 RaiseTestSkippedCase(args.Message, args.Message.TestCases, args.Message.TestCase);
+                return;
             }
 
             FailedTests++;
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitTestRunner.cs
```diff
@@ -225,6 +225,7 @@ void HandleTestFailed(MessageHandlerArgs<ITestFailed> args)
             if (exceptionType == "Microsoft.DotNet.XUnitExtensions.SkipTestException")
             {
                 RaiseTestSkippedCase(args.Message, args.Message.TestCases, args.Message.TestCase);
+                return;
```
mandel-macaque, line 228: I think we got this if statement wrong the first time, we might was well fix it now. The interface is defined as:
```csharp
public interface IFailureInformation
    {
        /// <summary>
        /// The fully-qualified type name of the exceptions.
        /// </summary>
        string[] ExceptionTypes { get; }
```
So the if statement is just checking if the first exception is the expected one Looks like we should do an any with the exception type, and then call RaiseTestSkippedCase and return.

I know is out of the scope for what the PR fixes, but we might as well do it now.

safern: While I agree with your point, the intent of this feature is that SkipTestException is the only exception thrown as usually we throw it early on the test execution and/or remote executor, so I don’t expect this to have multiple exceptions when a test throws SkipTestException.

Anyway I don’t see a problem with doing Any instead.

safern:
I don't have write permissions, could someone that does merge please?

mandel-macaque:
CI issues can be ignored since they are related to E2E that is being worked by @premun

Issue #144, opened 2020-05-01
Support running single tests 


steveisok:
In working w/ xharness, I was trying to run a single test in the same way we do on desktop.  I noticed this feature wasn't support in the way I would expect.  

Most / all of our interaction w/ xharness is going to be triggered via msbuild.  So, given this property, I would like xharness to be able to run a single test:

`/p:XunitMethodName={FullyQualifiedNamespace}.{ClassName}.{TestMethodName}`

mandel-macaque:
@steveisok first step towards that dir: https://github.com/dotnet/xharness/pull/146

safern:
> /p:XunitMethodName={FullyQualifiedNamespace}.{ClassName}.{TestMethodName}

Note that in runtime we transform this to a xunit parameter, `-method {FullyQualifiedNamespace}.{ClassName}.{TestMethodName}`

mandel-macaque:
@steveisok before I get to this, should we just use the format that @safern mentioned? you use case is just for a developer using locally, correct?

safern:
Right the use case is for local debugging or testing to avoid running the whole test suite. This is supported by xunit via the `-method` or `-class` argument. In dotnet/runtime we provide an MSBuild hook via properties as @steveisok mentioned above:

https://github.com/dotnet/runtime/blob/9b5805f9dacee96bac80b2ab10e45e02a9369347/eng/testing/xunit/xunit.console.targets#L19-L20

https://github.com/xunit/xunit/blob/7485b9b3819059dfa810cb5434ebed724be87686/src/xunit.console/ConsoleRunner.cs#L248-L257

mandel-macaque:
Then I think is best to add support for -method and -class, which is easy, and then the msbuild conversion can be reused.

PR #181, opened 2020-05-13
[Runners] Allow to specify to the test app not to run all tests.

mandel-macaque:
This is a step fwd allowing user to run just those tests that match a
filter or filters. To support that we first want to be able to ignore
all tests and later provide the inclusion filters.

This is related to issue https://github.com/dotnet/xharness/issues/144

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
+++ b/src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
@@ -1,4 +1,5 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+﻿#nullable enable
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
--- a/src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
+++ b/src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
@@ -38,7 +38,7 @@ namespace Microsoft.DotNet.XHarness.Tests.Runners
             var runner = await InternalRunAsync(logger);
             ConfigureRunner(runner, options);
 
-            TextWriter writer = null;
+            TextWriter? writer = null;
 
             if (options.EnableXml)
             {
--- a/src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
+++ b/src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
@@ -34,29 +34,23 @@ namespace Microsoft.DotNet.XHarness.Tests.Runners
             // we will write the normal console output using the LogWriter
             var logger = (Logger == null || options.EnableXml) ? new LogWriter(Device) : new LogWriter(Device, Logger);
             logger.MinimumLogLevel = MinimumLogLevel.Info;
-            var testAssemblies = GetTestAssemblies();
 
             var runner = await InternalRunAsync(logger);
+            ConfigureRunner(runner, options);
 
-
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.XHarness.Tests.Runners/ApplicationEntryPoint.cs
```diff
@@ -163,6 +165,30 @@ private async Task<List<string>> GetIgnoredCategories()
...
+
+        internal static string WriteResults(TestRunner runner, ApplicationOptions options, LogWriter logger, TextWriter writer = null)
+        {
+            TestRunner.Jargon jargon = options.XmlVersion.ToTestRunnerJargon();
+            string resultsFilePath = null;
+
+            if (options.EnableXml)
+            {
+                runner.WriteResultsToFile(writer ?? Console.Out, jargon);
+                logger.Info("Xml file was written to the tcp listener.");
```
premun, line 181: Should it say "to console"? I think this might help is with investigations

mandel-macaque: Will do

src/Microsoft.DotNet.XHarness.Tests.Runners/AndroidApplicationEntryPoint.cs
```diff
@@ -34,30 +34,24 @@ public override async Task RunAsync()
...
             var logger = (Logger == null || options.EnableXml) ? new LogWriter(Device) : new LogWriter(Device, Logger);
             logger.MinimumLogLevel = MinimumLogLevel.Info;
-            var testAssemblies = GetTestAssemblies();
 
             var runner = await InternalRunAsync(logger);
+            ConfigureRunner(runner, options);
 
-
-            TestRunner.Jargon jargon = options.XmlVersion.ToTestRunnerJargon();
+            TextWriter writer = null;
```
premun, line 41: Not sure we are doing that here, but I am starting to get used to it 😅

```suggestion
            TextWriter? writer = null;
```

src/Microsoft.DotNet.XHarness.Tests.Runners/ApplicationEntryPoint.cs
```diff
@@ -163,6 +165,30 @@ private async Task<List<string>> GetIgnoredCategories()
...
+            }
+            else
+            {
+                resultsFilePath = runner.WriteResultsToFile(jargon);
+                logger.Info($"Xml result can be found {resultsFilePath}");
```
premun, line 186:
```suggestion
                logger.Info($"XML results can be found in '{resultsFilePath}'");
```

src/Microsoft.DotNet.XHarness.Tests.Runners/ApplicationOptions.cs
```diff
@@ -82,6 +84,12 @@ public ApplicationOptions()
                 { "xmlversion", "The xml version.", v => XmlVersion = (XmlVersion) Enum.Parse (typeof (XmlVersion), v, false) },
                 { "logfile=", "A path where output will be saved.", v => LogFile = v },
                 { "result=", "The path to be used to store the result", v => ResultFile = v},
+                { "run-all-tests:", "Run all the tests fund in the assembly. Defaults to true.", v =>
```
premun, line 87:
```suggestion
                { "run-all-tests:", "Run all the tests found in the assembly. Defaults to true.", v =>
```

The `:` belongs here, I know 😁

mandel-macaque:
Integration tests can be ignored.

akoeplinger:
TBH this option doesn't make much sense to me. We need to provide a way to specify the single test name which will be run anyway and that can be used as a key to decide whether to run all tests or not.

The more I think about it the more I feel we should just provide a way to pass "arbitrary" xunit console options similar to how we use the xunit-excludes.txt file rather than adding one-by-one options to ApplicationEntrypoint.

premun:
@mandel-macaque ^^

mandel-macaque:
@akoeplinger it was a first step to the next PR ;)

PR #182, opened 2020-05-13
Rewrite the SimulatorInstaller tool

premun:
The tool is quite simple, yet the code was pretty much spaghetti somehow. I have rewritten it to follow the same Command/Argument structure as the CLI project and also changed the CLI to be the same as rest of dotnet so that now we have these commands:
- `install` - Installs given simulators
- `list` - Lists available simulators
- `find` - Finds whether given simulators are installed and outputs list of missing ones (returns 0 when none missing)
- `help`

I added the `help`:
```
> dotnet simulator-installer help install
usage: install [OPTIONS]

Installs given simulators

  -s, --simulator=VALUE      ID of the Simulator to install. Repeat multiple
                               times to define more
      --force                Install again even if already installed
      --xcode=VALUE          Path to where Xcode is located, e.g. /Application/
                               Xcode114.app. If not set, xcode-select is used
                               to determine the location
      --verbosity, -v[=VALUE]
                             Verbosity level - defaults to 'Information' if not
                               specified. If passed without value, 'Debug' is
                               assumed (highest)
      --help, -h
```

- Fixed the `--only-check`/`find` mode which got broken because of the Loggers
- Stopped requiring the `--xcode` parameter since we can use `xcode-select` from `ProcessManager`
- Enabled nullability for `ProcessManager`


tools/Microsoft.DotNet.XHarness.SimulatorInstaller/Microsoft.DotNet.XHarness.SimulatorInstaller.csproj
```diff
@@ -24,4 +24,10 @@
     <ProjectReference Include="..\..\src\Microsoft.DotNet.XHarness.iOS.Shared\Microsoft.DotNet.XHarness.iOS.Shared.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
premun, line 27: This will be replaced with a reference to the shared library

src/Microsoft.DotNet.XHarness.iOS.Shared/Execution/IProcessManager.cs
```diff
@@ -27,15 +28,15 @@ public interface IProcessManager
...
+        Task<ProcessExecutionResult> RunAsync(Process process, ILog log, TimeSpan? timeout = null, Dictionary<string, string>? environmentVariables = null, CancellationToken? cancellationToken = null, bool? diagnostics = null);
+        Task<ProcessExecutionResult> RunAsync(Process process, MlaunchArguments args, ILog log, TimeSpan? timeout = null, Dictionary<string, string>? environmentVariables = null, CancellationToken? cancellationToken = null, bool? diagnostics = null);
+        Task<ProcessExecutionResult> RunAsync(Process process, MlaunchArguments args, ILog log, ILog stdoutLog, ILog stderrLog, TimeSpan? timeout = null, Dictionary<string, string>? environmentVariables = null, CancellationToken? cancellationToken = null, bool? diagnostics = null);
+        Task<ProcessExecutionResult> RunAsync(Process process, ILog log, ILog stdoutLog, ILog stderrLog, TimeSpan? timeout = null, Dictionary<string, string>? environmentVariables = null, CancellationToken? cancellationToken = null, bool? diagnostics = null);
```
mandel-macaque, line 39: This changed in the Common lib, I believe I did the same changes.

src/Microsoft.DotNet.XHarness.iOS.Shared/Execution/ProcessManager.cs
```diff
@@ -16,17 +16,18 @@
 using Microsoft.DotNet.XHarness.iOS.Shared.Utilities;
 using Microsoft.DotNet.XHarness.iOS.Shared.Execution.Mlaunch;
 
+#nullable enable
 namespace Microsoft.DotNet.XHarness.iOS.Shared.Execution
 {
     public class ProcessManager : IProcessManager
```
mandel-macaque, line 22: Same, we will have conficts.

tools/Microsoft.DotNet.XHarness.SimulatorInstaller/Commands/SimulatorInstallerCommand.cs
```diff
@@ -0,0 +1,238 @@
...
+                var versionMinor = versions[1];
+                var dict = new Dictionary<string, string>() {
+                    { "DOWNLOADABLE_VERSION_MAJOR", versionMajor },
+                    { "DOWNLOADABLE_VERSION_MINOR", versionMinor },
+                    { "DOWNLOADABLE_VERSION", version },
+                };
+
+                var identifier = Replace(identifierNode.InnerText, dict);
+
+                dict.Add("DOWNLOADABLE_IDENTIFIER", identifier);
```
mandel-macaque, line 120: nit, lets use constants.

tools/Microsoft.DotNet.XHarness.SimulatorInstaller/Commands/SimulatorInstallerCommand.cs
```diff
@@ -0,0 +1,238 @@
...
+            var lines = pkgInfo.Split('\n');
+            var version = lines.First(v => v.StartsWith("version: ", StringComparison.Ordinal)).Substring("version: ".Length);
+            return Version.Parse(version);
+        }
+
+        private async Task<string?> GetSimulatorIndexXml()
+        {
+            var (xcodeVersion, xcodeUuid) = await GetXcodeInformation();
+
+            var url = $"https://devimages-cdn.apple.com/downloads/xcode/simulators/index-{xcodeVersion}-{xcodeUuid}.dvtdownloadableindex";
```
mandel-macaque, line 153: nit: lets move it to a constant. Will the change the url.. I can believe anything,

PR #185, opened 2020-05-13
Address https://github.com/dotnet/core-eng/issues/9820 

MattGal:
… and some minor timing issues noted while debugging it. 

Now always uses device serial when addressing devices because we can't call various APIs (e.g. "wait-for-device" in a multi-device space without it.


MattGal:
I'll hunt down the iOS integration test thing now.

PR #186, opened 2020-05-13
Work around https://github.com/dotnet/core-eng/issues/9824

MattGal:


markwilkie:
This makes me sad.  Thanks for digging in @MattGal

MattGal:
> This makes me sad. Thanks for digging in @MattGal

People have been prematurely declaring the death of hard coded sleeps as long as I've been in tha' game.

premun:
What would Matt Mitchell say when he saw we are slowing builds down on purpose

PR #187, opened 2020-05-13
Revert "Work around https://github.com/dotnet/core-eng/issues/9824"

mmitche:
Reverts dotnet/xharness#186

REVERT HORRIBLE PR.


MattGal:
You're quite right it was horrible, I was confused because the ADO API is different from 5.0-preview to 5.0.  All sorted now, mea culpa for the horribleness.

PR #188, opened 2020-05-13
Remove hard coded sleep and set_BuildConfig 

MattGal:
… so if this comes back we'll have a binlog.

The problem was in the GH repo itself, along with AzDO changing their response for when giving secrets to forks is turned off


MattGal:
All succeeded, going to merge now.

PR #189, opened 2020-05-13
Try to wait for online devices

MattGal:


MattGal:
> Approving, but maybe adding a param for those timeouts/sleeps would make this easier if the times change in CI.

Most likely these times would not want to be configurable rather they might just want to be basically infinite.  I thought it'd be OK on the devices to continue if they run out of but if we can't run a simple command on them and have them online it's probably a "let it time out trying" scenario.

Issue #177, opened 2020-05-13
If instrumentation crashes xharness exits with code 0


safern:
For example, I have an instrumentation that crashed:

```
05-12 17:57:09.386  2027  4471 I ActivityManager: Process net.dot.System.Runtime.Tests (pid 13396) has died: fore FGS 
05-12 17:57:09.386  1781  1781 I Zygote  : Process 13396 exited due to signal 11 (Segmentation fault)
05-12 17:57:09.387  2027  2058 I libprocessgroup: Successfully killed process cgroup uid 10148 pid 13396 in 0ms
05-12 17:57:09.387  2027  4471 W ActivityManager: Crash of app net.dot.System.Runtime.Tests running instrumentation ComponentInfo{net.dot.System.Runtime.Tests/net.dot.MonoRunner}
```

and this is the Xharness output:
```
  info: test[0]
        Running instrumentation class net.dot.MonoRunner took 54.0933972 seconds
  info: test[0]
        Wrote current ADB log to /Users/santifdezm/repos/runtime/artifacts/bin/System.Runtime.Tests/net5.0-Unix-Debug/android-x86/AppBundle/TestResults/adb-logcat-net.dot.System.Runtime.Tests.log
  info: test[0]
        Attempting to remove apk 'net.dot.System.Runtime.Tests': 
  info: test[0]
        Successfully uninstalled net.dot.System.Runtime.Tests.
  XHarness exit code: 0
```

cc: @MattGal @EgorBo @steveisok @akoeplinger

MattGal:
Discussed offline, there is a result property bag for exit code set but this is only if that code runs;  XHarness already will fail for any non-0 exit code from the instrumentation itself, so we'll chat about solutions in the AM but this likely is something that has to be fixed in the instrumentation class.

safern:
I think this is different than what we discussed. Here the Instrumentation crashed, so we didn't exit, if you look at the logs on a crash there is not even a exit code.

safern:
So @MattGal and I took a deeper look at this and it seems like when there is an unhandled exception the test runner we have handles it and calls `exit(1)`:

https://github.com/dotnet/runtime/blob/5859dadf20a9dae44624e0a79e48f1b3f1f26a0a/src/mono/msbuild/AndroidAppBuilder/Templates/runtime-android.c#L111-L125

[...]

PR #190, opened 2020-05-14
Address https://github.com/dotnet/xharness/issues/177

MattGal:
Still testing this out but this should solve https://github.com/dotnet/xharness/issues/177 and make it obvious to users when things crash.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
@@ -10,7 +10,6 @@ using System.Threading.Tasks;
 using Microsoft.DotNet.XHarness.Android;
 using Microsoft.DotNet.XHarness.CLI.CommandArguments;
 using Microsoft.DotNet.XHarness.CLI.CommandArguments.Android;
-using Microsoft.DotNet.XHarness.iOS.Shared.Logging;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.DotNet.XHarness.CLI.Commands.Android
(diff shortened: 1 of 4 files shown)


MattGal:
image

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -10,23 +10,38 @@
 using Microsoft.DotNet.XHarness.Android;
 using Microsoft.DotNet.XHarness.CLI.CommandArguments;
 using Microsoft.DotNet.XHarness.CLI.CommandArguments.Android;
+using Microsoft.DotNet.XHarness.iOS.Shared.Logging;
```
mandel-macaque, line 13: @MattGal looks like the IDE added a namespace you dont need?

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -117,6 +132,27 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
                         {
                             logger.LogInformation($"Test execution summary:{Environment.NewLine}{resultValues[_testRunSummaryVariableName]}");
                         }
+
+                        // Due to the particulars of how instrumentations work, ADB will report a 0 exit code for crashed instrumentations
+                        // We'll change that to a specific value and print a message explaining why.
+                        if (resultValues.ContainsKey(_returnCodeVariableName))
```
premun, line 138: nit: you could also `TryGetValue` and use it below at :140 and :147

MattGal: IMO that's a personal preference, doesn't really simplify or make more obvious, so I'm going to skip this.

PR #197, opened 2020-05-14
More offline emulator fun

MattGal:
With some help from @akoeplinger and some messing around with systemctl commands I think I've solved the "wait for online' issue.  Also added logging for shortMsg, a magical property bag value that seems to come up when an instrumentation crashes.


MattGal:
Noone around to give an approval, it's just in Android stuff I'm trying to unblock, so I'm going to merge.

PR #202, opened 2020-05-14
[Runners/iOS/Android] Remove extra enums.

mandel-macaque:
Remove the extra enums and use the ones in the common lib.


mandel-macaque:
Yes, common should be as small as possible

mandel-macaque:
Following @premun I remove the dep to common and linked the file, makes more sense.

PR #203, opened 2020-05-15
[iOS] Notify the user when there are issues when connecting to the device.

mandel-macaque:
This issue is very very rare and usually means that there is something
wrong either with the host or the device. The simplest way to fix this
is my rebooting the device. We now provide a nice message about it.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
@@ -24,7 +24,7 @@ namespace Microsoft.DotNet.XHarness.iOS
             ["IncorrectArchitecture"] = "IncorrectArchitecture: Failed to find matching device arch for the application.",
         };
 
-        static bool TryFindErrors (ILog? log, Dictionary<string, string> errorMap, out string? failureMessage)
+        static bool TryFindErrors (ILog log, Dictionary<string, string> errorMap, out string? failureMessage)
         {
             failureMessage = null;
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
```diff
@@ -13,50 +13,52 @@ namespace Microsoft.DotNet.XHarness.iOS
...
+                return false;
+
+            using var reader = log.GetReader ();
+            while (!reader.EndOfStream) {
+                string line = reader.ReadLine ();
+                if (line == null)
+                    continue;
+                //go over errors and return true as soon as we find one that matches
+                foreach (var error in errorMap.Keys) {
+                    if (!line.Contains (error))
```
premun, line 44: InvariantCultureIgnoreCase?

src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
```diff
@@ -13,50 +13,52 @@ namespace Microsoft.DotNet.XHarness.iOS
 {
     public class ErrorKnowledgeBase : IErrorKnowledgeBase
     {
-        const string IncorrectArchPrefix = "IncorrectArchitecture";
+        static readonly Dictionary<string, string> testErrorMaps = new Dictionary<string, string> {
```
premun, line 16: nit: formatting in this file - missing braces and braces on new lines mostly, some missing accessibility modifiers..

PR #204, opened 2020-05-15
[iOS] Allow users to use lldb to debug issues in test applications.

mandel-macaque:
Add support for the lldb feature in mtouch. User can pass a new flag to
state that lldb should be enabled, the CLI will then provide a command to
copy paste to access the lldb session.

The aggregated log was created to state which of the ILogs that was
passed should be considered the default log and therefore the one to
return the FullPath and the StreamReader.

When the flag is pass the user will get the following:

Screenshot 2020-05-14 at 23 27 10

While if the flag was found but not set by the current xharness execution, it will be respected but a warning will be printed:

Screenshot 2020-05-14 at 23 27 56

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/iOS/iOSTestCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/iOS/iOSTestCommandArguments.cs
@@ -126,7 +126,7 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.iOS
             { "xml-jargon=|xj=", $"The xml format to be used in the unit test results. Can be {XmlResultJargon.TouchUnit}, {XmlResultJargon.NUnitV2}, {XmlResultJargon.NUnitV3} or {XmlResultJargon.xUnit}.",
                 v => XmlResultJargon = ParseArgument("xml-jargon", v, invalidValues: XmlResultJargon.Missing)
             },
-            { "enable-lldb", "Allow to debug the launched application usinghe lldb.", v => EnableLldb = v != null },
+            { "enable-lldb", "Allow to debug the launched application using lldb.", v => EnableLldb = v != null },
         };
 
         public override void Validate()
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/iOS/iOSTestCommandArguments.cs
```diff
@@ -121,6 +126,7 @@ protected set
             { "xml-jargon=|xj=", $"The xml format to be used in the unit test results. Can be {XmlResultJargon.TouchUnit}, {XmlResultJargon.NUnitV2}, {XmlResultJargon.NUnitV3} or {XmlResultJargon.xUnit}.",
                 v => XmlResultJargon = ParseArgument("xml-jargon", v, invalidValues: XmlResultJargon.Missing)
             },
+            { "enable-lldb", "Allow to debug the launched application usinghe lldb.", v => EnableLldb = v != null },
```
safern, line 129: typo: `usinghe` -> `using`

src/Microsoft.DotNet.XHarness.iOS.Shared/Logging/AggregatedLog.cs
```diff
@@ -3,15 +3,23 @@
...
 using Microsoft.DotNet.XHarness.Common.Logging;
 
+#nullable enable
 namespace Microsoft.DotNet.XHarness.iOS.Shared.Logging
 {
 
     public abstract partial class Log
     {
 
+        public static ILog CreateAggregatedLogWithDefault(ILog defaultLog, params ILog[] logs)
```
premun, line 17: Yeah, I guess this is our way out!

premun, line 17: I think it would maybe make sense to call it `AppendLog`? And use the base one as the default one?

mandel-macaque: I was thinking about how to fix this mess. I have a plan. I'd say, lets merge this, then I'll create a PR with a proper fix and will move all the stuff. Specially since we do not yet consume this from xamarin-macios, we can change it and make it properly.

PR #208, opened 2020-05-15
[iOS] Fix the mess we had with the logs.

mandel-macaque:
The initial implementation of the logs, yet useful, it was wrong. It
made it very easy to pass Logs that do have certain missing properties
and methods to other classes that used those methods, resulting in
runtime issues.

We move to have different interfaces that will allow to say which is the
min amount of methods to be implemented (ILog) or the max one
(IFileBackedLog) so that the compiler will help us to not pass a
MemoryLog or a ConsoleLog to those methods that will like to use, for
example, the FullPath of the log.

This might look like a breaking change for xamarin-macios, but whatever
it breaks was a hidden bug.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Common/Logging/ILog.cs
+++ b/src/Microsoft.DotNet.XHarness.Common/Logging/ILog.cs
@@ -13,15 +13,29 @@ namespace Microsoft.DotNet.XHarness.Common.Logging
     {
         string Description { get; set; }
         bool Timestamp { get; set; }
-        string FullPath { get; }
         Encoding Encoding { get; }
-
         void Write(byte[] buffer, int offset, int count);
-        StreamReader GetReader();
         void Write(string value);
         void WriteLine(string value);
         void WriteLine(StringBuilder value);
         void WriteLine(string format, params object[] args);
         void Flush();
+    }
+
+    /// <summary>
+    /// Interface to be implemented by those loggers that provide a StreamReader to read
+    /// the already added info.
+    /// </summary>
+    public interface IReadable
+    {
+        StreamReader GetReader();
+
+    }
+
+    public interface IReadableLog : ILog, IReadable { }
+
+    public interface IFileBackedLog : IReadableLog
+    {
+        string FullPath { get; }
     }
 }
--- a/src/Microsoft.DotNet.XHarness.Common/Logging/ILog.cs
+++ b/src/Microsoft.DotNet.XHarness.Common/Logging/ILog.cs
@@ -11,7 +11,7 @@ namespace Microsoft.DotNet.XHarness.Common.Logging
 
     public interface ILog : IDisposable
     {
-        string Description { get; set; }
+        string? Description { get; set; }
         bool Timestamp { get; set; }
         Encoding Encoding { get; }
         void Write(byte[] buffer, int offset, int count);
@@ -22,17 +22,10 @@ namespace Microsoft.DotNet.XHarness.Common.Logging
         void Flush();
     }
 
-    /// <summary>
-    /// Interface to be implemented by those loggers that provide a StreamReader to read
-    /// the already added info.
-    /// </summary>
-    public interface IReadable
+    public interface IReadableLog : ILog
     {
         StreamReader GetReader();
-
     }
-
-    public interface IReadableLog : ILog, IReadable { }
 
     public interface IFileBackedLog : IReadableLog
     {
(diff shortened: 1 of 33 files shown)


src/Microsoft.DotNet.XHarness.Common/Logging/ILog.cs
```diff
@@ -13,15 +13,29 @@ public interface ILog : IDisposable
...
         void Flush();
     }
+
+    /// <summary>
+    /// Interface to be implemented by those loggers that provide a StreamReader to read
+    /// the already added info.
+    /// </summary>
+    public interface IReadable
+    {
+        StreamReader GetReader();
```
premun, line 31: Why do we have this `IReadable`? Why not put `GetReader()` straight to `IReadableLog`?

src/Microsoft.DotNet.XHarness.iOS.Shared/Logging/ILogs.cs
```diff
@@ -8,22 +8,22 @@
 
 namespace Microsoft.DotNet.XHarness.iOS.Shared.Logging
 {
-    public interface ILogs : IList<ILog>, IDisposable
+    public interface ILogs : IList<IFileBackedLog>, IDisposable
```
premun, line 11: I guess these are then `IFileBackedLogs`?

premun: Also wondering - why not just `IFileLog`?

src/Microsoft.DotNet.XHarness.iOS.Shared/Logging/ReadableLog.cs
```diff
@@ -0,0 +1,20 @@
+using System.IO;
+using Microsoft.DotNet.XHarness.Common.Logging;
+
+namespace Microsoft.DotNet.XHarness.iOS.Shared.Logging
+{
+    public abstract class ReadableLog : Log, IReadableLog
+    {
+
+        protected ReadableLog(string description = null) : base(description) { }
```
premun, line 9:
```suggestion
        protected ReadableLog(string? description = null) : base(description) { }
```

src/Microsoft.DotNet.XHarness.iOS.Shared/Logging/ReadableLog.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        public abstract StreamReader GetReader();
+    }
+
+    public abstract class FileBackedLog : ReadableLog, IFileBackedLog
+    {
+        protected FileBackedLog(string description = null) : base(description) { }
```
premun, line 16:
```suggestion
        protected FileBackedLog(string? description = null) : base(description) { }
```

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
```diff
@@ -30,14 +30,14 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared
     public class TestReporter : ITestReporter
     {
 
-        const string timeoutMessage = "Test run timed out after {0} minute(s).";
-        const string completionMessage = "Test run completed";
-        const string failureMessage = "Test run failed";
+        const string _timeoutMessage = "Test run timed out after {0} minute(s).";
+        const string _completionMessage = "Test run completed";
+        const string _failureMessage = "Test run failed";
```
premun, line 35:
```suggestion
        private const string _timeoutMessage = "Test run timed out after {0} minute(s).";
        private const string _completionMessage = "Test run completed";
        private const string _failureMessage = "Test run failed";
```

and below

premun: Also I think ConstsShouldBeLikeThis

src/Microsoft.DotNet.XHarness.Common/Logging/ILog.cs
mandel-macaque: Seemed to be a good idea at the time, but know that you mention it, I'll remove that, lets see what happens, no the compiler helps :)

src/Microsoft.DotNet.XHarness.iOS.Shared/Logging/ILogs.cs
mandel-macaque: I tried, less freedom :/

src/Microsoft.DotNet.XHarness.Common/Logging/ILog.cs
mandel-macaque: There are classes that just care about this interface. Gives us freedom.

PR #209, opened 2020-05-18
Add integration tests for SimulatorInstaller

premun:
Adds a new pipeline stage to our PR builds that will run the packaged SimulatorInstaller tool and verify the list and the find command

image


premun:
Thanks! The friendly name is unfortunately hidden in the implementation of the `send-to-helix` template and I cannot change it..

PR #216, opened 2020-05-20
Use TCP over Network for integration tests

premun:


mandel-macaque:
Is there a reason for this change? I have no issues with it, but I'd like to know in case I have to look closer into the tunnel

premun:
> Is there a reason for this change? I have no issues with it, but I'd like to know in case I have to look closer into the tunnel

@mandel-macaque so basically, the short answer is - we should be testing the default here to catch problems with it. I reckon people won't really be setting this option themselves. For example we are starting to look into how people will use it through Arcade SDK as I finished implementing it, and I noticed the difference between how we will call `xharness` there and how we do it for our E2E tests..

What do you think? Should we then make Network the default?

PR #219, opened 2020-05-21
Improve retry and logging around adb devices

MattGal:
I can't repro [this error](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-arcade-refs-heads-prvysoky-xharness-io432a676f47944fefa8/xharness-System.Numerics.Vectors.Tests-x86/console.ca83ad82.log) but I know it's out there.  

My approach is to try longer (not the problem in this case, but this is an absolutely breaking problem for > 1 device machines), print out the outputs in case of failure, and tighten up the condition for device listing. I may introduce retry at a different layer if this doesn't work.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -317,11 +317,12 @@ namespace Microsoft.DotNet.XHarness.Android
             var result = RunAdbCommand("devices -l", TimeSpan.FromSeconds(30));
             string standardOutput = result.StandardOutput;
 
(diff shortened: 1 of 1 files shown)


MattGal:
Interesting Helix API failure I'll check out in App insights, seems to not repro so merging.

premun:
@MattGal staging Helix API was down yesterday night and today. Jakub from our team rolled it back. There should be an issue for that but I cannot seem to find it..

MattGal:
> @MattGal staging Helix API was down yesterday night and today. Jakub from our team rolled it back. There should be an issue for that but I cannot seem to find it..

ah that explains it.  OK no investigation needed then unless Jakub wants help :)

PR #221, opened 2020-05-21
[iOS] Small refactor of the error knowledge base.

mandel-macaque:
Allow to provide a url to a knonw issue so that if we have a problem and
it has been reported the user can easily visit it for more informaiton.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
@@ -14,16 +14,16 @@ namespace Microsoft.DotNet.XHarness.iOS
 {
     public class ErrorKnowledgeBase : IErrorKnowledgeBase
     {
-        private static readonly Dictionary<string, (string HumanMessage, string IssueLink)?> _testErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)?>
+        private static readonly Dictionary<string, (string HumanMessage, string IssueLink)> _testErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)>
         {
             ["Failed to communicate with the device"] = // Known issue but not a failure.
                 (HumanMessage: "Failed to communicate with the device. Please ensure the cable is properly connected, and try rebooting the device", IssueLink: "")
 
         };
 
-        private static readonly Dictionary<string, (string HumanMessage, string IssueLink)?> _buildErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)?>();
+        private static readonly Dictionary<string, (string HumanMessage, string IssueLink)> _buildErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)>();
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
```diff
@@ -14,31 +14,39 @@ namespace Microsoft.DotNet.XHarness.iOS
 {
     public class ErrorKnowledgeBase : IErrorKnowledgeBase
     {
-        private static readonly Dictionary<string, string> _testErrorMaps = new Dictionary<string, string>
+        private static readonly Dictionary<string, (string HumanMessage, string IssueLink)> _testErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)>
```
premun, line 17: nit: at this point it might be worth consideration to define alias for this file for `Dictionary<string, (string ....`

src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
```diff
@@ -14,31 +14,39 @@ namespace Microsoft.DotNet.XHarness.iOS
...
 
-        private static readonly Dictionary<string, string> _buildErrorMaps = new Dictionary<string, string>();
+        private static readonly Dictionary<string, (string HumanMessage, string IssueLink)> _buildErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)>();
 
-        private static readonly Dictionary<string, string> _installErrorMaps = new Dictionary<string, string>
+        private static readonly Dictionary<string, (string HumanMessage, string IssueLink)> _installErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)>
         {
             ["IncorrectArchitecture"] =
-                "IncorrectArchitecture: Failed to find matching device arch for the application."
+                (HumanMessage: "IncorrectArchitecture: Failed to find matching device arch for the application.", IssueLink: "") // known failure, but not an issue
```
premun, line 29: Maybe make `IssueLink` nullable?

src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
mandel-macaque: Believe it or not, I did think about it and I was not sure. The fact that you had the same idea means is a good one. The amount of ? Is that is crazy, but worth it.

src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
mandel-macaque: Same, will do

mandel-macaque:
@premun needed for Xamarin-macios. But if we ever need to, we will.

src/Microsoft.DotNet.XHarness.iOS/ErrorKnowledgeBase.cs
premun: And we will be getting new `!` with C# 9.0! 😁

PR #229, opened 2020-05-26
Improve polling wait loop for devices -l 

MattGal:
I tried this enough times to finally see the output when it finds no devices.   Also address concurrency issue noted occasionally on the process output stringbuilders.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -315,28 +315,29 @@ namespace Microsoft.DotNet.XHarness.Android
             Dictionary<string, string?> devicesAndArchitectures = new Dictionary<string, string?>();
 
             var result = RunAdbCommand("devices -l", TimeSpan.FromSeconds(30));
-            string standardOutput = result.StandardOutput;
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -489,17 +490,27 @@ private string FormatProcessOutputs((string StandardOutput, string StandardError
...
+            Thread.Sleep(1000);
 
-            // If we allow any timespan to be max value, it could cause unusual behavior, force max-int max (still a LOT)
-            // (int.MaxValue ms is about 24 days).  Large values are effectively timeouts for the outer harness
-            if (!p.WaitForExit((int)Math.Min(timeOut.TotalMilliseconds, int.MaxValue)))
+            bool finishedNormally = p.WaitForExit((int)Math.Min(timeOut.TotalMilliseconds, int.MaxValue));
+
+            // Lock the stringbuilders used as rarely this can cause concurrency issues
+            // resulting in "Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'chunkLength')"
+            lock (standardOut)
```
mandel-macaque, line 499: Approving, just a nit: you can do:
```csharp
lock (standardOut)
lock (standardErr)
{
}
```
That way we reduce a little the indentation, but that is only useful for people like me that use terminal editors :)

PR #231, opened 2020-05-27
Install the dotnet tools via --tool-path in our CI tests

### LINKED ISSUE #123: `dotnet tool install` uses machine cache and doesn't install newest bits
I am seeing following behavior:
1. I have the [Azure DevOps pipeline with E2E integration](https://dev.azure.com/dnceng/public/_build/results?buildId=622097&view=results) tests for XHarness.
2. In one step, I produce a debug version of the NuGet package with my tool that I want to test in the next step. The tool has **always the same version** `1.0.0-ci` since it is not being published anywhere.
3. In the next job on our OSX machine (we have 4 currently), I download this NuGet as a build artifact file and install the tool locally by doing:
```
dotnet new tool-manifest
dotnet tool install --version 1.0.0-ci --add-source [dir] [name of the app]
```
5. The tool installs correctly, but after using it, I find out it is some old cached version installed by older build previously. It cached 1.0.0 before and uses that one.

- I tried --no-cache but no change for this command
- I am definitely sure I use the latest NuGet package because of other files inside (the helix job payload for example that does the install)
- I opened this https://github.com/dotnet/sdk/issues/11484 to ask

There are 2 ways to resolve this (I think):
1. Produce a new version numbers on each build and install that specific version
2. Remove the CLI from the cache using `nuget locals all -List`

premun:
This changes how we install the dotnet tools that are being tested during our integration testing.

We won't need to purge the NuGet cache anymore so this also fixes #123.

### DIFF
--- a/tests/integration-tests/helix-payloads/cli-integration-tests.sh
+++ b/tests/integration-tests/helix-payloads/cli-integration-tests.sh
@@ -2,46 +2,24 @@
 
 set -e
 
-if [ $# -ne 2 ] || [ -z "$1" ] || [ -z "$2" ]; then
+if [ "$#" -ne 2 ] || [ -z "$1" ] || [ -z "$2" ]; then
     echo "The script expects 2 arguments: where to store result logs and where to upload test results (ie. \$HELIX_WORKITEM_UPLOAD_ROOT \$HELIX_WORKITEM_ROOT)" 1>2
     exit 1
 fi
-
-version='1.0.0-ci'
-
-# Clean the NuGet cache from the previous 1.0.0-ci version of the tool
-# TODO: This might have a better solution: https://github.com/dotnet/xharness/issues/123
-echo "Cleaning the NuGet cache from the previous version of the tool..."
-
-# Call dotnet to get rid of the welcome message since the nuget command doesn't respect the --no-logo
-dotnet nuget locals All -l
-cache_dirs=`dotnet nuget locals All -l | cut -d':' -f 2 | tr -d ' '`
-while IFS= read -r path; do
-    echo "Purging cache in $path..."
-    rm -vrf "$path/microsoft.dotnet.xharness.cli/$version"
-    rm -vrf "$path/Microsoft.DotNet.XHarness.CLI/$version"
-done <<< "$cache_dirs"
 
 set -x
 
 here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+export DOTNET_ROOT=$(dirname $(which dotnet))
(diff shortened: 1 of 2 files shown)


tests/integration-tests/helix-payloads/cli-integration-tests.sh
```diff
@@ -2,47 +2,25 @@
...
-    rm -vrf "$path/microsoft.dotnet.xharness.cli/$version"
-    rm -vrf "$path/Microsoft.DotNet.XHarness.CLI/$version"
-done <<< "$cache_dirs"
-
 set -x
 
 here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+export DOTNET_ROOT=$(dirname $(which dotnet))
+dotnet tool install --no-cache --tool-path "$here/xharness-cli" --version "1.0.0-ci" --add-source "$here" Microsoft.DotNet.XHarness.CLI
```
MattGal, line 15: `$here` = HELIX_WORKITEM_ROOT.

While this will work, for real usage I'd recommend installing it to $HELIX_CORRELATION_PAYLOAD/xharness-cli so that it's cached for multiple work items in the same job.

tests/integration-tests/helix-payloads/cli-integration-tests.sh
```diff
@@ -2,47 +2,25 @@
...
+
 # TODO - Package the app bundle here using `dotnet xharness ios package ..`
 # For now we download a pre-bundled app:
-curl -L --output $here/app.zip "https://netcorenativeassets.blob.core.windows.net/resource-packages/external/macos/test-ios-app/System.Numerics.Vectors.Tests.app.zip"
+curl -L --output "$here/app.zip" "https://netcorenativeassets.blob.core.windows.net/resource-packages/external/macos/test-ios-app/System.Numerics.Vectors.Tests.app.zip"
```
MattGal, line 19: you could make this URL a correlation or work item payload directly.

premun: Oh, this is great! That would save me a lot of trouble with unzipping and things but unfortunately looks like this `send-to-helix` yaml template doesn't surface this capability ​​so I will just leave it as is since it will be replaced with Helix SDK soon​​

tests/integration-tests/helix-payloads/cli-integration-tests.sh
premun: Good to know for future reference but here I think it's fine since we only have 1 work item per job always. I would have to pass the variable in as I am not sourcing and then check for existence of the folder so I will just leave it as is since it will get replaced with Helix SDK anyway

PR #236, opened 2020-05-28
[Runners] Do publish the Common pkg.

mandel-macaque:
Else we will have issues like the following one:

```
error NU1101: Unable to find package Microsoft.DotNet.XHarness.TestRunners.Common. No packages exist with this id in source(s): dotnet-eng, dotnet-tools, dotnet5, dotnet5-transport, nuget.org
```


premun:
This doesn't seem right.. We are not publishing iOS or Android projects and CLI works.

Isn't there a way to tell it to include this Common inside? Or maybe Common is listed as a dependency instead of project reference?

mandel-macaque:
It is a project ref:
```
<ItemGroup>
    <ProjectReference Include="..\Microsoft.DotNet.XHarness.TestRunners.Common\Microsoft.DotNet.XHarness.TestRunners.Common.csproj" />
  </ItemGroup>
```

akoeplinger:
The difference is most likely that the CLI project is an exe/tool so nuget bundles all project references.

This looks good 👍

PR #237, opened 2020-05-28
Make sure Common is packed together with the Xunit project

premun:


premun:
We did the workaround in #236

PR #240, opened 2020-05-29
[wasm] Add running wasm tests as a process starting a javascript engine with arguments

MaximLipnin:
There are several javascript engines (e.g. v8, javascriptcore etc) which wasm tests can run on currently  and this list might be extended. Each of js engine can have (potentially different) set of arguments. e.g.
`v8 --expose_wasm runtime.js -- --enable-gc --run WasmTestRunner.dll System.Buffers.Tests.dll`

So the idea is to run the wasm runner inside of the wasm app by starting a js engine as a process and pass in respective arguments, in particular test runner and test assembly name.

This change is simplified version of what we want to get in the end.

cc @akoeplinger @steveisok

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
@@ -45,7 +45,7 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
             { "engine=|e=", "Specifies the JavaScript engine to be used",
                 v => Engine = ParseArgument<JavaScriptEngine>("engine", v)
             },
-            { "engine-arg=", "Argument to pass to the javascript engine. It can be used more than once.",
+            { "engine-arg=", "Argument to pass to the JavaScript engine. Can be used more than once.",
                 v => EngineArgs.Add(v)
             },
             { "js-file=", "Main JavaScript file to be run on the JavaScript engine",
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
@@ -48,7 +48,7 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
             { "engine-arg=", "Argument to pass to the javascript engine. It can be used more than once.",
                 v => EngineArgs.Add(v)
             },
-            { "js-file=", "Main javascript file to be run on the javscript engine",
+            { "js-file=", "Main JavaScript file to be run on the JavaScript engine",
                 v => JSFile = v
             },
         };
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
@@ -9,7 +9,9 @@ using Mono.Options;
 namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
 {
     /// <summary>
-    /// Specifies the javascript engine that is used to run WASM.
+    /// Specifies a name of a JavaScript engine binary used to run WASM application.
+    /// The invocation of the engine binary doesn't depend on a case,
+    /// i.g. both "v8 ..." and "V8 ..." should work.
     /// </summary>
     internal enum JavaScriptEngine
     {
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
@@ -8,8 +8,55 @@ using Mono.Options;
 
 namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
 {
+    /// <summary>
+    /// Specifies the javascript engine that is used to run WASM.
+    /// </summary>
+    internal enum JavaScriptEngine
+    {
+        /// <summary>
+        /// V8
+        /// </summary>
+        V8,
+        /// <summary>
+        /// JavaScriptCore
+        /// </summary>
+        JSC,
+        /// <summary>
+        /// SpiderMonkey
+        /// </summary>
+        SM,
+    }
+
     internal class WasmTestCommandArguments : TestCommandArguments
     {
-        protected override OptionSet GetTestCommandOptions() => new OptionSet();
+        private JavaScriptEngine? _engine;
+
+        public JavaScriptEngine Engine
+        {
+            get => _engine ?? throw new ArgumentException("Engine not specified");
+            set => _engine = value;
+        }
+
+        public List<string> EngineArgs { get; set; } = new List<string>();
+
+        public string JSFile { get; set;} = "runtime.js";
+
+        protected override OptionSet GetTestCommandOptions() => new OptionSet{
+            { "engine=|e=", "Run on the listed javascript engine.",
+                v => Engine = ParseArgument<JavaScriptEngine>("engine", v)
+            },
+            { "engine-arg=", "Argument to pass to the javascript engine. It can be used more than once.",
+                v => EngineArgs.Add(v)
+            },
+            { "js-file=", "Main javascript file to be run on the javscript engine",
+                v => JSFile = v
+            },
+        };
+
+        public override void Validate()
+        {
+            base.Validate();
+            Engine = Engine;
+        }
     }
 }
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
@@ -42,7 +42,7 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
         public string JSFile { get; set;} = "runtime.js";
 
         protected override OptionSet GetTestCommandOptions() => new OptionSet{
-            { "engine=|e=", "Run on the listed javascript engine.",
+            { "engine=|e=", "Specifies the JavaScript engine to be used",
                 v => Engine = ParseArgument<JavaScriptEngine>("engine", v)
             },
             { "engine-arg=", "Argument to pass to the javascript engine. It can be used more than once.",
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
@@ -48,7 +48,7 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
             { "engine-arg=", "Argument to pass to the JavaScript engine. Can be used more than once.",
                 v => EngineArgs.Add(v)
             },
-            { "js-file=", "Main JavaScript file to be run on the JavaScript engine",
+            { "js-file=", "Main JavaScript file to be run on the JavaScript engine. Default is runtime.js",
                 v => JSFile = v
             },
         };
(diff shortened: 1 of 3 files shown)


MaximLipnin:
Premek has a PR https://github.com/dotnet/xharness/pull/241 with additional logic to parse test command arguments which could cover the WASM case

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
```diff
@@ -10,6 +10,28 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
...
+        {
+            get => _engineArgs ?? throw new ArgumentException("Engine args not specified");
+            set => _engineArgs = value;
+        }
+
+        protected override OptionSet GetTestCommandOptions() => new OptionSet{
+            { "engine=|e=", "Run on the listed javascript engine.",
+                v => Engine = v
+            },
+            { "arg=", "Argument to pass to the javscript engine",
```
akoeplinger, line 32:
```suggestion
            { "arg=", "Argument to pass to the javascript engine",
```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
```diff
@@ -10,6 +10,28 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
 {
     internal class WasmTestCommandArguments : TestCommandArguments
```
premun, line 11: Please note that as it is now the [`TestCommandArguments`](https://github.com/dotnet/xharness/blob/master/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/TestCommandArguments.cs) which the WASM command inherits from already contains some required arguments (such as `--output-directory` that is used by Helix to say where to put results). So these need to be passed as well.

If you don't want that, you need to change the parent class.

premun, line 11: You could create a new override of `Validate()` in this class and validate the args here. It should also call `base.Validate()` too

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
```diff
@@ -25,17 +24,13 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+            string engineArgs = _arguments.EngineArgs;
 
-            var args = new List<string> { "test-assembly-filename" };
-            var cmdline = new CmdLineParser(args.ToArray());
-
-            var runner = new WasmRunner(cmdline.Project);
-            var result = runner.Run();
+            Process runner = Process.Start(engine, engineArgs);
+            runner.WaitForExit(timeout);
+            int result = runner.ExitCode;
```
premun, line 33: I would log the exit code

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
```diff
@@ -25,17 +24,13 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+            var timeout = _arguments.Timeout.Milliseconds;
+            string engine = _arguments.Engine;
+            string engineArgs = _arguments.EngineArgs;
 
-            var args = new List<string> { "test-assembly-filename" };
-            var cmdline = new CmdLineParser(args.ToArray());
-
-            var runner = new WasmRunner(cmdline.Project);
-            var result = runner.Run();
+            Process runner = Process.Start(engine, engineArgs);
```
premun, line 31: We have a class called `ProcessManager` available in the Common project that does provide a more safe way to run processes and collects logs. Have a look if you could leverage that

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
```diff
@@ -25,17 +24,13 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
             logger.LogDebug($"Wasm Test command called");
 
-            // FIXME This command line parsing should be moved to CLI command arguments
-            //  Parameters: 
-            // - directory name
-            // - test assembly filename
-            // - command line arguments
+            var timeout = _arguments.Timeout.Milliseconds;
+            string engine = _arguments.Engine;
+            string engineArgs = _arguments.EngineArgs;
```
premun, line 29: nit: Not sure there is any value in declaring these

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
MaximLipnin: yeah, thanks, it makes sense to add something first, and then we could get rid of all the redundant things

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
mandel-macaque: Yes, lets try to use the ProcessManager everywhere, will make our life easier, let us know if you need examples.

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
```diff
@@ -10,6 +10,28 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
...
+            get => _engineArgs ?? throw new ArgumentException("Engine args not specified");
+            set => _engineArgs = value;
+        }
+
+        protected override OptionSet GetTestCommandOptions() => new OptionSet{
+            { "engine=|e=", "Run on the listed javascript engine.",
+                v => Engine = v
+            },
+            { "arg=", "Argument to pass to the javascript engine",
+                v => EngineArgs = v
```
mandel-macaque, line 33: I wonder if we could add this to be a list rather than a single list. In mono.options you can allow an argument to be passed more than once, for example:
```
xharness wasm test --arg="foo" --arg="bar" --arg="tar"
```
Then in the Mono.Options we can do something like:
```csharp
{ "arg=" , "Argument to pass to the javascript engine",
    v => EngineArgs.Add (v);
}
```
The process manager does have a method that takes a collection of arguments and will quote them for you, which will make your life much easier.

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
```diff
@@ -8,8 +8,55 @@
...
+        public string JSFile { get; set;} = "runtime.js";
+
+        protected override OptionSet GetTestCommandOptions() => new OptionSet{
+            { "engine=|e=", "Run on the listed javascript engine.",
+                v => Engine = ParseArgument<JavaScriptEngine>("engine", v)
+            },
+            { "engine-arg=", "Argument to pass to the javascript engine. It can be used more than once.",
+                v => EngineArgs.Add(v)
+            },
+            { "js-file=", "Main javascript file to be run on the javscript engine",
```
premun, line 51: small typo

```suggestion
            { "js-file=", "Main JavaScript file to be run on the JavaScript engine",
```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
```diff
@@ -8,8 +8,55 @@
...
+            get => _engine ?? throw new ArgumentException("Engine not specified");
+            set => _engine = value;
+        }
+
+        public List<string> EngineArgs { get; set; } = new List<string>();
+
+        public string JSFile { get; set;} = "runtime.js";
+
+        protected override OptionSet GetTestCommandOptions() => new OptionSet{
+            { "engine=|e=", "Run on the listed javascript engine.",
```
premun, line 45:
```suggestion
            { "engine=|e=", "Specifies the JavaScript engine to be used",
```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
```diff
@@ -8,8 +8,55 @@
...
+                v => Engine = ParseArgument<JavaScriptEngine>("engine", v)
+            },
+            { "engine-arg=", "Argument to pass to the javascript engine. It can be used more than once.",
```
premun, line 48:
```suggestion
            { "engine-arg=", "Argument to pass to the JavaScript engine. Can be used more than once.",
```

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
```diff
@@ -18,16 +22,25 @@ internal class WasmTestCommand : TestCommand
...
+        protected override async Task<ExitCode> InvokeInternal(ILogger logger)
         {
-            logger.LogDebug($"Wasm Test command called");
+            var processManager = new MacOSProcessManager();
+
+            // Expected syntax
+            // xharness wasm test --engine=v8 --engine-arg=--expose_wasm --engine-arg=--some-other-thing --js-file=runtime.js -- --enable-gc --run WasmTestRunner.dll System.Buffers.Tests.dll
+           
+            var result = await processManager.ExecuteCommandAsync(
+                _arguments.Engine.ToString(),
```
premun, line 35: Do you want to do `.ToLower()` here maybe? Or the binaries are called exactly like the enum values?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
```diff
@@ -18,16 +22,25 @@ internal class WasmTestCommand : TestCommand
         protected override TestCommandArguments TestArguments => _arguments;
         protected override string CommandUsage { get; } = "wasm test [OPTIONS]";
```
premun, line 23: Please update this to match the `--` extra args

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
MaximLipnin: The engine binaries invocation doesn't depend on a case, i.g. both `v8 ...` and `V8 ...` should work.

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
```diff
@@ -8,8 +8,55 @@
...
+        public string JSFile { get; set;} = "runtime.js";
+
+        protected override OptionSet GetTestCommandOptions() => new OptionSet{
+            { "engine=|e=", "Specifies the JavaScript engine to be used",
+                v => Engine = ParseArgument<JavaScriptEngine>("engine", v)
+            },
+            { "engine-arg=", "Argument to pass to the JavaScript engine. Can be used more than once.",
+                v => EngineArgs.Add(v)
+            },
+            { "js-file=", "Main JavaScript file to be run on the JavaScript engine",
```
premun, line 51:
```suggestion
            { "js-file=", "Main JavaScript file to be run on the JavaScript engine. Default is runtime.js",
```

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
premun: Oh, ok. My other concern was that someone might change the enums, thinking it's a safe operation, but it would break this silent contract. Maybe update the docs of the enum at least to not that this silent contract exists?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
MaximLipnin: will update with `wasm test [OPTIONS] -- [ENGINE OPTIONS]`

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
MaximLipnin: updated the doc

PR #241, opened 2020-05-29
Add the -- pass through arguments and unit tests

premun:
A requirement for the WASM command where we need to pass arbitrary arguments to the engine/runner inside and also need to parse some the regular way.

Example:
```
xharness wasm test --enable-gc --assembly=WasmTestRunner.dll --assembly=System.Buffers.Tests.dll -- v8 --expose_wasm runtime.js
```

This will take everything after `--` and store it into `PassThroughArguments` property:
- `v8`
- `--expose_wasm`
- `runtime.js`

And then the rest will be parsed via Options in the Arguments class supplied for this:
```csharp
protected override OptionSet GetTestCommandOptions() => new OptionSet
{
    { "assembly=", v => Assemblies.Add(v) },
    { "enable-gc", _ => EnableGc = true },
};
```

Also contains unit tests for `XHarnessCommand` and `XHarnessCommandArguments`.


akoeplinger:
Hm there are actually three layers :)

1) `--expose_wasm runtime.js` -> arguments to the JS engine
2) `--enable-gc --assembly=WasmTestRunner.dll` -> arguments to runtime.js, WasmTestRunner.dll is basically the main app/executable and there can only be one
3) `System.Buffers.Tests.dll` -> arguments to the Main() method in WasmTestRunner.dll, can basically be anything

What's the best way to represent that? should we even try to parse this or rather just pass everything after the `--` verbatim

I think the `--expose_wasm` doesn't need to be an argument, it can be hardcoded internally when calling `v8`. Other JS engines probably need a different option to enable WASM.

premun:
@akoeplinger seems I misunderstood. Yeah, we could pass everything together I guess since that is what is happening in Maxim's PR I guess..

**Question:** should the pass-through args be before or after `--`? Currently in this iteration they are before but maybe that is weird?

akoeplinger:
@premun yes I think pass-through args should be everything after `--` since we want to pass these as-is without parsing.

So the only options parsed by xharness would be:
- `v8` == engine
- `--expose_wasm`== engine-args, though this should actually be hardcoded for v8 and we only allow passing _additional_ args
- `runtime.js` == main JS file

premun:
@akoeplinger currently, the way we have it is that we only have named arguments in XHarness and all arguments are defined via Mono.Options. I can make changes to support `v8 --expose_wasm runtime.js`, that shouldn't be hard, I can do it as part of this PR. Or these params will be known in advance and we can assign them names? This will make it impossible to pass additional args to v8 which is exactly what you asked for above, right?

**Question:** Do you think the engines will differ enough so that we will make `wasm test` a `CommandSet` that will have commands inside called `v8` and so on? We would then have `v8WasmTestCommand` for that. Or they are similar enough that `v8` would be just a positional param?

akoeplinger:
Yes I'm fine with e.g. `--engine-arg=--expose_wasm --js-file=runtime.js` or something like that.

Hm I don't think the JS engines are that different but I don't know exactly the tradeoffs for making a CommandSet so whatever you think is best.

premun:
@akoeplinger then if they are similar, we can just have `--engine=v8 --engine-arg=--expose_wasm --engine-arg=--some-other-thing --js-file=runtime.js`.

For the `--engine` we support automatic parsing of Enums and we can handle everything in 1 Command class:
https://github.com/dotnet/xharness/blob/9f4bb1e6f71d9cb1c789195656c1d74b75df94bd/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/iOS/iOSTestCommandArguments.cs#L113
(this also makes sure user all options get printed out in case they mess up or so..)

So I can merge this and we can wire it up in the other PR

PR #242, opened 2020-05-29
Add proper WebAssembly xunit runner

akoeplinger:
It uses a more lightweight xunit runner configuration that doesn't use threads.


src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
```diff
@@ -0,0 +1,137 @@
...
+            var testCasesToRun = discoverySink.TestCases.Where(filters.Filter).ToList();
+            Console.WriteLine($"Discovery finished.");
+
+            var summarySink = new DelegatingExecutionSummarySink(testSink, () => false, (completed, summary) => { Console.WriteLine($"Tests run: {summary.Total}, Errors: 0, Failures: {summary.Failed}, Skipped: {summary.Skipped}{Environment.NewLine}Time: {TimeSpan.FromSeconds((double)summary.Time).TotalSeconds}s"); });
+            var resultsXmlAssembly = new XElement("assembly");
+            var resultsSink = new DelegatingXmlCreationSink(summarySink, resultsXmlAssembly);
+
+            testSink.Execution.TestPassedEvent += args => { Console.WriteLine($"[PASS] {args.Message.Test.DisplayName}"); };
+            testSink.Execution.TestSkippedEvent += args => { Console.WriteLine($"[SKIP] {args.Message.Test.DisplayName}"); };
+            testSink.Execution.TestFailedEvent += args => { Console.WriteLine($"[FAIL] {args.Message.Test.DisplayName}{Environment.NewLine}{ExceptionUtility.CombineMessages(args.Message)}{Environment.NewLine}{ExceptionUtility.CombineStackTraces(args.Message)}"); };
```
mandel-macaque, line 58: Lets move this to use the logger, which you'll pass from the app entry point. That way all runners have the same interface, and we can write the a log file or any other location if we want. For example, if we want to run this on an iOS device, we will need a tcp writer to the usb port, so better be ready now.

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            testSink.Execution.TestAssemblyStartingEvent += args => { Console.WriteLine($"Running tests for {args.Message.TestAssembly.Assembly}"); };
+            testSink.Execution.TestAssemblyFinishedEvent += args => { Console.WriteLine($"Finished {args.Message.TestAssembly.Assembly}{Environment.NewLine}"); };
```
mandel-macaque, line 61: Same as above, bad selection in the review 🤦

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
```diff
@@ -0,0 +1,137 @@
...
+            resultsSink.Finished.WaitOne();
+
+            if (printXml)
+            {
+                Console.WriteLine($"STARTRESULTXML");
+                var resultsXml = new XElement("assemblies");
+                resultsXml.Add(resultsXmlAssembly);
+                Console.WriteLine(resultsXml.ToString());
+                Console.WriteLine();
+                Console.WriteLine($"ENDRESULTXML");
```
mandel-macaque, line 73: In the other runners, we write the xml to a file if we want and to the logger, same as the previous reason, lets try to have all of the them write into a logger that can later abstract the tranport.

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
```diff
@@ -0,0 +1,137 @@
...
+            var value = parts[1];
+            List<string> excludedTraits;
+            if (targetDictionary.TryGetValue(name, out excludedTraits!))
+            {
+                excludedTraits.Add(value);
+            }
+            else
+            {
+                targetDictionary[name] = new List<string> { value };
+            }
```
mandel-macaque, line 98: We already have this logic in the Common namespace, why re-doing it? Should we move it to a re-usable file?

akoeplinger:
As discussed on Slack the WebAssembly platform is quite limited right now so using the existing stuff will need some refactoring. Filed https://github.com/dotnet/xharness/issues/243

PR #246, opened 2020-06-03
Fix the pass-through arguments

premun:
The delimiter argument `--` used for pass-through args is removed by Mono.Options when parsing CommandSets so in Program.cs, we temporarily replace it with a placeholder


premun:
Merged to unblock Alexander

Issue #205, opened 2020-05-15
[Runners] The SkipTestException tests are considered a failure in the xml result


mandel-macaque:
We added a fix to do not count the tests that raise a SkipTestException as a failure but as a Skip. The issue with that fix is that it just focused in the the handler, but not the actual test results. 

Once everything has been ran, and xunit is used, xharness parses the results, and marks the build as a failure. The correct way to fix this is to modify the runner to use a correct test sink that **does not** communicate those tests as failures.

The dotnet team already does that:  https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.XUnitExtensions/src/SkippedFactTestCase.cs#L31 and https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.XUnitExtensions/src/SkippedTestMessageBus.cs#L19

Probably means we will need to use a new sink here: https://github.com/dotnet/xharness/blob/master/src/Microsoft.DotNet.XHarness.Tests.Runners/xUnit/XUnitTestRunner.cs#L49

safern:
fyi: @steveisok

mandel-macaque:
Closing after a conversation with @safern and @akoeplinger about the other route that will be taken.

PR #249, opened 2020-06-04
[Runners] Add a new delegating sink that converts failures to skips.

mandel-macaque:
Create a new delegating sink that will convert those tests that failed
due to the SkipException to sskipped tests rather than failed ones.

fixes: https://github.com/dotnet/xharness/issues/205

### DIFF
--- a/src/Microsoft.DotNet.XHarness.TestRunners.Xunit/DelegatingSkippedTestSink.cs
+++ b/src/Microsoft.DotNet.XHarness.TestRunners.Xunit/DelegatingSkippedTestSink.cs
@@ -1,0 +1,76 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Microsoft.DotNet.XUnitExtensions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.XHarness.TestRunners.Xunit
+{
+    /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.XHarness.TestRunners.Xunit/DelegatingSkippedTestSink.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                // if we failed due to the SkipException, create a test skipped
+                var exceptionType = testFailed.ExceptionTypes.FirstOrDefault();
+                if (exceptionType == typeof(SkipTestException).FullName)
+                {
+                    _skipCount++;
+                    var testSkipped = new TestSkipped(testFailed.Test, testFailed.Messages.FirstOrDefault());
+                    return _innerSink.OnMessage(testSkipped);
+                }
+                else
```
safern, line 43: No need for the `else` clause since you're returning inside the if.

akoeplinger:
As discussed offline we won't be going with this approach as it leads to problems in other layers of xunit, instead we'll be using the standard ConditionalFact/Theory.

safern:
Sorry for leading towards this path, I thought it was the more natural, simpler way without having to change a bunch of tests in libraries to use ConditionalFact/Theory.

mandel-macaque:
No problem, closing and will close the issue too.

PR #252, opened 2020-06-08
Plumbing process usage through IAdbProcessManager, basic tests

MattGal:


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -54,7 +59,7 @@ public AdbRunner(ILogger log, string adbExePath = "")
 
         public void SetActiveDevice(string? deviceSerialNumber)
         {
-            _currentDevice = deviceSerialNumber;
+            _processManager.DeviceSerial = deviceSerialNumber ?? "";
```
premun, line 62:
```suggestion
            _processManager.DeviceSerial = deviceSerialNumber ?? string.Empty;

```

src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
```diff
@@ -0,0 +1,97 @@
...
+    public class AdbProcessManager : IAdbProcessManager
+    {
+        private readonly ILogger log;
+        public AdbProcessManager(ILogger logger) => log = logger;
+
+        /// <summary>
+        ///  Whenever there are multiple devices attached to a system, most ADB commands will fail
+        ///  unless the specific device id is provided with -s {device serial #}
+        /// </summary>
+        public string DeviceSerial { get; set; } = "";
```
premun, line 19:
```suggestion
        public string DeviceSerial { get; set; } = string.Empty;
```

src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public ProcessExecutionResults Run(string adbExePath, string arguments)
+        { 
+            return Run(adbExePath, arguments, TimeSpan.FromMinutes(5));
+        }
+
+        public ProcessExecutionResults Run(string adbExePath, string arguments, TimeSpan timeOut)
+        {
+            string deviceSerialArgs = string.IsNullOrEmpty(DeviceSerial) ? string.Empty : $"-s {DeviceSerial}";
```
premun, line 28: I don't see why not make `DeviceSerial` nullable as well since now we still keep juggling with the empty strings (like in the AdbRunner `_processManager.DeviceSerial = deviceSerialNumber ?? string.Empty;`). Instead, `null` would just mean not set and we could react to it here accordingly

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
```diff
@@ -5,6 +5,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.DotNet.XHarness.Android;
+using Microsoft.DotNet.XHarness.Android.Execution;
```
premun, line 8: nit: revert this?

tests/Microsoft.DotNet.XHarness.Android.Tests/AdbRunnerTests.cs
```diff
@@ -0,0 +1,265 @@
...
+namespace Microsoft.DotNet.XHarness.Android.Tests
+{
+    public class AdbRunnerTests : IDisposable
+    {
+        private static readonly string s_scratchAndOutputPath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+        private static readonly string s_adbPath = Path.Combine(s_scratchAndOutputPath, "adb");
+        private static string s_currentDeviceSerial = "";
+        private readonly Mock<ILogger> _mainLog;
+        private readonly Mock<IAdbProcessManager> _processManager;
+        private Dictionary<Tuple<string, string>, int> _fakeDeviceList;
```
premun, line 27: nit: This one can be made `readonly` and the new assigned here since class is instantiated for every test again

src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
MattGal: Will consider it for later.  In this line whether it's empty or null I have to make the same check so I don't see the benefit.

PR #256, opened 2020-06-10
Kill the Simulator.app at the end of integration tests

premun:


premun:
@mandel-macaque you were saying there is some kind of a folder I could clean as well?

premun:
This won't be needed since I am removing the tests altogether in #271

premun:
This won't be needed since I am removing the tests altogether in #271

PR #259, opened 2020-06-11
Fix unit tests relying on local IP addresses

premun:
This started to fail today 100% of the time:
https://dev.azure.com/dnceng/public/_build/results?buildId=682330&view=ms.vss-test-web.build-test-results-tab&runId=21158168&resultId=100384&paneView=debug


premun:
Merging to unblock myself

PR #267, opened 2020-06-15
Resolve all code style warnings in the repo

premun:
No functional changes, just Visual Studio quick fixes of all the issues we have (200+)

Also fixes invalid severity level in .editorconfig: https://github.com/dotnet/runtime/commit/e45d130e9e01c553371ef47f21514ad07f2e0575


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -97,7 +97,7 @@ public void DumpAdbLog(string outputFilePath, string filterSpec = "")
             // Workaround: Doesn't seem to have a flush() function and sometimes it doesn't have the full log on emulators.
             Thread.Sleep(3000);
 
-            var result = RunAdbCommand($"logcat -d {filterSpec}", TimeSpan.FromMinutes(2));
+            ProcessExecutionResults? result = RunAdbCommand($"logcat -d {filterSpec}", TimeSpan.FromMinutes(2));
```
akoeplinger, line 100: I don't like replacing var with the type here and other places, can we turn that off?

mandel-macaque: I do prefer 'var' but approving nevertheless

premun: Yup. I was just doing what `dotnet/runtime` was doing. But I guess we can diverge, no problem

premun:
@MattGal as I said in Teams, the messages that are left now actually do have value because they point out dead code, skipped tests and similar. Before they were lost in the sea of code style and not so important messages.

Apart from that, this helped to move the Xamarin C# style into .NET Core and now we follow rest of the repositories which also has value.

I will revert the commit with `var` and change the editorconfig to not care.

Issue #268, opened 2020-06-16
lldb backtrace fails with an exception when lldb is not installed


akoeplinger:
Saw this on a Helix run:
```
[40m[32minfo[39m[22m[49m: test[0]
      05:28:27.3839256 /usr/bin/lldb --source /tmp/tmprrsiyZ.tmp
      
[41m[1m[37mcrit[39m[22m[49m: test[0]
      System.ComponentModel.Win32Exception (2): No such file or directory
         at System.Diagnostics.Process.ForkAndExecProcess(String filename, String[] argv, String[] envp, String cwd, Boolean redirectStdin, Boolean redirectStdout, Boolean redirectStderr, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
         at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
         at System.Diagnostics.Process.Start()
         at Microsoft.DotNet.XHarness.Common.Execution.ProcessManager.RunAsyncInternal(Process process, ILog log, ILog stdout, ILog stderr, Action`2 kill, Func`3 getChildrenPS, Nullable`1 timeout, Dictionary`2 environmentVariables, Nullable`1 cancellationToken, Nullable`1 diagnostics) in /_/src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs:line 282
         at Microsoft.DotNet.XHarness.Common.Execution.ProcessManager.KillTreeAsyncInternal(Int32 pid, Action`2 kill, Func`3 getChildrenPS, ILog log, Nullable`1 diagnostics) in /_/src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs:line 156
         at Microsoft.DotNet.XHarness.Common.Execution.ProcessManager.RunAsyncInternal(Process process, ILog log, ILog stdout, ILog stderr, Action`2 kill, Func`3 getChildrenPS, Nullable`1 timeout, Dictionary`2 environmentVariables, Nullable`1 cancellationToken, Nullable`1 diagnostics) in /_/src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs:line 313
[...]

PR #270, opened 2020-06-17
Check for lldb existence before invoking it

premun:
Fixes https://github.com/dotnet/xharness/issues/268


src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs
```diff
@@ -135,44 +136,49 @@ static async Task KillTreeAsyncInternal(int pid, Action<int, int> kill, Func<ILo
                         diagnostics: false);
                 }
 
-                foreach (var diagnose_pid in pids)
+                const string lldbPath = "/usr/bin/lldb";
```
premun, line 139: @akoeplinger should we just do `lldb` and let PATH handle it?

akoeplinger: Yeah. I'd also not check for existence of lldb and just make a catch-all so any exception from launching lldb is caught and ignored.

premun: Fixed

src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs
```diff
@@ -157,13 +159,21 @@ static async Task KillTreeAsyncInternal(int pid, Action<int, int> kill, Func<ILo
...
                                 stdout: log,
-                                stderr:log,
+                                stderr: log,
                                 kill: kill,
                                 getChildrenPS: getChildrenPS,
                                 timeout: TimeSpan.FromSeconds(30),
                                 diagnostics: false);
                         }
                     }
+                    catch (Win32Exception e) when (e.Message == "No such file or directory")
```
akoeplinger, line 169: Checking the message isn't ideal since it could be localized. Better to check `NativeError == 2` which is the error code for file not found on Unix.

Issue #273, opened 2020-06-22
xharness uses macOS ProcessManager on Linux


akoeplinger:
It tries to load the wrong libc:

```
      System.DllNotFoundException: Unable to load shared library '/usr/lib/libc.dylib' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: /usr/lib/libc.dylib: cannot open shared object file: No such file or directory
         at Microsoft.DotNet.XHarness.iOS.Shared.Execution.MacOSProcessManager.kill(Int32 pid, Int32 sig)
         at Microsoft.DotNet.XHarness.iOS.Shared.Execution.MacOSProcessManager.Kill(Int32 pid, Int32 sig) in /_/src/Microsoft.DotNet.XHarness.iOS.Shared/Execution/MacOSProcessManager.cs:line 146
         at Microsoft.DotNet.XHarness.Common.Execution.ProcessManager.<RunAsync>b__5_0(Int32 pid, Int32 sig) in /_/src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs:line 93
         at Microsoft.DotNet.XHarness.Common.Execution.ProcessManager.KillTreeAsyncInternal(Int32 pid, Action`2 kill, Func`3 getChildrenPS, ILog log, Nullable`1 diagnostics) in /_/src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs:line 193
         at Microsoft.DotNet.XHarness.Common.Execution.ProcessManager.RunAsyncInternal(Process process, ILog log, ILog stdout, ILog stderr, Action`2 kill, Func`3 getChildrenPS, Nullable`1 timeout, Dictionary`2 environmentVariables, Nullable`1 cancellationToken, Nullable`1 diagnostics) in /_/src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs:line 332
         at Microsoft.DotNet.XHarness.Common.Execution.ProcessManager.ExecuteCommandAsync(String filename, IList`1 args, ILog log, ILog stdout, ILog stderr, TimeSpan timeout, Dictionary`2 environmentVariables, Nullable`1 cancellationToken) in /_/src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs:line 59
[...]

PR #274, opened 2020-06-23
Add Windows/Linux ProcessManagers and improve ProcessManager logging

premun:
This change introduces platform dependent `ProcessManager` implementations to avoid problems like in #273 (WASM can be run anywhere).

Some notes:
- `MacOSProcessManager` was moved and with it logging had to move as well.
- I couldn't utilize the `Process.Kill(entireProcessTree: true)` because Xamarin is dependent on `netstandard2.1` version of the library (they are on .NET Framework) and this overload doesn't exist, only the one without the param. So tree killing using the `kill` command as it was implemented for Xamarin has to stay as is.
- Logging was improved for cases when processes time out
- I tested on OSX, Windows and Linux that a process can be killed properly

Fixes: #273


src/Microsoft.DotNet.XHarness.Common/Execution/LinuxProcessManager.cs
```diff
@@ -0,0 +1,16 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+
+namespace Microsoft.DotNet.XHarness.Common.Execution
+{
+    public class LinuxProcessManager : UnixProcessManager
+    {
+        [DllImport("/lib/x86_64-linux-gnu/libc.so.6")]
```
MattGal, line 11: Remind me why this doesn't use the managed C# process APIs?

premun: Please read the PR description

src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs
```diff
@@ -18,31 +19,22 @@ namespace Microsoft.DotNet.XHarness.Common.Execution
 {
     public abstract class ProcessManager : IProcessManager
     {
-
-        #region Abstract methods
+#region Abstract methods
 
         protected abstract int Kill(int pid, int sig);
         protected abstract List<int> GetChildrenPS(ILog log, int pid);
```
MattGal, line 25: not your change but this API is very poorly named.  If we're making massive changes, why not name it something like "GetChildProcessIds" or something that actually makes it clear what it's doing?  I don't know what a ChildrenPS is.

premun: Yeh, makes sense

src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs
```diff
@@ -286,16 +262,17 @@ private static async Task KillTreeAsyncInternal(int pid, Action<int, int> kill,
...
                     if (a != b)
                     {
                         sb.Append($"{key}={StringUtils.Quote(b)} ");
                     }
                 }
             }
+
             sb.Append($"{StringUtils.Quote(process.StartInfo.FileName)} {process.StartInfo.Arguments}");
-            log.WriteLine(sb);
+            log.WriteLine(sb.ToString());
```
MattGal, line 275: this doesn't change the behavior and at least on my VS would have a "you don't need to do this" squiggle.   Same for many of your ToString()s

src/Microsoft.DotNet.XHarness.Common/Execution/LinuxProcessManager.cs
akoeplinger: please use `[DllImport("libc.so")]` instead since that is way more portable.

src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs
premun: I removed the `StringBuilder` overloads because they are only used here and all implementations of `ILog` had to implement this (in the same way)

PR #278, opened 2020-06-29
Make sure redirected stdout events are finished in ProcessManager

akoeplinger:
We hit an issue in the wasm testing in dotnet/runtime where xharness would exit with 0 failed tests but the testResult.xml was empty.
This happens because we get the xml by reading stdout from the JS engine, but if xharness exits quickly enough the stdout event isn't handled anymore.

The documentation for the Process class mentions that you need to call WaitForExit() to make sure these events are called: See remarks in https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=netcore-3.1#System_Diagnostics_Process_WaitForExit_System_Int32_

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs
+++ b/src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs
@@ -318,8 +318,17 @@ namespace Microsoft.DotNet.XHarness.Common.Execution
                     }
                 }
             }
+            else
+            {
+                await WaitForExitAsync(process);
+            }
 
-            await WaitForExitAsync(process);
+            if (process.HasExited)
+            {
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.Common/Execution/ProcessManager.cs
```diff
@@ -318,8 +318,17 @@ public abstract class ProcessManager : IProcessManager
                     }
                 }
             }
+            else
+            {
+                await WaitForExitAsync(process);
+            }
```
akoeplinger, line 324: this isn't related to the PR but seems like an unrelated issue: if we have a timeout and WaitForExitAsync returns false then we'd call it a second time

PR #280, opened 2020-06-29
WASM: Add support for class/method filters to runner

akoeplinger:


premun:
cc @mandel-macaque

mandel-macaque:
Merging since it is needed for the pending status.

PR #285, opened 2020-07-28
Remove Windows build leg from internal builds

premun:
Internal builds will now only run on OSX since running on both Windows and OSX was causing problems with Arcade trying to push to BAR twice as can be seen here:
https://dev.azure.com/dnceng/internal/_build/results?buildId=747943&view=logs&j=74e82b2b-41b5-505f-7026-605948fa01e2&t=45774f52-cdc2-5894-6dc7-2cfd734f27bf

I decided to keep OSX and not Windows leg since we test artifacts from OSX in our PR integration tests

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,61 +34,8 @@ pr:
     - master
 
 stages:
-- ${{ if eq(variables._RunAsInternal, False) }}:
-  - stage: Build_Windows_NT
-    displayName: Build Windows
-    jobs:
-    - template: /eng/common/templates/jobs/jobs.yml
-      parameters:
-        enableTelemetry: true
-        enablePublishBuildArtifacts: true
-        enableMicrobuild: true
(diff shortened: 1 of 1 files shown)


MattGal:
Sorry not enough coffee, this means the bits won't be signed no?

MattGal:
Took a closer look; looks like the assets are actually identical, can you please just do the same change but disabling OSX.

azure-pipelines.yml
```diff
@@ -349,7 +348,6 @@ stages:
       enableSourceLinkValidation: false
       validateDependsOn:
       - Build_Windows_NT
```
MattGal, line 350: I don't think you want to / need to remove this here; otherwise it means that validate OSX starts before OSX finishes on public CI?

premun: There is just one Validate and there will be no OSX in internal build so not sure it should depend on a stage that won't happen?

image

PR #291, opened 2020-08-03
[wasm][tests] Add new `wasm test-browser` command

radical:
.. to run tests with a browser.

- This uses Selenium to run, and interact with the browser
- It specifically uses Chrome right now
- We start a webserver to host the test files, and depends on an
`index.html` generated by `WasmAppBuilder`.


src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmBrowserTestRunner.cs
```diff
@@ -0,0 +1,190 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+using Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm;
+using Microsoft.DotNet.XHarness.Common.CLI;
+using Microsoft.Extensions.Logging;
+
+using OpenQA.Selenium;
+using OpenQA.Selenium.Chrome;
+
+using SLogLevel = OpenQA.Selenium.LogLevel;
```
akoeplinger, line 20: nit: make this SeleniumLogLevel so it's clearer

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmBrowserTestRunner.cs
```diff
@@ -0,0 +1,190 @@
+using System;
```
akoeplinger, line 1: this needs the copyright header:

```suggestion
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

using System;
```

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmBrowserTestRunner.cs
```diff
@@ -0,0 +1,190 @@
...
+            }
+
+            uriBuilder.Query = sb.ToString();
+            return uriBuilder.ToString();
+        }
+
+        private static async Task<string> StartWebServer(CancellationToken token)
+        {
+            var host = new WebHostBuilder()
+                .UseSetting("UseIISIntegration", false.ToString())
```
akoeplinger, line 173: curious why we need that, shouldn't it be the default?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmBrowserTestRunner.cs
```diff
@@ -0,0 +1,190 @@
...
+                .UseKestrel()
+                .UseContentRoot(Directory.GetCurrentDirectory())
```
akoeplinger, line 175: Instead of using `Directory.GetCurrentDirectory()` we should flow the `_arguments.OutputDirectory` somehow here and create a folder there. This makes it more predictable in case the working directory of the process is not what we expect.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
 {
     internal class WasmTestCommand : TestCommand
     {
-        private const string CommandHelp = "Executes tests on WASM using a selected JavaScript engine";
+        private const string CommandHelp = "Executes tests on WASM using a selected JavaScript engine or a browser";
```
akoeplinger, line 21: I wonder whether it would make sense to split this off from the `wasm test` command into e.g. `wasm test-browser` since the features are pretty different.

@premun thoughts?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestWebServerStartup.cs
```diff
@@ -0,0 +1,27 @@
...
+            provider.Mappings [".wasm"] = "application/wasm";
+
+            foreach (var extn in new string[] { ".dll", ".pdb", ".dat", ".blat" })
+            {
+                provider.Mappings [extn] = "application/octet-stream";
+            }
+
+            app.UseStaticFiles (new StaticFileOptions
+            {
+                FileProvider = new PhysicalFileProvider (Directory.GetCurrentDirectory ()),
```
akoeplinger, line 22: same comment about the `GetCurrentDirectory` as above

src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
```diff
@@ -33,6 +33,10 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="3.1.3" />
     <PackageReference Include="Mono.Options" Version="6.6.0.161" />
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.2.0" />
+    <PackageReference Include="Selenium.WebDriver" Version="3.141.0" />
+    <PackageReference Include="Selenium.WebDriver.ChromeDriver" Version="2.43.0" />
```
akoeplinger, line 39: this shouldn't be needed

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestWebServerStartup.cs
```diff
@@ -0,0 +1,27 @@
+using System.IO;
```
akoeplinger, line 1: this needs the copyright header:

```suggestion
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

using System.IO;
```

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestCommand.cs
premun: Yeah, if we have a completely different code handling this then it shouldn't be difficult and I would be up for it.
We then wouldn't need the Chrome engine type which makes sense because it's not an engine.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmBrowserTestRunner.cs
radical: Hm I took the code from what we had for ProxyDriver. This part seems to be from old code. AFAICS, it is the default, so it can be removed. 👍

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmBrowserTestRunner.cs
premun: Isn't this where the web server gets files to serve from? Then I guess you would want the current folder with the files that you included in the payload of the work item?

I would also be careful with `_arguments.OutputDirectory` - this directory gets picked up by Helix and files inside uploaded with the results.
`Directory.GetCurrentDirectory()` is probably the Helix working directory which gets deleted after job is over so makes sense for that flow but not sure what this will cause for local development.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmBrowserTestRunner.cs
```diff
@@ -0,0 +1,190 @@
...
+    internal class WasmBrowserTestRunner
+    {
+        private readonly WasmTestCommandArguments _arguments;
+        private readonly Action<string> _processLogMessage;
+        private readonly ILogger _logger;
+        private readonly IEnumerable<string> _passThroughArguments;
+
+        // Messages from selenium prepend the url, and location where the message originated
+        // Eg. `foo` becomes `http://localhost:8000/xyz.js 0:12 "foo"
+        static readonly Lazy<Regex> s_consoleLogRegex = new Lazy<Regex>(() => new Regex(@"^\s*[a-z]*://[^\s]+\s+\d+:\d+\s+""(.*)""\s*$"));
```
premun, line 38: Does it make sense to precompile the regex maybe?

premun: Also not sure about the Laziness? Why not just get it ready before we start executing tests and measure time?

radical: Yeah, compiled sounds good. I'll remove the Lazy.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmBrowserTestRunner.cs
radical: Here, the outputDirectory seems to be set to `artifacts/bin/System.Collections.Immutable.Tests/net5.0-release/browser-wasm/AppBundle/xharness-output`, so that just gets the test results xml file. But the files that we need to host are one directory up.

akoeplinger: yeah my mental model of where the served files come from was broken :)
I still think we should pass the path to the files into xharness rather than relying on `Directory.GetCurrentDirectory()` being the right thing.

Basically pass the directory in like we do for `_arguments.JSFile`.

radical: Can I use `--app, -a=VALUE        Path to already-packaged app`, (`AppPackagePath`)? It doesn't seem to have a default value though, and is not a required parameter. I could use curdir as the default for this.

radical: To be clear, that's an existing parameter

radical: Err though the property throws if it isn't set (https://github.com/dotnet/xharness/blob/master/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/TestCommandArguments.cs#L26). And making it required will probably break existing stuff.

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
```diff
@@ -25,6 +25,7 @@ internal enum JavaScriptEngine
         /// SpiderMonkey
         /// </summary>
         SpiderMonkey,
+        Chrome
```
lewing, line 28: This can be removed now

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestBrowserCommandArguments.cs
```diff
@@ -0,0 +1,57 @@
...
+        Chrome
+    }
+
+    internal class WasmTestBrowserCommandArguments : TestCommandArguments
+    {
+        private Browser? _browser;
+
+        public Browser Browser
+        {
+            get => _browser ?? throw new ArgumentException("Browser not specified");
```
lewing, line 29: Is there a reason to make this a required option when there is only one value?

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestBrowserCommandArguments.cs
```diff
@@ -0,0 +1,57 @@
...
+            },
+        };
+
+        public override void Validate()
+        {
+            base.Validate();
+            if (Path.IsPathRooted (HTMLFile))
+            {
+                throw new ArgumentException("--html-file argument must be a relative path");
+            }
```
lewing, line 54: if browser is required why not check it here and use an auto property instead?

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestBrowserCommandArguments.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Collections.Generic;
+using System.IO;
+using Mono.Options;
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
+{
+    /// <summary>
+    /// Specifies a name of a JavaScript engine used to run WASM application.
+    /// </summary>
+    internal enum Browser
```
premun, line 15: This enum is probably not needed

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestCommandArguments.cs
```diff
@@ -24,7 +24,7 @@ internal enum JavaScriptEngine
         /// <summary>
         /// SpiderMonkey
         /// </summary>
-        SpiderMonkey,
+        SpiderMonkey
```
premun, line 27: nit: could be reverted

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmBrowserTestRunner.cs
akoeplinger: Hm yeah we should probably move `--app` and `--targets` down into Android/iOSTestCommandArguments.cs since they don't really make sense on wasm, but that can be done in a future PR :)

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestBrowserCommandArguments.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Mono.Options;
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
+{
+    /// <summary>
+    /// Specifies a name of a JavaScript engine used to run WASM application.
```
akoeplinger, line 13: should be 
```suggestion
    /// Specifies a name of a Browser used to run the WASM application.
```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestBrowserCommandArguments.cs
akoeplinger: Given we'll likely soon extend it with `Firefox`, `Safari` etc I think it's fine to keep this.

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestBrowserCommandArguments.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    internal class WasmTestBrowserCommandArguments : TestCommandArguments
+    {
+        public Browser Browser { get; set; } = Browser.Chrome;
+
+        public List<string> BrowserArgs { get; set; } = new List<string>();
+        public string HTMLFile { get; set; } = "index.html";
+
+        protected override OptionSet GetTestCommandOptions() => new OptionSet{
+            { "browser=|e=", "Specifies the browser to be used",
```
akoeplinger, line 31:
```suggestion
            { "browser=|b=", "Specifies the browser to be used",
```

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
```diff
@@ -0,0 +1,203 @@
...
+using Microsoft.AspNetCore.Hosting.Server.Features;
+using Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm;
+using Microsoft.DotNet.XHarness.Common.CLI;
+using Microsoft.Extensions.Logging;
+
+using OpenQA.Selenium;
+
+using SeleniumLogLevel = OpenQA.Selenium.LogLevel;
+using LogLevel = Microsoft.Extensions.Logging.LogLevel;
+using OpenQA.Selenium.Support.UI;
```
akoeplinger, line 25: nit: move this below `using OpenQA.Selenium;` on L22 so the usings are sorted

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        private string GetPackagePath ()
+        {
+            try
+            {
+                return _arguments.AppPackagePath;
+            }
+            catch (ArgumentException)
+            {
+                return Directory.GetCurrentDirectory();
```
akoeplinger, line 198: I'm not sure we really need this fallback, requiring passing `--app` is fine imho. It will be passed by the RunTests.sh script in dotnet/runtime anyway.

As a compromise I could live with supporting `--app .` and translating that into `Directory.GetCurrentDirectory()`

premun: I agree, leaving it as required parameter will not hurt anything

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            var options = new ChromeOptions();
+            options.SetLoggingPreference(LogType.Browser, SeleniumLogLevel.All);
+            options.AddArguments(new List<string>(_arguments.BrowserArgs)
+            {
+                "--incognito",
+                "--headless"
+            });
+
+            return new ChromeDriver(options);
```
directhex, line 64: This forks off a `chromedriver` instance from $PATH, but if the matching Chrome isn't in a location `chromedriver` can figure out by itself (which will be the case for us on Helix, as the plan is to drop Chrome in the correlation directory not in a system location which `chromedriver` can infer from the registry) then we need a way to tell `chromedriver` where to look for Chrome. The canonical way to set the path to Chrome is via `options.setBinary("/path/to/other/chrome/binary");` and we ought to have a mechanism to pass this to xharness, either via a command-line arg or the environment

akoeplinger: @directhex can `chromedriver` locate Chrome if it's in the $PATH?

directhex: You know, I didn't think it would, but apparently yes - https://github.com/chromium/chromium/blob/master/chrome/test/chromedriver/chrome/chrome_finder.cc (and tested locally on MacOS)

Which means if chromedriver and chromium are in the same dir (which they are in my branch) and added to %PATH% then it's fine and works

radical:
@akoeplinger any idea about these errors:

```
##[error].packages\microsoft.dotnet.arcade.sdk\5.0.0-beta.20374.1\tools\Sign.proj(40,5): error SIGN001: Signing 3rd party library 'F:\workspace\_work\1\s\artifacts\tmp\Release\ContainerSigning\58\tools/netcoreapp3.1/any/Newtonsoft.Json.dll' with Microsoft certificate 'Microsoft400'. The library is considered 3rd party library due to its copyright: 'Copyright © James Newton-King 2008'.
F:\workspace\_work\1\s\.packages\microsoft.dotnet.arcade.sdk\5.0.0-beta.20374.1\tools\Sign.proj(40,5): error SIGN001: Signing 3rd party library 'F:\workspace\_work\1\s\artifacts\tmp\Release\ContainerSigning\58\tools/netcoreapp3.1/any/Newtonsoft.Json.dll' with Microsoft certificate 'Microsoft400'. The library is considered 3rd party library due to its copyright: 'Copyright © James Newton-King 2008'.
##[error].packages\microsoft.dotnet.arcade.sdk\5.0.0-beta.20374.1\tools\Sign.proj(40,5): error SIGN001: Signing 3rd party library 'F:\workspace\_work\1\s\artifacts\tmp\Release\ContainerSigning\58\tools/netcoreapp3.1/any/Newtonsoft.Json.dll' with Microsoft certificate 'Microsoft400'. The library is considered 3rd party library due to its copyright: 'Copyright © James Newton-King 2008'.
F:\workspace\_work\1\s\.packages\microsoft.dotnet.arcade.sdk\5.0.0-beta.20374.1\tools\Sign.proj(40,5): error SIGN001: Signing 3rd party library 'F:\workspace\_work\1\s\artifacts\tmp\Release\ContainerSigning\59\tools/netcoreapp3.1/any/WebDriver.dll' with Microsoft certificate 'Microsoft400'. The library is considered 3rd party library due to its copyright: 'Copyright © Software Freedom Conservancy 2018'.
##[error].packages\microsoft.dotnet.arcade.sdk\5.0.0-beta.20374.1\tools\Sign.proj(40,5): error SIGN001: Signing 3rd party library 'F:\workspace\_work\1\s\artifacts\tmp\Release\ContainerSigning\59\tools/netcoreapp3.1/any/WebDriver.dll' with Microsoft certificate 'Microsoft400'. The library is considered 3rd party library due to its copyright: 'Copyright © Software Freedom Conservancy 2018'.
F:\workspace\_work\1\s\.packages\microsoft.dotnet.arcade.sdk\5.0.0-beta.20374.1\tools\Sign.proj(40,5): error SIGN001: Signing 3rd party library 'F:\workspace\_work\1\s\artifacts\tmp\Release\ContainerSigning\59\tools/netcoreapp3.1/any/WebDriver.dll' with Microsoft certificate 'Microsoft400'. The library is considered 3rd party library due to its copyright: 'Copyright © Software Freedom Conservancy 2018'.
##[error].packages\microsoft.dotnet.arcade.sdk\5.0.0-beta.20374.1\tools\Sign.proj(40,5): error SIGN001: Signing 3rd party library 'F:\workspace\_work\1\s\artifacts\tmp\Release\ContainerSigning\59\tools/netcoreapp3.1/any/WebDriver.dll' with Microsoft certificate 'Microsoft400'. The library is considered 3rd party library due to its copyright: 'Copyright © Software Freedom Conservancy 2018'.
  Signing Round 0: 28 files to sign.
```

radical:
@premun Can this be merged now?

premun:
Hm.. the master build failed on signing validation:
https://dev.azure.com/dnceng/internal/_build/results?buildId=763565&view=logs&j=0a6c679f-72be-5867-422e-acb741a068a3&t=d8e44ddf-8ea2-5518-25e7-4300471a271a

Maybe we skipped more signing than was appropriate?
(I haven't looked closely yet)

radical:
This is what I had added:
```
  <ItemGroup>
     <!-- For Selenium.WebDriver -->
     <ItemsToSign Remove="@(ItemsToSign->WithMetadataValue('Filename', 'WebDriver'))" />
     <ItemsToSign Remove="@(ItemsToSign->WithMetadataValue('Filename', 'Newtonsoft.Json'))" />
   </ItemGroup>
```

premun:
@radical I quickly checked the binlogs and it seems that it removes everything somehow. I compared a build from before:
image

And now:
image

I will check again tomorrow to see what's going on.

PR #293, opened 2020-08-10
Fix signing that broke with an earlier commit

radical:
https://github.com/dotnet/xharness/pull/291 - introduced a change to
skip some 3rd party assemblies for the signing step, but this broke
because of a msbuild issue: https://github.com/dotnet/msbuild/issues/5445 .

- Essentially, it ended up removing *all* the items to sign.

### DIFF
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -11,23 +11,8 @@
     <FileSignInfo Include="mscorlib.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="System.Core.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="System.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="WebDriver.dll" CertificateName="3PartySHA2" />
+    <FileSignInfo Include="Newtonsoft.Json.dll" CertificateName="3PartySHA2" />
   </ItemGroup>
-
-  <!-- We need  this to be inside a target to workaround: https://github.com/microsoft/msbuild/issues/5445 -->
-  <Target Name="PrepareItemsToSign" BeforeTargets="Sign">
-    <ItemGroup>
-      <!-- For Selenium.WebDriver -->
-      <ItemsToSign Remove="@(ItemsToSign->WithMetadataValue('Filename', 'WebDriver'))" />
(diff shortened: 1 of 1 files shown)


eng/Signing.props
```diff
@@ -13,11 +13,14 @@
...
+  <!-- We need  this to be inside a target to workaround: https://github.com/microsoft/msbuild/issues/5445 -->
+  <Target Name="PrepareItemsToSign" BeforeTargets="Sign">
+    <ItemGroup>
+      <!-- For Selenium.WebDriver -->
+      <ItemsToSign Remove="@(ItemsToSign->WithMetadataValue('Filename', 'WebDriver'))" />
+      <ItemsToSign Remove="@(ItemsToSign->WithMetadataValue('Filename', 'Newtonsoft.Json'))" />
+    </ItemGroup>
+  </Target>
 
   <PropertyGroup>
```
premun, line 25: @radical can you please remove the property group around `AllowEmptySignList` below?

I think we don't want that behavior and it would have warned us before that we are removing everything.

@MattGal or you remember why we have it?

premun:
It looks like we are signing only the nupkgs:
image

And inside, there are the 3rd party libraries. So the removal won't work.

I believe the way to fix this is to list these libraries using `FileSignInfo` like it is already present in `Signing.props` above your changes. This way it will hopefully know how to handle the contents.

eng/Signing.props
MattGal: @premun sorry I don't ; other once something works whether it's right or not it tends to be left alone.  Are you saying signing never worked?

premun: @MattGal not that - something else happened - because of a bug in MSBuild, @radical removed all `ItemsToSign` instead of just the ones he didn't want to get signed and our PR build then went green and we didn't notice nothing got signed actually. And then after merging, our internal CI failed on signing validation. So I was thinking we get rid of this (I checked and for example arcade doesn't have this set) and catch these type of mistakes earlier

radical:
@premun @MattGal I pushed a change to remove `AllowEmptySignList`, and added the two assemblies explicitly to `@(FileSignInfo)`. Is this what you had in mind?

premun:
Yes, let's see if the build is green and we can merge

PR #299, opened 2020-08-20
[wasm] Allow serving of unknown files

directhex:
By default, we only serve files with known MIME types, but that is an issue when dotnet library tests consume over a hundred different file extensions in their tests


src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestWebServerStartup.cs
```diff
@@ -31,7 +31,8 @@ public void Configure(IApplicationBuilder app)
             app.UseStaticFiles(new StaticFileOptions
             {
                 FileProvider = new PhysicalFileProvider(s_hostingEnvironment.ContentRootPath),
-                ContentTypeProvider = provider
+                ContentTypeProvider = provider,
+                ServeUnknownFileTypes = true
```
radical, line 35: Maybe add the commit message here as a comment too, so, somebody in future doesn't end up removing that?

radical: heh true. I guess, I had ended up missing it the last time, but that was because of a comment! Ignore me 👍

akoeplinger (approved):

PR #300, opened 2020-08-21
[wasm-browser][tests] Fix issue with some tests exiting with no output

radical:
The problem here is that these tests run very quickly, writing to the
console.log very fast, and selenium seems to not return the logs during
that time. So, when the tests are done, which is indicated by an html
element being set, we stop listening for more log messages.
And exit with the test exit code, but no output messages from the test
run.

Eg: `System.Data.Common`


premun:
@radical would you happen to have example builds where I could see the problem?

premun (approved):

PR #309, opened 2020-09-24
[wasm][tests] Fix timeout issues with `test-browser`

radical:
We use selenium to get the console logs from the browser, via
`chromedriver`. But in case of some test libraries (eg.
`System.Memory`), we get too many messages written to `console.log`,
which seems to stall the whole thing (probably, chrome is rate limiting
it).

Instead, we switch to using a websocket, which the test page connects
to, and redirects all the messages to that. (thanks to @lewing for the suggestion)

- So, the test page can connect to `ws://<server>/console`
- We also continue to check selenium for any other messages, or errors
that might show up

This allows the test output to show up much faster, and longer running
tests work fine.

Issues:

- If we exceed the timeout, then we want to shutdown the tests, but
selenium doesn't seem to be able to close the browser correctly
	- it takes a long time
	- and, leaves child chrome processes around

	- So, for this *specific* case, we kill the process tree
	manually

Fixes issues: dotnet/runtime#41269, and dotnet/runtime#41160 .

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
@@ -4,6 +4,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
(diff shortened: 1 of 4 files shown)


akoeplinger (approved):
Awesome!

PR #313, opened 2020-09-26
[wasm][tests] Identify tests that throw PNSE

radical:
- These tests show up as

`[PNSE] System.Net.Http.Functional.Tests.SocketsHttpHandler_Connect_Test.ConnectMethod_Success`

And the test results show:
`Tests run: 1774, Errors: 0, Failures: 13, PNSE: 1226, Skipped: 71.  Time: 10.198805s`

Here `PNSE:` shows the count for tests that throw PNSE.
And `Failures` *do not* include PNSE failures.

In the xml report, such tests are shown with `result="PNSE"`:

```xml
<test
	name="System.Net.Http.Functional.Tests.SyncHttpHandler_Connect_Test.ConnectMethod_Success"
	type="System.Net.Http.Functional.Tests.SyncHttpHandler_Connect_Test"
	method="ConnectMethod_Success" time="0.000525"
	result="PNSE">
```

.. instead of `result="Fail"`.

To detect these, we look for exception-type==PNSE .
This still misses cases where the exception might be wrapped in others.
We could check the messages xml element to get those too, but that might give us false positives.

Eg problem case: `System.Net.Http.Functional.Tests.HttpClientTest.CancelAllPending_AllPendingOperationsCanceled`
	- this uses `Assert.All`, so you get bunch of exceptions wrapped in
	`Xunit.Sdk.AllException`.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
@@ -163,8 +163,6 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
                     if (result.EndOfMessage)
                     {
                         var line = Encoding.UTF8.GetString(mem.GetBuffer(), 0, (int)mem.Length);
-                        line += Environment.NewLine;
-
                         _processLogMessage(line);
 
                         // the test runner writes this as the last line,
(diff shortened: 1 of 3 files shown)


radical:
I'll refactor this a bit to move the wasm specific stuff.

akoeplinger:
I don't understand why we want this?

There will be issues with other parts of the system that don't expect a `result="PNSE"` in the xml result.

lewing (commented):
Yeah, we need to work through the tests on wasm and mark/skip them appropriately, not assume anything.

lewing:
For the record I'd love to be able to supply custom filter code to categorize failures at the point of failure.

PR #314, opened 2020-09-27
[wasm][tests] Remove extra newline from output, making it a bit tighter 

radical:
This changes the output from:

```
dbug: test-browser[0]
      [FAIL] System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpProtocolTests.GetAsync_InvalidStatusLine_ThrowsExceptionOnSocketsHttpHandler(responseString: "HTTP/1.1 200\tOK")

dbug: test-browser[0]
      System.PlatformNotSupportedException : System.Net.Sockets is not supported on this platform.

```

.. to ..

```
dbug: test-browser[0]
      [FAIL] System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpProtocolTests.GetAsync_InvalidStatusLine_ThrowsExceptionOnSocketsHttpHandler(responseString: "HTTP/1.1 200\tOK")
dbug: test-browser[0]
      System.PlatformNotSupportedException : System.Net.Sockets is not supported on this platform.
```

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
@@ -163,8 +163,6 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
                     if (result.EndOfMessage)
                     {
                         var line = Encoding.UTF8.GetString(mem.GetBuffer(), 0, (int)mem.Length);
-                        line += Environment.NewLine;
-
                         _processLogMessage(line);
 
                         // the test runner writes this as the last line,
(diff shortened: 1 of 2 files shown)


radical:
Do we want to have some way to get just a regular Xunit like output? IOW, with no `dbug:..` lines. Is there some way to do that already (I couldn't find it)? If not, then what would it look like from a user pov, eg. command line arg?

CoffeeFlux (approved):
Thanks!

akoeplinger (approved):

akoeplinger:
> Do we want to have some way to get just a regular Xunit like output? IOW, with no dbug:.. lines.

Yeah I've got many request for doing that, I have a stash somewhere which implements this which I need to dust off :)

CoffeeFlux:
Regarding
> We can't easily get rid of the separate dbug: lines since they're hardcoded in Microsoft.Extensions.Logging but with my change above we'd have a separate wasm-console.log file which contains the plain stdout.

Can we open an issue about that specifically? Even if it's non-trivial, it would be really nice to see them go away one day.

akoeplinger:
Yeah, this seems to be a long-standing request: https://stackoverflow.com/questions/44230373/is-there-a-way-to-format-the-output-format-in-net-core-logging

Looks like we could write our own custom log format, I'll take a look at that.

PR #315, opened 2020-09-28
Improve wasm test logging

akoeplinger:
Don't show PASS/SKIP messages by default in the console log, unless verbose logging is enabled in xharness.
We capture all output and store it in the wasm-console.log file in the xharness output directory instead.

---

I got many requests from various dotnet team members to make the default wasm output more similar to the desktop xunit output, which only displays failed tests by default.

The wasm output now looks like this:

```
  info: test[0] Discovering: System.Runtime.Handles.Tests.dll (method display = ClassAndMethod, method display options = None)
  info: test[0] Discovered:  System.Runtime.Handles.Tests.dll (found 14 of 15 test cases)
  info: test[0] Starting:    System.Runtime.Handles.Tests.dll
  fail: test[0] [FAIL] SafeWaitHandleTests.SafeWaitHandle_False
  info: test[0] Assert.False() Failure
  info: test[0] Expected: False
  info: test[0] Actual:   True
  info: test[0]    at SafeWaitHandleTests.SafeWaitHandle_False()
  info: test[0]    at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
  info: test[0] Finished:    System.Runtime.Handles.Tests.dll
  info: test[0]
  info: test[0] === TEST EXECUTION SUMMARY ===
  info: test[0] Total: 14, Errors: 0, Failed: 1, Skipped: 0, Time: 0.250726s
```

PR #324, opened 2020-10-02
Use latest version of ASP.NET Core via FrameworkReference instead of nuget

akoeplinger:
We were pulling in Microsoft.AspNetCore 2.2 via a nuget PackageReference, but after upgrading Microsoft.Extensions.Logging in https://github.com/dotnet/xharness/pull/315 we saw the following error when using the "wasm test-browser" command:

```
fail: Microsoft.AspNetCore.Server.Kestrel[0]
      Heartbeat.OnHeartbeat
      System.TypeLoadException: Could not load type 'Microsoft.Extensions.Primitives.InplaceStringBuilder' from assembly 'Microsoft.Extensions.Primitives, Version=5.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.
         at Microsoft.Net.Http.Headers.DateTimeFormatter.ToRfc1123String(DateTimeOffset dateTime, Boolean quoted)
         at Microsoft.Net.Http.Headers.HeaderUtilities.FormatDate(DateTimeOffset dateTime, Boolean quoted)
         at Microsoft.Net.Http.Headers.HeaderUtilities.FormatDate(DateTimeOffset dateTime)
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.DateHeaderValueManager.SetDateValues(DateTimeOffset value)
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.DateHeaderValueManager.OnHeartbeat(DateTimeOffset now)
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure.Heartbeat.OnHeartbeat()
```

The reason is that the newer version of Microsoft.Extensions.Primitives removed InplaceStringBuilder.

Instead of using an old ASP.NET we can use the one in the framework instead, which is what's recommended since .NET Core 3.0

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -7,9 +7,9 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>61cde6e8fb9d5c9790867b279deb41783a780cd8</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.DotNet.Helix.Sdk" Version="5.0.0-beta.20474.4">
+    <Dependency Name="Microsoft.DotNet.Helix.Sdk" Version="5.0.0-beta.20502.2">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>61cde6e8fb9d5c9790867b279deb41783a780cd8</Sha>
+      <Sha>baebe30c7969b4148a6a3b90cf5f4e3a96d4a11e</Sha>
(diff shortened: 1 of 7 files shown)


radical (approved):
Ah! I had a local fix but this looks better 👍

premun (approved):

PR #327, opened 2020-10-05
Target .NET Standard 2.0 instead of 2.1 for code shared with Xamarin

premun:
We need to consume the Shared library in `xamarin-macios` which targets **.NET Framework 4.7.2**. When creating the project in `dotnet/xharness` we made a mistake and targeted .NET Standard 2.1 instead while 4.7.2 needs 2.0.

Some of the APIs are not available so minor changes are needed.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS.Shared/IErrorKnowledgeBase.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS.Shared/IErrorKnowledgeBase.cs
@@ -20,7 +20,7 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared
         /// <param name="installLog">The installation log.</param>
         /// <param name="knownFailureMessage">A string message for the user to understand the reason for the failure.</param>
         /// <returns>True if the failure is due to a known reason, false otherwise.</returns>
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.XHarness.iOS.Shared/IErrorKnowledgeBase.cs
```diff
@@ -20,22 +20,22 @@ public interface IErrorKnowledgeBase
         /// <param name="installLog">The installation log.</param>
         /// <param name="knownFailureMessage">A string message for the user to understand the reason for the failure.</param>
         /// <returns>True if the failure is due to a known reason, false otherwise.</returns>
-        bool IsKnownInstallIssue(IFileBackedLog installLog, [NotNullWhen(true)] out (string HumanMessage, string? IssueLink)? knownFailureMessage);
+        bool IsKnownInstallIssue(IFileBackedLog installLog, [Nullable(true)] out (string HumanMessage, string? IssueLink)? knownFailureMessage);
```
premun, line 23: Oops, wrong change

mandel-macaque (approved):

PR #328, opened 2020-10-06
[wasm][tests] Log messages with correct log level

radical:
This is done by accepting json messages from the test runtime.js, which
are of the form `{ method: 'console.debug', payload: 'foo bar' }`.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
@@ -4,6 +4,8 @@
 
 using System;
 using System.IO;
+using System.Text.Json;
+using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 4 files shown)


premun (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
```diff
@@ -24,8 +28,20 @@ public WasmTestMessagesProcessor(string xmlResultsFilePath, string stdoutFilePat
...
         }
 
-        public void Invoke(string line)
+        public void Invoke(string message)
         {
+            WasmLogMessage? logMessage = null;
+            string line;
+            try
+            {
+                logMessage = JsonSerializer.Deserialize<WasmLogMessage>(message);
```
akoeplinger, line 37: @radical can we check whether the `message` starts with `{` before trying to json-deserialize it to avoid throwing a lot of exceptions?

akoeplinger (commented):
LGTM apart from one comment, sorry for missing this PR :)

PR #331, opened 2020-10-12
Add option to set expected exit code for Android instrumentation

akoeplinger:
This is necessary so we can run the runtime tests from dotnet/runtime since they use exit code 100 to signal "success".


MattGal (approved):
The exit codes of 15 years ago continue to be honored :)

PR #332, opened 2020-10-15
Bring over changes done in xamarin/xamarin-macios since April

premun:
This PR forward integrates all changes done to the Shared library in the Xamarin repo since April (when the Shared library was copied here)


premun:
@rolfbjarne FYI - bringing over all of the changes so that we can then integrate the NuGet back to Xamarin

mandel-macaque (approved):
I forgot I landed som many fixes for the xml code :/

Issue #338, opened 2020-10-27
xharness sometimes doesn't parse the return-code


akoeplinger:
Example test run: https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-43804-merge-9c3fe24fa96341daa0/System.Xml.RW.FactoryReader.Tests/console.c6b2c6e3.log

It says:
```
04:16:56] info: Starting instrumentation class 'net.dot.MonoRunner' on net.dot.System.Xml.RW.FactoryReader.Tests
[04:16:56] dbug: Raw command: 'shell am instrument  -w net.dot.System.Xml.RW.FactoryReader.Tests/net.dot.MonoRunner'
[04:16:56] dbug: Executing command: '/home/helixbot/work/B3070977/p/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20519.2/runtimes/any/native/adb/linux/adb -s emulator-5556 shell am instrument  -w net.dot.System.Xml.RW.FactoryReader.Tests/net.dot.MonoRunner'
[04:17:21] info: Running instrumentation class net.dot.MonoRunner took 25.140556 seconds
[04:17:21] dbug: Exit code: 0
                 Standard Output:
                 
                 
[04:17:21] fail: No value for 'return-code' provided in instrumentation result.  This may indicate a crashed test (see log)
```

However if you look at the adb log you see that the test run completed successfully and we returned the correct `return-code` key:
https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-43804-merge-9c3fe24fa96341daa0/System.Xml.RW.FactoryReader.Tests/adb-logcat-net.dot.System.Xml.RW.FactoryReader.Tests.log

```
10-27 04:17:20.099  4349  4370 I DOTNET  : Xml file was written to the provided writer.
10-27 04:17:20.110  4349  4370 I DOTNET  : Tests run: 2292 Passed: 2292 Inconclusive: 0 Failed: 0 Ignored: 0
10-27 04:17:20.114  4349  4370 I DOTNET  : ----- Done -----
10-27 04:17:20.134  4349  4366 D DOTNET  : Exit code: 0.
10-27 04:17:20.134  4349  4366 I DOTNET  : MonoRunner finished, return-code=0
[...]

PR #341, opened 2020-10-28
Correctly use Process.WaitForExit() in AdbProcessManager

akoeplinger:
We were missing a call to the parameter-less Process.WaitForExit() since the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=netcore-3.1#System_Diagnostics_Process_WaitForExit) state:

> This overload ensures that all processing has been completed, including the handling of asynchronous events for redirected standard output. You should use this overload after a call to the WaitForExit(Int32) overload when standard output has been redirected to asynchronous event handlers.

Should help with https://github.com/dotnet/xharness/issues/338

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
@@ -67,27 +67,33 @@ namespace Microsoft.DotNet.XHarness.Android.Execution
             p.BeginOutputReadLine();
             p.BeginErrorReadLine();
 
-            // Sleeping 1 second allows the process time to send messages to the above delegates
-            // if the process exits very quickly
-            Thread.Sleep(1000);
-
             bool timedOut = false;
+            int exitCode = -1;
 
             // (int.MaxValue ms is about 24 days).  Large values are effectively timeouts for the outer harness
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
```diff
@@ -67,27 +67,33 @@ public ProcessExecutionResults Run(string adbExePath, string arguments, TimeSpan
             p.BeginOutputReadLine();
             p.BeginErrorReadLine();
 
-            // Sleeping 1 second allows the process time to send messages to the above delegates
-            // if the process exits very quickly
-            Thread.Sleep(1000);
-
             bool timedOut = false;
+            int exitCode = -1;
```
MattGal, line 71: maybe make this a large weird # to make it clear it happened.

MattGal (approved):

src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
akoeplinger: Hm yeah good point, I think ADB even returns -1 for some "normal" case. Will fix.

PR #342, opened 2020-10-29
Remove the `ios package` command

premun:
The package command was never finished, it is not maintained, it is not tested and packages are built by other tools in `dotnet/runtime`. If we ever choose to do that, we will re-use those tools or integrate them in XHarness.

It will be better if this code stays in Xamarin (where it changed already in the meantime) and Xamarin guys have full control over it. I imagine it will be way easier for them to make changes there.


premun:
The code has been backed up in branch `ios-package-command`

akoeplinger (approved):

akoeplinger:
@mandel-macaque please take a look

premun:
Oops, @akoeplinger didn't mean that! Only wanted to ping @mandel-macaque

mandel-macaque:
lets kill this. We can think of an aditonal helper tool if needed.

mandel-macaque (approved):

PR #340, opened 2020-10-28
Attempt https://github.com/dotnet/runtime/issues/43955 workaround using "--no-sandbox"

lewing:

PR #343, opened 2020-10-29
[wasm][test-browser] Retry launching chrome if it fails on startup

radical:
When starting chrome via selenium+chromedriver, it seems to crash
inexplicably (https://github.com/dotnet/runtime/issues/43955).
As a workaround, let's retry that a few times.

Also, adds `--no-sandbox` from https://github.com/dotnet/xharness/pull/340 .

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
@@ -50,7 +50,7 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
                                 logProcessor,
                                 logger);
 
-            var (driverService, driver) = GetChromeDriver();
+            var (driverService, driver) = GetChromeDriver(logger);
             try
             {
                 return await runner.RunTestsWithWebDriver(driverService, driver);
@@ -62,7 +62,7 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
             }
         }
 
-        private (DriverService, IWebDriver) GetChromeDriver()
+        private (DriverService, IWebDriver) GetChromeDriver(ILogger logger)
         {
             var options = new ChromeOptions();
             options.SetLoggingPreference(LogType.Browser, SeleniumLogLevel.All);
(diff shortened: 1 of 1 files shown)


akoeplinger:
I don't think we should be adding `--no-sandbox` as the docs say:

```
A common cause for Chrome to crash during startup is running Chrome as root user (administrator) on Linux. While it is possible to work around this issue by passing '--no-sandbox' flag when creating your WebDriver session, such a configuration is unsupported and highly discouraged. Please configure your environment to run Chrome as a regular user instead.
```

We're definitely not running Chrome as root so I'm a bit hesitant to add this highly discouraged option. The retry should hopefully fix it too.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -87,7 +88,29 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
+
+            int max_retries = 3;
+            int retry_num = 0;
+            while(true)
+            {
+                try
+                {
+                    return (driverService, new ChromeDriver(driverService, options, _arguments.Timeout));
+                }
+                catch (WebDriverException wde) when (wde.Message.Contains("exited abnormally") && retry_num < max_retries - 1)
```
akoeplinger, line 100: should we make this check for `failed to start: exited abnormally` so we don't accidentally retry other cases of abnormal exits, e.g. if the browser crashes due to some WASM bug while running tests.

akoeplinger:
We could also try passing the options puppeteer passes by default: https://github.com/puppeteer/puppeteer/blob/main/src/node/Launcher.ts#L159-L181

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
radical: This is catching exceptions only while starting the browser, before any page gets opened.

radical:
I have removed `--no-sandbox`, and added some extra arguments for chrome based on @akoeplinger's comment. I removed the ones that seem to be passed already (by chromedriver/selenium?).

akoeplinger (approved):

PR #344, opened 2020-10-30
Run "adb bugreport" when instrumentation process crashes

akoeplinger:
We've seen a few cases where we get `INSTRUMENTATION_RESULT: shortMsg=Process crashed.` and the adb log shows that the process crashed with a segmentation fault.
Android writes a crashdump into `/data/tombstones` but you can't just pull from there. Instead there's an adb command to create a bugreport which produces a .zip file with all sorts of info about the OS, including the tombstones.

See https://developer.android.com/studio/debug/bug-report.html


premun (approved):
Matt is not in today I think, so approving.

cc @MattGal

PR #347, opened 2020-10-30
[wasm][test-browser] Improve logging, and try to be more resilient to errors

radical:
This tries to handle new failures seen on CI after the last update to add verbose logging.

```
[13:08:27] crit: System.AggregateException: One or more errors occurred. (Cannot access a closed file.)
                  ---> System.ObjectDisposedException: Cannot access a closed file.
                    at System.IO.FileStream.ValidateReadWriteArgs(Byte[] array, Int32 offset, Int32 count)
                    at System.IO.FileStream.Write(Byte[] array, Int32 offset, Int32 count)
                    at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
                    at System.IO.StreamWriter.WriteLine(String value)
                    at Microsoft.DotNet.XHarness.CLI.Commands.Wasm.WasmTestMessagesProcessor.Invoke(String message) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs:line 103
                    at Microsoft.DotNet.XHarness.CLI.Commands.Wasm.WasmBrowserTestRunner.RunSeleniumLogMessagePump(IWebDriver driver, CancellationToken token) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs:line 203
                    at Microsoft.DotNet.XHarness.CLI.Commands.Wasm.WasmBrowserTestRunner.<>c__DisplayClass6_0.<RunTestsWithWebDriver>b__0() in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs:line 72
                    at System.Threading.Tasks.Task.InnerInvoke()
                    at System.Threading.Tasks.Task.<>c.<.cctor>b__274_0(Object obj)
                    at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
                 --- End of stack trace from previous location where exception was thrown ---
                    at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
                    at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
                    --- End of inner exception stack trace ---
                    at Microsoft.DotNet.XHarness.CLI.Commands.Wasm.WasmBrowserTestRunner.RunTestsWithWebDriver(DriverService driverService, IWebDriver driver) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs:line 128
                    at Microsoft.DotNet.XHarness.CLI.Commands.Wasm.WasmTestBrowserCommand.InvokeInternal(ILogger logger) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs:line 56
                    at Microsoft.DotNet.XHarness.Common.CLI.Commands.XHarnessCommand.Invoke(IEnumerable`1 arguments) in /_/src/Microsoft.DotNet.XHarness.Common/CLI/Commands/XHarnessCommand.cs:line 120
XHarness exit code: 1001
```

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
@@ -170,7 +170,7 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
                         }
                         catch (Exception ex) when (_messagesProcessor.WasmExitReceivedTcs.Task.IsCompletedSuccessfully)
                         {
-                            _logger.LogWarning($"Test has return a result already, but message processor threw {ex} while logging message: {line}");
+                            _logger.LogWarning($"Test has returned a result already, but message processor threw {ex} while logging message: {line}");
                         }
                     }
 
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
@@ -164,7 +164,14 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
                         var line = Encoding.UTF8.GetString(mem.GetBuffer(), 0, (int)mem.Length);
                         line += Environment.NewLine;
 
-                        _messagesProcessor.Invoke(line);
+                        try
+                        {
+                            _messagesProcessor.Invoke(line);
+                        }
+                        catch (Exception ex) when (_messagesProcessor.WasmExitReceivedTcs.Task.IsCompletedSuccessfully)
+                        {
+                            _logger.LogWarning($"Test has return a result already, but message processor threw {ex} while logging message: {line}");
+                        }
                     }
 
                     mem.SetLength(0);
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
@@ -164,14 +164,7 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
                         var line = Encoding.UTF8.GetString(mem.GetBuffer(), 0, (int)mem.Length);
                         line += Environment.NewLine;
 
-                        try
-                        {
-                            _messagesProcessor.Invoke(line);
-                        }
-                        catch (Exception ex) when (_messagesProcessor.WasmExitReceivedTcs.Task.IsCompletedSuccessfully)
-                        {
-                            _logger.LogWarning($"Test has returned a result already, but message processor threw {ex} while logging message: {line}");
-                        }
+                        _messagesProcessor.Invoke(line);
                     }
 
                     mem.SetLength(0);
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
@@ -67,9 +67,6 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
             var options = new ChromeOptions();
             options.SetLoggingPreference(LogType.Browser, SeleniumLogLevel.All);
 
-            // Enable this for more debugging info
-            options.SetLoggingPreference(LogType.Driver, SeleniumLogLevel.All);
-
             options.AddArguments(new List<string>(_arguments.BrowserArgs)
             {
                 "--incognito",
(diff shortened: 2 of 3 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
```diff
@@ -164,7 +164,14 @@ private async Task RunConsoleMessagesPump(WebSocket socket, TaskCompletionSource
...
                         line += Environment.NewLine;
 
-                        _messagesProcessor.Invoke(line);
+                        try
+                        {
+                            _messagesProcessor.Invoke(line);
+                        }
+                        catch (Exception ex) when (_messagesProcessor.WasmExitReceivedTcs.Task.IsCompletedSuccessfully)
+                        {
+                            _logger.LogWarning($"Test has return a result already, but message processor threw {ex} while logging message: {line}");
```
premun, line 173: Beware that if you do `threw {ex}`, then `ex.ToString()` returns all details with stack trace. So the message will be broken up.

```suggestion
                            _logger.LogWarning($"Test has returned a result already, but message processor threw {ex} while logging message: {line}");
```

radical: Yeah, I want the full trace here.

steveisok (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
premun: Yeah, but do you want it in the middle of the sentence?

Maybe doing this is better?
```csharp
$"Test has returned a result already, but message processor threw {name of type only} while logging message: {line}{Environment.NewLine}{ex}"
```

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
```diff
@@ -164,7 +164,14 @@ private async Task RunConsoleMessagesPump(WebSocket socket, TaskCompletionSource
                         var line = Encoding.UTF8.GetString(mem.GetBuffer(), 0, (int)mem.Length);
                         line += Environment.NewLine;
 
-                        _messagesProcessor.Invoke(line);
+                        try
+                        {
+                            _messagesProcessor.Invoke(line);
+                        }
+                        catch (Exception ex) when (_messagesProcessor.WasmExitReceivedTcs.Task.IsCompletedSuccessfully)
```
akoeplinger, line 171: Don't we need the same fix for `RunSeleniumLogMessagePump()` ?

I feel like this is just papering over the real issue, which is that the WasmTestMessagesProcessor is disposed before we're finished writing.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
radical: > Yeah, but do you want it in the middle of the sentence?

Heh I completely missed what you meant ;) /fixes

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
radical: > Don't we need the same fix for RunSeleniumLogMessagePump() ?

hm yeah

> I feel like this is just papering over the real issue, which is that the WasmTestMessagesProcessor is disposed before we're finished writing.

It is! I'm trying to get us to passing test runs on CI, ASAP, while trying to collect info to debug what might be causing this. The traces from the current runs were not helpful.

radical: I think that we were getting that error because we had selenium logging messages, and we would get some after the test was done, as selenium was closing the session. Since, now we are logging to a file instead, we don't get those messages. But trying to be careful anyway, and handle the error if, and when it pops up!

I think this might help too, but that's just based on reading the code, and not traces etc!
```diff
diff --git a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
index ccfd00a..7803f6a 100644
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
@@ -99,9 +99,11 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
             {
                 if (line.Contains("ENDRESULTXML"))
                 {
-                    _xmlResultsFileWriter.Flush();
-                    _xmlResultsFileWriter.Dispose();
+                    var oldWriter = _xmlResultsFileWriter;
                     _xmlResultsFileWriter = null;
+
+                    oldWriter.Flush();
+                    oldWriter.Dispose();
                     return;
                 }
```

akoeplinger (approved):

PR #357, opened 2020-11-09
Modulo the exit codes by 127 so they correspond with posix

premun:
Shell was already doing this and logs were a bit confusing:
```
XHarness exit code: 1107
+ exit_code=83
```


MattGal (approved):
Seems fine, the important part is just being distinct.

PR #366, opened 2020-11-18
Dont capture entire iOS simulator logs, only during test

premun:
Full iOS simulator log can have 2+MB while we are interested only in the part when we were running the current app.
We also don't want logs from previous runs of different apps.

Also remove some dead code from device log capturer.


akoeplinger (approved):
LGTM but please make sure cases like corrupted apps or install failures still produce useful data.

PR #367, opened 2020-11-18
Do not log the full XML when listing simulators

premun:
We have this output in a separate file, it is not as important and only clutters the logs.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/SimulatorLoader.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/SimulatorLoader.cs
@@ -81,9 +81,7 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared.Hardware
                         $"mlaunch {(result.TimedOut ? "timed out" : "exited")} with {result.ExitCode})");
                 }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/SimulatorLoader.cs
```diff
@@ -83,7 +83,7 @@ public async Task LoadDevices(ILog log, bool includeLocked = false, bool forceRe
 
                 var xmlContent = File.ReadAllText(tmpfile);
 
-                log.WriteLine("Simulator listing returned:" + Environment.NewLine + xmlContent);
+                log.WriteLine($"Simulator listing finished ({Math.Ceiling(((double)Encoding.UTF8.GetByteCount(xmlContent)) / 1024)} kB)");
```
akoeplinger, line 86: we're not using `xmlContent` anymore except for getting this byte count so it seems expensive to do a `File.ReadAllText()` when we could just do a `new FileInfo(tmpfile).Length` instead.

akoeplinger (changes requested):

akoeplinger (approved):

PR #369, opened 2020-11-19
Add exit code detector

### LINKED ISSUE #364: Detect exit code of iOS apps
**Context**
Currently, XHarness and mlaunch are unable to detect exit code of the iOS app that was spawned.
For scenarios where we don't have a TestRunner inside of the application, we have no information about the app run at all.

**Goal**
We should investigate if this is easy to do. Rolf confirmed that this was hard to do reliably across all iOS versions but we might just start doing it from the current one and see.

The way it can be done currently is from system logs of the device which we capture already.
We should use the `IErrorKnowledgeBase` strategy to scan for following lines in the device logs:

```
Nov 16 14:03:39 ML-MacVM com.apple.CoreSimulator.SimDevice.2E1EE736-5672-4220-89B5-B7C77DB6AF18[55655] (UIKitApplication:net.dot.HelloiOS[9a0b][rb-legacy][57331]): Service exited with abnormal code: 200
```

The important parts are the namespace of the app bundle which we know (**net.dot.HelloiOS**) and the message line itself.

premun:
Probably just a temporary solution before we figure out how to read this using mlaunch

Resolves #364

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS/ExitCodeDetector.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS/ExitCodeDetector.cs
@@ -11,12 +11,12 @@ namespace Microsoft.DotNet.XHarness.iOS
 {
     public interface IExitCodeDetector
     {
-        int DetectExitCode(AppBundleInformation appBundleInfo, IFileBackedLog systemLog);
+        int DetectExitCode(AppBundleInformation appBundleInfo, IReadableLog systemLog);
     }
 
     public class ExitCodeDetector : IExitCodeDetector
     {
-        public int DetectExitCode(AppBundleInformation appBundleInfo, IFileBackedLog systemLog)
+        public int DetectExitCode(AppBundleInformation appBundleInfo, IReadableLog systemLog)
         {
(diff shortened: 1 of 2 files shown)


premun:
The run command is also ready to follow after this. I only need to add tests:
```
XHarness command issued: ios run -a /Users/prvysoky/Documents/HelloiOS.app -t ios-simulator-64 -o=o
info: Preparing run for ios-simulator-64
info: Getting app bundle information..
info: Starting application 'HelloiOS' on ios-simulator-64
info: Application has finished with exit code: 200
XHarness exit code: 200
```

MaximLipnin (approved):

MaximLipnin:
@premun Great, thank you!

greenEkatherine (approved):

MaximLipnin:
Can we merge it or it's waiting for someone else's approve?

premun:
I was out on Friday and will merge this shortly. Today, I will open the second one with the `ios run` command. That one will need more review.

MaximLipnin:
Thank you Premek!

PR #379, opened 2020-11-25
Use xharness tool directly

mattleibow:
```
dotnet xharness

Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET program, but dotnet-xharness does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
```

I tried even `dotnet-xharness`.

Using the plain `xharness` works fine.


premun (approved):
Thanks! The dotnet CLI changed between 3.1 and 5.0 several times and we never fixed this.

PR #386, opened 2020-11-30
[wasm][test-browser] Improve logging, and retrying

radical:
- Log arguments passed to chromedriver
- Log `driver.log` if the driver crashes

- Re-create ChromeDriverService, when we are retrying. This is because
the exception from a failed chromedriver can show up when creating a
`ChromeDriverService`, or `ChromeDriver`.

- And delete older `driver.log` before starting

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
@@ -106,6 +106,9 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
                 "Cannot start the driver service"
             };
 
+            foreach (var file in Directory.EnumerateFiles(_arguments.OutputDirectory, "chromedriver-*.log"))
+                File.Delete(file);
+
             int max_retries = 3;
             int retry_num = 0;
             while(true)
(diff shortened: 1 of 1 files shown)


premun (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -112,8 +110,17 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
             int retry_num = 0;
             while(true)
             {
+                ChromeDriverService? driverService = null;
                 try
                 {
+                    var driverLogPath = Path.Combine(_arguments.OutputDirectory, $"chromedriver-{retry_num}.log");
+                    File.Delete(driverLogPath);
```
lewing, line 117: does this need to be here now that the logs are numbered?

lewing (approved):
Looks good.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
radical: Just useful when doing multiple runs locally, so that you are not really looking at a stale file. For example, if `chromedriver` cannot be found then the older log file won't get overwritten. Though maybe, we should delete `chromedriver-*.log` before we start, so be completely clear of stale log files?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -108,12 +106,21 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
                 "Cannot start the driver service"
             };
 
+            foreach (var file in Directory.EnumerateFiles(_arguments.OutputDirectory, "chromedriver-*.log"))
+                File.Delete(file);
```
akoeplinger, line 110: this is probably not really necessary since on helix each work item gets a new output directory anyway and locally you don't care if you have multiple files.

akoeplinger (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
radical: Just useful to to not have stale files, when you are trying to do multiple locals runs, and check if there were any logs produced, and/or if they had anything useful.

radical:
I don't have permission to merge this.

Issue #389, opened 2020-12-02
XHarness needs to handle emulators stuck in offline state


MattGal:
[Example log](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-44932-merge-72df977dfcb145978f/Microsoft.Extensions.Configuration.UserSecrets.Tests/console.777c9703.log?sv=2019-07-07&se=2020-12-22T18%3A27%3A18Z&sr=c&sp=rl&sig=z34s9Ke7rvxc%2B%2Bo9u363RAypA4iLIlUnKTDf9sLgKMk%3D)

```
[18:35:45] dbug: Executing command: '/home/helixbot/work/B99B09EC/p/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20601.1/runtimes/any/native/adb/linux/adb  devices -l'
[18:35:45] dbug: Evaluating output line for device serial: emulator-5554          offline transport_id:1
[18:35:45] dbug: Executing command: '/home/helixbot/work/B99B09EC/p/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20601.1/runtimes/any/native/adb/linux/adb  -s emulator-5554 shell getprop ro.product.cpu.abi'
[18:35:45] warn: Device 'emulator-5554' is offline; retrying up to one minute.
[18:35:55] dbug: Executing command: '/home/helixbot/work/B99B09EC/p/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20601.1/runtimes/any/native/adb/linux/adb  -s emulator-5554 shell getprop ro.product.cpu.abi'
[18:35:55] warn: Device 'emulator-5554' is offline; retrying up to one minute.
[18:36:05] dbug: Executing command: '/home/helixbot/work/B99B09EC/p/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20601.1/runtimes/any/native/adb/linux/adb  -s emulator-5554 shell getprop ro.product.cpu.abi'
[18:36:06] warn: No devices found with architecture 'x86_64'.  Just returning first available device; installation will likely fail, but we'll try anyways.
```

This example, caused by the rollout, was interesting.  Rebooting, for unknown reasons put at least one of the emulators into an offline state.  Rebooting again unblocked it.  The following Helix Kusto queries show this clearly:

```
WorkItems 
| where Started < datetime(2020-12-02 20:05:26.9020000) and Started > datetime(2020-12-02 18:31:05.1860000)
[...]

Issue #385, opened 2020-11-29
What is the correct "test-results-path" value for Android?


mattleibow:
I have tried a few places, but it seems xharness can't access the files:

I am trying this now:
`Application.Context.GetExternalFilesDir(null)?.AbsolutePath`

```
info: Attempting to pull contents of /storage/emulated/0/Android/data/com.microsoft.azure.devicetests/files/Documents/TestResults.xml to /Users/runner/work/1/s/output/test-results/android
dbug: Executing command: '/Users/runner/.dotnet/tools/.store/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20575.2/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20575.2/runtimes/any/native/adb/macos/adb -s emulator-5554 pull /storage/emulated/0/Android/data/com.microsoft.azure.devicetests/files/Documents/TestResults.xml /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ae3z53fb.5bv'
fail: ERROR: Exit code: 1
      Standard Output:
      adb: error: failed to stat remote object '/storage/emulated/0/Android/data/com.microsoft.azure.devicetests/files/Documents/TestResults.xml': Permission denied
```

And another thing is that the failed pull does _not_ actually fail the command. I still get a 0 return code and so all my builds were passing. I only noticed when I was looking for the results.

MattGal:
@mattleibow in the earliest versions of XHarness Android support, I would set up permissions to arbitrary paths and use these.  Currently, there's not actually any attempt to do this; the permission AFAIK (@steveisok please comment) comes from the APK itself, then the instrumentation passes back the path(s) to be pulled via key-value pair like you've noted e.g.:

```
test-results-path=/storage/emulated/0/Android/data/net.dot.System.Xml.Xsl.XslCompiledTransformApi.Tests/files/Documents/testResults.xml
```
The reason this is passing (can certainly make this error out in the case where the pull fails... just an oversight) is that the return code key value pair (like `return-code=0` in instrumentation results) contains the "success" exit code value.

I'm doing another fix for XHarness Android stuff today and I can make this an error scenario.

steveisok:
@EgorBo @akoeplinger Do you recall how perms are set?

EgorBo:
> @EgorBo @akoeplinger Do you recall how perms are set?

Not sure I follow, 
[...]

Issue #384, opened 2020-11-27
Failing launches on older (API 23) Android emulators


mattleibow:
I have a few runs on API 30, 29, 26, and 23. I get a good run on all each time, except for API 23. I randomly get a failure on DevOps (using the default hosted agents), but it appears to work locally.

Logcat:

[adb-logcat-com.microsoft.azure.devicetests (3).log](https://github.com/dotnet/xharness/files/5609733/adb-logcat-com.microsoft.azure.devicetests.3.log)

Console log:

```
info: Starting instrumentation class 'com.microsoft.azure.devicetests.TestInstrumentation' on com.microsoft.azure.devicetests
dbug: Raw command: 'shell am instrument  -w com.microsoft.azure.devicetests/com.microsoft.azure.devicetests.TestInstrumentation'
dbug: Executing command: '/Users/runner/.dotnet/tools/.store/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20575.2/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20575.2/runtimes/any/native/adb/macos/adb -s emulator-5554 shell am instrument  -w com.microsoft.azure.devicetests/com.microsoft.azure.devicetests.TestInstrumentation'
info: Running instrumentation class com.microsoft.azure.devicetests.TestInstrumentation took 0.255454 seconds
dbug: Exit code: 0
      Standard Output:
      
      
fail: No value for 'return-code' provided in instrumentation result.  This may indicate a crashed test (see log)
dbug: Executing command: '/Users/runner/.dotnet/tools/.store/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20575.2/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20575.2/runtimes/any/native/adb/macos/adb -s emulator-5554 logcat -d '
info: Wrote current ADB log to /Users/runner/work/1/s/output/test-results/android/adb-logcat-com.microsoft.azure.devicetests.log
info: Attempting to remove apk 'com.microsoft.azure.devicetests': 
dbug: Executing command: '/Users/runner/.dotnet/tools/.store/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20575.2/microsoft.dotnet.xharness.cli/1.0.0-prerelease.20575.2/runtimes/any/native/adb/macos/adb -s emulator-5554 uninstall com.microsoft.azure.devicetests'
info: Successfully uninstalled com.microsoft.azure.devicetests.
[...]

PR #392, opened 2020-12-03
Fixes for Xharness issues 354(again), 385, and 389

MattGal:
- Make failure to pull files or enumerate devices fail instead of warning.
- Detect an alternative version of the broken pipe issue (same recourse)
- Fail, and introduce exit code for inability to enumerate devices ; we'll do an infra retry and reboot to recover from this in Arcade.

https://github.com/dotnet/xharness/issues/384
https://github.com/dotnet/xharness/issues/385
https://github.com/dotnet/xharness/issues/389

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -134,7 +134,7 @@ namespace Microsoft.DotNet.XHarness.Android
             }
         }
 
-        public void WaitForDevice()
+        public void WaitForDevice(int timeToWaitForBootCompletionSeconds = 300)
         {
             // This command waits for ANY kind of device to be available (emulator or real)
             // Needed because emulators start up asynchronously and take a while.
@@ -148,16 +148,27 @@ namespace Microsoft.DotNet.XHarness.Android
                 throw new Exception($"Error waiting for Android device/emulator.  Std out:{result.StandardOutput} Std. Err: {result.StandardError}.  Do you need to set the current device?");
             }
 
-            // Once wait-for-device returns, we'll give it up to 30s for 'adb shell getprop sys.boot_completed' to be '1' (as opposed to empty) to make package managers happy
+            // Some users will be installing the emulator and immediately calling xharness, they need to be able to expect the device is ready to load APKs.
+            // Once wait-for-device returns, we'll give it up to timeToWaitForBootCompletionSeconds seconds (default 5 min) for 'adb shell getprop sys.boot_completed'
+            // to be '1' (as opposed to empty) to make subsequent automation happy.
+            var began = DateTimeOffset.UtcNow;
+            var waitingUntil = began.AddSeconds(timeToWaitForBootCompletionSeconds);
             var bootCompleted = RunAdbCommand($"shell getprop {AdbShellPropertyForBootCompletion}");
 
-            int triesLeft = 3;
-            while (!bootCompleted.StandardOutput.Trim().StartsWith("1") && triesLeft > 1)
+            while (!bootCompleted.StandardOutput.Trim().StartsWith("1") && DateTimeOffset.UtcNow < waitingUntil)
             {
                 bootCompleted = RunAdbCommand($"shell getprop {AdbShellPropertyForBootCompletion}");
                 _log.LogDebug($"{AdbShellPropertyForBootCompletion} = '{bootCompleted.StandardOutput.Trim()}'");
                 Thread.Sleep((int)TimeSpan.FromSeconds(10).TotalMilliseconds);
-                triesLeft--;
+            }
+
+            if (bootCompleted.StandardOutput.Trim().StartsWith("1"))
+            {
+                _log.LogDebug($"Waited {DateTimeOffset.UtcNow.Subtract(began).TotalSeconds} seconds for device for {AdbShellPropertyForBootCompletion} to be 1.");
+            }
+            else
+            {
+                _log.LogWarning($"Did not detect boot completion variable on device; variable used ('{AdbShellPropertyForBootCompletion}') may be incorrect or device may be in a bad state");
             }
         }
 
@@ -182,7 +193,6 @@ namespace Microsoft.DotNet.XHarness.Android
 
         public int InstallApk(string apkPath)
         {
-            bool needToRetry = false;
             _log.LogInformation($"Attempting to install {apkPath}: ");
             if (string.IsNullOrEmpty(apkPath))
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -210,22 +240,30 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
-                    logger.LogInformation($"Using first-found compatible device of {allDevicesAndTheirArchitectures.Count} total- serial: '{firstAvailableCompatible.Key}' - Arch: {firstAvailableCompatible.Value}");
-                    return firstAvailableCompatible.Key;
-                }
-                else
-                {
-                    logger.LogWarning($"No devices found with architecture '{apkRequiredArchitecture}'.  Just returning first available device; installation will likely fail, but we'll try anyways.");
-                    return allDevicesAndTheirArchitectures.Keys.First();
+                    if (allDevicesAndTheirArchitectures.Any(kvp => kvp.Value != null && kvp.Value.Equals(apkRequiredArchitecture, StringComparison.OrdinalIgnoreCase)))
+                    {
+                        var firstAvailableCompatible = allDevicesAndTheirArchitectures.Where(kvp => kvp.Value != null && kvp.Value.Equals(apkRequiredArchitecture, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
```
premun, line 250: I believe you can supply the condition right to `FirstOrDefault` and you could also switch the `Equals` if you're sure `apkRequiredArchitecture !== null`:

```suggestion
                        var firstAvailableCompatible = allDevicesAndTheirArchitectures.FirstOrDefault(kvp => apkRequiredArchitecture.Equals(kvp.Value, StringComparison.OrdinalIgnoreCase));
```

premun, line 250: I think it would be helpful for readability to state the type here instead of `var`

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -210,22 +240,30 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
                 {
-                    var firstAvailableCompatible = allDevicesAndTheirArchitectures.Where(kvp => kvp.Value != null && kvp.Value.Equals(apkRequiredArchitecture, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
-                    logger.LogInformation($"Using first-found compatible device of {allDevicesAndTheirArchitectures.Count} total- serial: '{firstAvailableCompatible.Key}' - Arch: {firstAvailableCompatible.Value}");
-                    return firstAvailableCompatible.Key;
-                }
-                else
-                {
-                    logger.LogWarning($"No devices found with architecture '{apkRequiredArchitecture}'.  Just returning first available device; installation will likely fail, but we'll try anyways.");
-                    return allDevicesAndTheirArchitectures.Keys.First();
+                    if (allDevicesAndTheirArchitectures.Any(kvp => kvp.Value != null && kvp.Value.Equals(apkRequiredArchitecture, StringComparison.OrdinalIgnoreCase)))
```
premun, line 248: Here, you could switch the `Equals` too and get rid of the `!= null`

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -210,22 +240,30 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
 
         private string? GetDeviceToUse(ILogger logger, AdbRunner runner, string apkRequiredArchitecture)
         {
-            var allDevicesAndTheirArchitectures = runner.GetAttachedDevicesAndArchitectures();
-            if (allDevicesAndTheirArchitectures.Count > 0)
+            try
             {
-                if (allDevicesAndTheirArchitectures.Any(kvp => kvp.Value != null && kvp.Value.Equals(apkRequiredArchitecture, StringComparison.OrdinalIgnoreCase)))
+                var allDevicesAndTheirArchitectures = runner.GetAttachedDevicesAndArchitectures();
+                if (allDevicesAndTheirArchitectures.Count > 0)
```
premun, line 246: Imho it would be better to do `if (allDevicesAndTheirArchitectures.Count == 0)` and return fast instead of nesting since if you forget to `return` from one of the branches below, you will get an error

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -210,22 +240,30 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+                        // In this case, the enumeration worked but nothing matched the APK; fail out.
+                        logger.LogError($"No devices found with architecture '{apkRequiredArchitecture}'.");
+                        return null;
+                    }
                 }
+                logger.LogError("No attached device detected");
+            }
+            catch (Exception toLog)
+            {
+                logger.LogError(toLog, "Exception thrown while trying to find compatible device with architecture [architecture]", apkRequiredArchitecture);
```
premun, line 265: Should this use `{}`? Or did the templating change?

```suggestion
                logger.LogError(toLog, "Exception thrown while trying to find compatible device with architecture {architecture}", apkRequiredArchitecture);
```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -193,26 +195,23 @@ public int InstallApk(string apkPath)
...
+            if (result.ExitCode == (int)AdbExitCodes.ADB_BROKEN_PIPE || result.StandardError.Contains(AdbInstallBrokenPipeError))
             {
                 _log.LogWarning($"Hit broken pipe error; Will make one attempt to restart ADB server, then retry the install");
                 KillAdbServer();
                 StartAdbServer();
-                needToRetry = true; 
+                result = RunAdbCommand($"install \"{apkPath}\"");
             }
 
+            // 2. Installation cache on device is messed up; restrting the device reliably seems to unblock this (unless the device is actually full, if so this will error the same)
```
premun, line 209:
```suggestion
            // 2. Installation cache on device is messed up; restarting the device reliably seems to unblock this (unless the device is actually full, if so this will error the same)

```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
premun, line 265: Are you sure you want to put the whole method into a try block?

Maybe `runner.GetAttachedDevicesAndArchitectures();` would be enough?

Where else can it fail?

premun (approved):
No big issues, just some suggestions left

PR #409, opened 2021-01-11
Add the expected exit code argument to the wasm test command

MaximLipnin:
This pull request adds `--expected-exit-code` argument to the `wasm test` command. It will allow to validate the result of the wasm functional test as we do for other platforms (ios, android).


MaximLipnin:
@premun @steveisok @akoeplinger

premun (approved):

Issue #67, opened 2020-04-23
Throw error on Windows/Linux.


mandel-macaque:
The ios verb cannot work on Windows, makes no sense. Throw an error when a user tries to execute any of the ios verbs on windows or linux.

premun:
There goes my development experience on Windows 😅

We also shouldn't show it in help on Windows as well.

But maybe we really just should have 2 packages - CLI.iOS and CLI.Android..
No Android users download a large package thanks to mlaunch and then won't be able to use most of it.

mandel-macaque:
@premun that is actually a good point, making the windows users to download mlaunch seems evil.

PR #412, opened 2021-01-12
Throw error when calling ios commands on Windows or Linux

greenEkatherine:
This should fix https://github.com/dotnet/xharness/issues/67

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/XHarnessHelpCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/XHarnessHelpCommand.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
 using Microsoft.DotNet.XHarness.CLI.Android;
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/XHarnessHelpCommand.cs
```diff
@@ -39,7 +40,15 @@ public override int Invoke(IEnumerable<string> arguments)
...
                     break;
                 case "ios":
-                    PrintCommandHelp(new iOSCommandSet(), subCommand);
+                    if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                    {
+                        PrintCommandHelp(new iOSCommandSet(), subCommand);
+                    }
+                    else
+                    {
+                        Console.WriteLine($"Command 'ios' could be run on OSX only.");
```
premun, line 49: nit:

```suggestion
                        Console.WriteLine($"Command '{command}' could be run on OSX only.");
```

because we will be changing it soon so that we don't forget it here..

premun (approved):

PR #425, opened 2021-01-22
[wasm] test-browser: Add support for Edge

radical:
The code for Edge, and Chrome is pretty much the same, except for the
concrete types, so sharing all that.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
@@ -115,28 +115,18 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
                         logger);
 
         private (DriverService, IWebDriver) GetEdgeDriver(ILogger logger)
-        {
-            string driverName = "msedgedriver";
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-                driverName = "msedgedriver.exe";
+            => GetChromiumDriver<EdgeOptions, EdgeDriver, EdgeDriverService>(
+                        "edgedriver",
+                        options =>
+                        {
+                            options.UseChromium = true;
+                            return EdgeDriverService.CreateDefaultServiceFromOptions(options);
+                        }, logger);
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -105,8 +109,37 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
+                        logger);
+
+        private (DriverService, IWebDriver) GetEdgeDriver(ILogger logger)
+        {
+            string driverName = "msedgedriver";
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                driverName = "msedgedriver.exe";
+
+            string? driverPath = Environment.GetEnvironmentVariable("PATH")
+                                    ?.Split(':')
```
akoeplinger, line 124: PATH is split by semicolon on Windows, should be changed to:

```suggestion
                                    ?.Split(Path.PathSeparator)
```

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -105,8 +109,37 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
+            => GetChromiumDriver<ChromeOptions, ChromeDriver, ChromeDriverService>(
+                        "chromedriver",
+                        options => ChromeDriverService.CreateDefaultService(),
+                        logger);
+
+        private (DriverService, IWebDriver) GetEdgeDriver(ILogger logger)
+        {
+            string driverName = "msedgedriver";
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                driverName = "msedgedriver.exe";
```
akoeplinger, line 121: I wonder if this is because it's looking for the legacy EdgeHTML driver, the `options.UseChromium` parameter seems to be false by default: https://alexanderontesting.com/2020/05/25/new-tricks-with-chromium-edge-and-msedgedriver-exe/

Can you try whether setting that option causes it to look for the correct driver name?

akoeplinger (changes requested):

akoeplinger (approved):

PR #426, opened 2021-01-25
Make sure iOS apps get uninstalled

premun:


src/Microsoft.DotNet.XHarness.CLI/Commands/iOS/iOSAppCommand.cs
```diff
@@ -162,38 +162,45 @@ private async Task<ExitCode> RunApp(ILogger logger, TestTargetOs target, Logs lo
...
                 }
                 catch (Exception e)
                 {
-                    installException = e;
+                    var message = new StringBuilder()
+                        .AppendLine("Application installation failed:")
+                        .AppendLine(e.ToString());
+
+                    logger.LogError(message.ToString());
+                    return ExitCode.PACKAGE_INSTALLATION_FAILURE;
```
greenEkatherine, line 179: Shouldn't app be uninstalled in case of exception?

premun: Fair point but if this actually blows up, we never get the `deviceName` where we tried to install the app. We could technically still look for it again but realistically if we don't find the device, we blow up with exception, if we do, mlaunch just quits with non-zero exit code and then we don't have an exception and we have `deviceName`.

greenEkatherine (approved):

akoeplinger (approved):

Issue #430, opened 2021-01-26
NullReferenceException when "getprop ro.product.cpu.abi" times out


akoeplinger:
Saw this here: https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-47344-merge-5a6c8098772545e0bb/Microsoft.Extensions.DependencyModel.Tests/console.070bf74d.log?sv=2019-07-07&se=2021-02-14T23%3A59%3A26Z&sr=c&sp=rl&sig=9t3%2FjnZR19lAaIBfy%2FQIdPKl3sxScllglrVTOxr5T8s%3D

```
[15:59:37] dbug: Executing command: 'C:\h\w\A76F0978\p\microsoft.dotnet.xharness.cli\1.0.0-prerelease.21068.3\runtimes\any\native\adb\windows\adb.exe  devices -l'
[15:59:37] dbug: Evaluating output line for device serial: 9CJAY1PTHW             device product:sargo model:Pixel_3a device:sargo transport_id:48
[15:59:37] dbug: Executing command: 'C:\h\w\A76F0978\p\microsoft.dotnet.xharness.cli\1.0.0-prerelease.21068.3\runtimes\any\native\adb\windows\adb.exe  -s 9CJAY1PTHW shell getprop ro.product.cpu.abi'
[16:04:37] fail: Waiting for command timed out: execution may be compromised.
[16:04:37] fail: Error trying to get device architecture: 
[16:04:37] crit: Failure to run test package: Object reference not set to an instance of an object.
                 System.NullReferenceException: Object reference not set to an instance of an object.
                    at Microsoft.DotNet.XHarness.CLI.Commands.Android.AndroidTestCommand.<>c__DisplayClass17_0.<GetDeviceToUse>b__0(KeyValuePair`2 kvp) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs:line 259
                    at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
                    at Microsoft.DotNet.XHarness.CLI.Commands.Android.AndroidTestCommand.GetDeviceToUse(ILogger logger, AdbRunner runner, String apkRequiredArchitecture) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs:line 259
[...]

PR #432, opened 2021-01-26
Correctly handle ADB timeouts when device calls fail, shorten timeout

MattGal:
Addresses https://github.com/dotnet/xharness/issues/430


MattGal:
@akoeplinger this, combined with https://github.com/dotnet/arcade/pull/6799, will cause the timeouts to get process architecture to happen faster and cause infrastructure retry and machine reboot when this happens.

akoeplinger (approved):

Issue #434, opened 2021-01-29
Increase test coverage for Android


greenEkatherine:
Check what unit and instrumentation tests should be added.

For unit tests coverage state at the moment is:
![image](https://user-images.githubusercontent.com/19972657/106282995-a5439c80-6241-11eb-94ce-7bebc1bc6aac.png)

PR #437, opened 2021-02-02
Add more unit tests

greenEkatherine:
Addressed #434 

Ideas about what else should be added as Android unit tests are welcome


premun (approved):
Surprised we don't really need all that code, but good job!

Is it possible we might need this code later?

greenEkatherine:
> Is it possible we might need this code later?

Not `GrantPermissions` I guess. It was used before but now it is set implicitly.

I was not sure about `PushFiles` but the comment indicates that it is also not ready to use:
`// This functionality may not end up getting used, so don't need to polish it until we're sure it is.`

PR #443, opened 2021-02-11
[iOS] Change the min sim for watchOS.

mandel-macaque:
The min simulator that can be used in xcode12.5 and later has changed.
We need to bump it.

This is a quick fix but we should be able to support the desired version
from the API.


premun (approved):

src/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/SimulatorLoader.cs
```diff
@@ -367,7 +367,7 @@ public async Task<(ISimulatorDevice Simulator, ISimulatorDevice? CompanionSimula
                 TestTarget.Simulator_iOS32 => "com.apple.CoreSimulator.SimDeviceType.iPhone-5",
                 TestTarget.Simulator_iOS64 => "com.apple.CoreSimulator.SimDeviceType." + (minVersion ? "iPhone-6" : "iPhone-X"),
                 TestTarget.Simulator_tvOS => "com.apple.CoreSimulator.SimDeviceType.Apple-TV-1080p",
-                TestTarget.Simulator_watchOS => "com.apple.CoreSimulator.SimDeviceType." + (minVersion ? "Apple-Watch-38mm" : "Apple-Watch-Series-3-38mm"),
+                TestTarget.Simulator_watchOS => "com.apple.CoreSimulator.SimDeviceType.Apple-Watch-Series-3-38mm"), // min version from xcode12.5 and later.
```
premun, line 370:
```suggestion
                TestTarget.Simulator_watchOS => "com.apple.CoreSimulator.SimDeviceType.Apple-Watch-Series-3-38mm", // min version from xcode12.5 and later.
```

PR #449, opened 2021-02-15
Remove dead code around TestRunner argument passing

premun:
- I compared what we send to and what we receive from the TestRunner EntryPoint and noticed quite a bit of unused options. These are remnants of the Xamarin code with which we don't share the EntryPoint anymore
- We are passing arguments both via env variables and arguments so we will only pass the env vars now
- There was a bug in the definition of the `class` argument and it wasn't parsed properly
- Another thing I fixed are the "extra" arguments - I don't they they were working until know because we prepend them with a `-app-arg:` prefix which is only recognized in the Xamarin's mono Main

### DIFF
--- a/src/Microsoft.DotNet.XHarness.TestRunners.Common/ApplicationOptions.cs
+++ b/src/Microsoft.DotNet.XHarness.TestRunners.Common/ApplicationOptions.cs
@@ -114,14 +114,16 @@ namespace Microsoft.DotNet.XHarness.TestRunners.Common
                     // if cannot parse, use default
                     if (bool.TryParse(v, out var runAll)) { RunAllTestsByDefault = runAll; } }},
                 {
-                    "method|m=", "Method to be ran in the test application. When this parameter is used only the " +
+                    "method|m=",
+                    "Method to be ran in the test application. When this parameter is used only the " +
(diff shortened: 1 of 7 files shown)


akoeplinger (approved):

greenEkatherine (approved):

src/Microsoft.DotNet.XHarness.TestRunners.Common/ApplicationOptions.cs
```diff
@@ -99,30 +68,26 @@ public ApplicationOptions()
 
             var os = new OptionSet() {
                 { "autoexit", "Exit application once the test run has completed.", v => TerminateAfterExecution = true },
-                { "autostart", "If the app should automatically start running the tests.", v => AutoStart = true },
```
mandel-macaque, line 102: This is wrong. Yes, it is not currently used, the issue is that if we want to support forms or do any type of IDE integration we want to be able to tell the application not to auto start running all the tests. If we remove this from the option set we will have to add it leter.

mandel-macaque (approved):
Do not remove autostart and i'm ok with it landing. That is a feature that will be helpful in the future.

PR #451, opened 2021-02-16
Set the NuGet label for the xcode/12.5 branch

premun:


mandel-macaque (approved):
Approving as a temp fix. The name should be chosen from from the branch name using an azdo var (already discussed)

Issue #383, opened 2020-11-27
Create a "default" or "common" instrumentation base


mattleibow:
When I am creating my Android apps, I have to implement an `Instrumentation` class. This is fine and works very well, however, I can't help but feel that this is somewhat "boilerplate". And I am sure I am doing it wrong.

For example, this is my edition: https://github.com/xamarin/Essentials/blob/cc08b9608999f40db02501cd27f3c7086e60eba7/DeviceTests/DeviceTests.Android/TestInstrumentation.cs

Or an older version: https://github.com/Azure/azure-mobile-apps-net-client/pull/559/files#diff-57ff0bad3452e9223031b3cdbbe553100132531c82f20d59534afe8246c0e97d

It works, but I am not doing the parsing of arguments for which test methods to run or properly set the `return-code`.

Maybe there should be some additional NuGet that adds the default logic for all of this and then have points to override various properties.

mattleibow:
This is actually a good thing for most because the bit for parsing the CLI arguments to do the filter and excludes is a non-trivial task. Doing this once in a library will be useful.

mattleibow:
I have another for reference that is more up-to-date: https://github.com/xamarin/Essentials/blob/main/DeviceTests/DeviceTests.Android/TestInstrumentation.cs

mattleibow:
Also, might be worth adding logic to automatically return the results file as I assume that is what most people actually want from this. Or at least have a way to easily get them without having to do it manually.

greenEkatherine:
I discussed my change with the team today and it seems that it cannot be merged in current state. The crucial point is that for default Instrumentation the dependency to `Instrumentation` and `Bundle` should be set which means dependency to Xamarin in either way. Taking this into account it doesn't look like XHarness is a good place for such base Instrumentation library since it's not generic and add extra dependency that we decided to avoid.

The solution that comes to my mind is splitting this change. Instrumentation base class' part is going to be created in Xamarin but the XHarness library will provide `TestEntryPoint` and dealing with test results path, test filtering etc. @mattleibow does it sound good for you?

@tkapin @premun please feel free to add your thoughts if I miss anything

mattleibow:
Sounds good. We can create a Xamarin + XHarness package and just do what you did in the PR.

I was playing around and I think we can do stuff. However, it will be nice to be able to put this code in this repo. It is just an `net6-android;net6-ios` project. and since this is a new package, we do not actually affect the dependency for the rest of the packages.

[...]

Issue #446, opened 2021-02-14
Add an option to return an error code if the tests failed


mattleibow:
I have to add custom code to detect any failed tests in the results so I can fail the build.

If there was a command line arg to basically use the test failure to also mean the command failed. This can wait until everything is cleaned up, but should still return a non 0 exit code if my tests failed.

mattleibow:
Maybe this is a "bug" since iOS actually fails on test failures...

premun:
As far as I know, XHarness exits with `1` if tests fail (at least on iOS):
https://github.com/dotnet/xharness/blob/66b9a52dfea930ff2cfb588e5f2636932a1f6868/src/Microsoft.DotNet.XHarness.Common/CLI/ExitCode.cs#L21

Or is this about Android only?

mattleibow:
This is Android only.

premun:
I see that the way it is currently handled in `dotnet/runtime` is that the wrapper itself returns 1 when it detects a failed test in the `TestRunner`:
https://github.com/dotnet/runtime/blob/372248b9b4e3e7c08e0c0aa53d4a94f826fe273b/src/libraries/Common/tests/AndroidTestRunner/AndroidTestRunner.cs#L34-L38

In XHarness we currently do this:
- **iOS** - we cannot reliable detect exit code and we understand the contents of `testResults.xml` enough to tell and then exit with 1
- **Android** - we can detect exit code of the Android run easily and just return that. However, we don't even open `testResults.xml` and we don't try to guess from that.

**Question 1**: Is it a lot of trouble for you to do what runtime does?
**Question 2**: Should we unify this across XHarness and have Android analyze test results? (prolongs run time, is more complicated)

mattleibow:
That runner appears to be a console app? Is it not possible for xharness to look at the exit cote returned by the bundle and determine a fail?

```

      
      
info: Found XML result file: '/storage/emulated/0/Documents/TestResults.xml'(key: test-results-path)
info: Attempting to pull contents of /storage/emulated/0/Documents/TestResults.xml to /Users/runner/work/1/s/output/testlogs/SkiaSharp.Android.Tests
[...]

Issue #431, opened 2021-01-26
Refactor arguments classes help info and common parts


greenEkatherine:
Recently several new classes for arguments were introduced for iOS and Android. Now some of them have duplicated parts that could be reused by adding base classes. https://github.com/dotnet/xharness/pull/429#discussion_r564446570

Also there are different parameters for iOS and Android commands that are showing in help info all together:

```XHarness command issued: help android test
usage: android test --output-directory=... --package-name=... --app=... [OPTIONS]
<...>
Arguments:


      --app, -a=VALUE        Path to already-packaged app
      --output-directory, -o=VALUE
                             Directory in which the resulting package will be
                               outputted
      --targets, -t=VALUE    Comma-delimited list of targets to test for
      --timeout=VALUE        TimeSpan or number of seconds to wait for
                               instrumentation to complete
      --method, -m=VALUE     Method to be ran in the test application. When
                               this parameter is used only the tests that have
                               been provided by the '--method' and '--class'
                               arguments will be ran. All other test will be
                               ignored. Can be used more than once.
      --class, -c=VALUE      Test class to be ran in the test application. When
                               this parameter is used only the tests that have
                               been provided by the '--method' and '--class'
                               arguments will be ran. All other test will be
                               ignored. Can be used more than once.
<...>

premun:
I have two things to add to this:
- I am more and more sure this can only be achieved using [C# 8.0 default interface implementations](https://devblogs.microsoft.com/dotnet/default-implementations-in-interfaces/)
- I think we should first have unit tests for the CLI so that we don't break passing any arguments. Then it will be safe to do this

PR #454, opened 2021-02-16
Add base entry point class for android tests

greenEkatherine:
Addressed #383 

With this approach test result file could be retrieved only on devices with API <=29

In scope of #385 I will add ability to pass test result file for API 30 as well


src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+            {
+                var message =
+                    $"Tests run: {results.ExecutedTests} " +
+                    $"Passed: {results.PassedTests} " +
+                    $"Inconclusive: {results.InconclusiveTests} " +
+                    $"Failed: {results.FailedTests} " +
+                    $"Ignored: {results.SkippedTests}";
+                bundle.PutString(SummaryMessageArgumentName, message);
+                
+                bundle.PutLong(ExitCodeArgumentName, results.FailedTests == 0 ? 0 : 1);
```
premun, line 86: I guess this sort of deals with #446? Does this mean the app will exit with `1`?
And if yes, can we reference the `ExitCode` enum and use the number from there (`TestsFailed` or how is it called)?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+            foreach (var key in arguments.KeySet())
+            {
+                string value = arguments.GetString(key);
+                if (!string.IsNullOrEmpty(value))
+                {
+                    bundleArguments.Add(key, value);
+                }
+            }
+
+            if (bundleArguments.ContainsKey(ResultsFileArgumentName) == false)
```
premun, line 61: This `== false` seems fishy

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+
+        public override void OnCreate(Bundle arguments)
+        {
+            base.OnCreate(arguments);
+
+            foreach (var key in arguments.KeySet())
+            {
+                string value = arguments.GetString(key);
```
premun, line 54: Not sure where this class comes from but is there also something like `GetObject` so that we don't lose type information or it has `GetString` only?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+                if (!string.IsNullOrEmpty(value))
+                {
+                    bundleArguments.Add(key, value);
+                }
+            }
+
+            if (bundleArguments.ContainsKey(ResultsFileArgumentName) == false)
+            {
+                bundleArguments.Add(ResultsFileArgumentName, "TestResults.xml");
```
premun, line 63: I think we use `testResults.xml` in other places. Can you please check for the `T` or `t`?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+            entryPoint.Tests = Tests;
+
+            await entryPoint.RunAsync();
+
+            if (File.Exists(entryPoint.TestsResultsFinalPath))
+            {
+                bundle.PutString(ResultsPathArgumentName, entryPoint.TestsResultsFinalPath);
+            }
+            
+            if (bundle.GetLong(ExitCodeArgumentName, -1) == -1)
```
premun, line 98: I think we had `ExitCode` for `ExitCodeNotSet`. Is this `1` expected by the Android API or it's up to us?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+                    yield return new TestAssemblyInfo(Assembly.LoadFrom(file), file);
+                }
+            }
+            protected override void TerminateWithSuccess()
+            {
+            }
+
+            protected override TestRunner GetTestRunner(LogWriter logWriter)
+            {
+                var testRunner = base.GetTestRunner(logWriter);
```
premun, line 161: Please compare with the `iOSApplicationEntryPoint`. When we were ignoring some methods, we would also flip some flag saying "do not execute all tests". Not sure we should do the same here

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+            protected override int? MaxParallelThreads => System.Environment.ProcessorCount;
+
+            protected override IDevice Device => null;
+
+            public IEnumerable<string> Tests { get; internal set; }
+
+            protected override IEnumerable<TestAssemblyInfo> GetTestAssemblies()
+            {
+                //yield return new TestAssemblyInfo(Assembly.GetExecutingAssembly(), Assembly.GetExecutingAssembly().Location);
+                //yield return new TestAssemblyInfo(typeof(Battery_Tests).Assembly, typeof(Battery_Tests).Assembly.Location);
```
premun, line 149: Forgotten code?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+            {
+                bundle.PutLong(ExitCodeArgumentName, 1);
+            }
+
+            Finish(Result.Ok, bundle);
+        }
+
+        class TestsEntryPoint : AndroidApplicationEntryPoint
+        {
+            readonly string resultsPath;
```
premun, line 108: nit: I think this project requires visibility keywords explicitly and private fields to begin with `_`. The settings in `.editorconfig` and VS should hint you

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+                var root = Application.Context.GetExternalFilesDir(null)?.AbsolutePath;
+
+                var docsDir = Path.Combine(root, "Documents");
+
+                if (!Directory.Exists(docsDir))
+                {
+                    Directory.CreateDirectory(docsDir);
+                }
+
+                resultsPath = Path.Combine(docsDir, arguments[ResultsFileArgumentName]);
```
premun, line 126: Does this mean all of the different bundles will keep overwriting the same testResults file?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <OutputType>Library</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.DotNet.XHarness.TestRunners.Xunit\Microsoft.DotNet.XHarness.TestRunners.Xunit.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="Mono.Android">
+      <HintPath>..\..\..\..\..\..\..\Program Files (x86)\Microsoft Visual Studio\2019\Preview\Common7\IDE\ReferenceAssemblies\Microsoft\Framework\MonoAndroid\v11.0\Mono.Android.dll</HintPath>
```
premun, line 16: As discussed on the sync, this looks fishy and won't work for non-Windows platforms. Can we get this from a NuGet maybe?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <IsPackable>true</IsPackable>
+    <TargetFrameworks>netstandard2.1;netcoreapp3.1</TargetFrameworks>
```
premun, line 5: If the project stays, it should probably be only `netstandard2.1`:
```suggestion
    <TargetFrameworks>netstandard2.1</TargetFrameworks>
```

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+        private const string ExitCodeArgumentName = "return-code";
+        private const string ExcludeCategoriesArgumentName = "exclude-categories";
+        private const string ExcludeMethodArgumentName = "exclude-method";
+        private const string IncludeMethodArgumentName = "include-method";
+        private const string ExcludeClassArgumentName = "exclude-class";
+        private const string IncludeClassArgumentName = "include-class";
+
+        protected Dictionary<string, string> bundleArguments;
+
+        protected AndroidInstrumentationBase()
```
premun, line 32: Do we need/want this ctor?

premun (commented):
Couple of questions left. Also please consider describing the usage in the README

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
mandel-macaque: While private is the default, we indeed use the explicit form due to the netcore style.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+            readonly string resultsPath;
+            readonly string includeMethod;
+            readonly string excludeCategories;
+            readonly string excludeMethod;
+            readonly string includeClass;
+            readonly string excludeClass;
+
+            public TestsEntryPoint(Dictionary<string, string> arguments)
+            {
+                var root = Application.Context.GetExternalFilesDir(null)?.AbsolutePath;
```
mandel-macaque, line 117: ? implies that root is nullable, right? Looks like it is not considered later in the file, for example, Path.Combine will throw a NRE parameter name path1.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+                includeMethod = arguments[IncludeMethodArgumentName];
+                excludeCategories = arguments[ExcludeCategoriesArgumentName];
+                excludeMethod = arguments[ExcludeMethodArgumentName];
+                includeClass = arguments[IncludeClassArgumentName];
+                excludeClass = arguments[ExcludeClassArgumentName];
+            }
+
+            protected override bool LogExcludedTests => true;
+
+            public override TextWriter Logger => null;
```
mandel-macaque, line 136: Hmm do we have #enable nullable? If that is the cqse TextWriter cannot be null, if it is, we will probably have issues. If indeed can be null, we should add TextWriter?

We should use nullable references when possible: https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            public override string TestsResultsFinalPath => resultsPath;
+
+            protected override int? MaxParallelThreads => System.Environment.ProcessorCount;
```
mandel-macaque, line 140: Careful! What happens if we add to many threads? The OS (in this case Android) will decide if an application is too greedy, if it considers that, it will terminate the application. That would result in a bad test run, we prefer slower test runs that will always finish. 

This has happened on iOS, will probably happen on Android too.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            protected override IDevice Device => null;
```
mandel-macaque, line 142: Same, nullable references, we should make sure that this will not result in a NRE, if it does not, lets add IDevice?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+                }
+                if (includeClass != null)
+                {
+                    foreach (var item in includeClass.Split(' '))
+                    {
+                        testRunner.SkipMethod(item, false);
+                    }
+                }
+                return testRunner;
+            }
```
mandel-macaque, line 195: Just as an exercise, you do not need to write that many times the same code, you can see the pattern:
```csharp
if (filter != null) {
   foreach (var f in fiter.Split (' ') {
      Lambda (f, included);
   }
}
```
where Lambda:
```csharp
Lambda = Action<string, bool> ()
```
For ref: https://docs.microsoft.com/en-us/dotnet/api/system.action-2?view=net-5.0

So, you can get all those lines in a single method, and no need to copy paste which is error prone.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            var entryPoint = new TestsEntryPoint(bundleArguments);
+            entryPoint.TestsCompleted += (sender, results) =>
+            {
+                var message =
+                    $"Tests run: {results.ExecutedTests} " +
+                    $"Passed: {results.PassedTests} " +
+                    $"Inconclusive: {results.InconclusiveTests} " +
+                    $"Failed: {results.FailedTests} " +
+                    $"Ignored: {results.SkippedTests}";
```
mandel-macaque, line 83: Exercise, how many times and how many strings are we using here? 

$"Text {var}" does something you are not expecting: https://sharplab.io/#gist:94115c1be673ee9bcc131ce02d9dc433

I count 5 calls to format and 1 call to contact while you can do 1 call to Format. Is a nit, but is good to use PRs to learn the platform.

mandel-macaque (changes requested):

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
premun: This is a good catch. @greenEkatherine can you please turn on the nullability for this project?

It is done by adding `<nullable>enable</nullable>` in the csproj (please check the other csproj, I'm not sure it's `enable` or `enabled`)

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit.csproj
greenEkatherine: It seems that there is no suitable NuGet for this. Since I use `Instrumentation` and `Bundle` from there I should rework the change significantly. Let me discuss this with @mattleibow

premun: Yeah, there must be a way since Matthew also must be implementing it somehow in the place that you are trying to now replace

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit.csproj
mattleibow: Not sure if we can use msbuild.sdk.extras?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit.csproj
mattleibow: Think think it is because I am building an Android library. Basically using msbuild.sdk.extras with a tfm of monoandroid10.0

mattleibow: You should be able to multi target across netstd and monoandroid.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
mattleibow: It has other methods https://docs.microsoft.com/en-us/dotnet/api/android.os.bundle?view=xamarin-android-sdk-9&viewFallbackFrom=xamarin-android-sdk-11

But, since this is a bundle from the OS and everything is required to be serializable, it might not change anything.

However, it might just be worth keeping a copy of the whole Bundle directly, no need to copy the values out. For the entrypoint, you could also use multitargeting to have an Android-specific variant that take a bundle and let the entrypoint worry about what it needs.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
mattleibow: You could also do `bundleArguments.TryGetValue(key, out var value)` and just let the system do all the checks for you.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
mattleibow: I know the runtime may be different, but the standard for xunit tests is `T`. I think. At least it was that with NUnit.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        protected AndroidInstrumentationBase()
+        {
+        }
+        protected AndroidInstrumentationBase(IntPtr handle, JniHandleOwnership transfer)
+            : base(handle, transfer)
+        {
+        }
+
+        protected virtual IEnumerable<string> Tests
```
mattleibow, line 40: Maybe this should be Assembly instead? Do all assemblies have a physical location? What about thinks like AOT and is there a case where the assemblies are read from a stream out of the apk? Maybe @jonathanpeppers  or @jonpryor  has a comment? Or even just use `TestAssemblyInfo` directly.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
mattleibow: Is it safe to do something like `max/2 + 1`?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit.csproj
mattleibow: might be worth just doing `netstandard2.1;monoandroid10.0;xamarin.ios10`. Then packing it will make sure to copy all the files to the right place and each OS gets what it needs.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
jonathanpeppers: In some cases, this path would not be useful. `.dll` files remain in the `.apk` on disk and you can't pull them with `adb pull` or anything.

In what way is this path used?

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: Unfortunately no, I commented in #446 how it works now. Also I replace numbers with `ExitCode` enum

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: I've checked in XHarness we use TestResults for NUnit and XUnit, so I leave it as it is

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: I've replaced it with `ExitCodeNotSet`, it's up to us I guess

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: I added handling with `runAllTests`

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: Now I get it why it was underlined in VS 😃

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: Not necessary because we just provide default name and for each test different names could be provided from the parameter. But I may try to generate somehow the name from test info

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit.csproj
greenEkatherine: Removed this dependency and targeted monoandroid

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: Changed to half of the threads

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: I checked, it could be null and don't be cause of exceptions

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: Agree, replaced with Assembly

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,192 @@
...
+            protected override IDevice? Device => null;
+
+            public IEnumerable<Assembly> Tests
+            {
+                get
+                {
+                    return new List<Assembly>();
+                }
+                internal set { }
+            }
```
mandel-macaque, line 155: This property looks very very strange. Lets thing about it for a second. 

get => Will always return a new List, if we for example do the following:

```csharp
var first = this.Tests;
var second = this.Tests;
Object.ReferenceEquals (first, first); // will return true
Object.ReferenceEquals (first, second); // will return false
```
Therefore we do have 2 instances that are different, either we do not want a property, we have an unusual behaviour or we got a bug.

Lets know look at the set. The way you wrote it is doing **nothing**. So 
```csharp
this.Test = new List<Assembly>(){"foo", "bar"}; // will have no effect, we alway return an empty list
var first = this.Test; // results in an empty list
```
You probably wanted to do the following:
```csharp
public IEnumerable<Assembly> Tests { get; set; } = new List<Assembly>();
```
The **get** and the **set** will be generated by the compiler, and it will be init to be an empty list. Take a look at https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,192 @@
...
+            : base(handle, transfer)
+        {
+        }
+
+        protected virtual IEnumerable<Assembly> Tests
+        {
+            get
+            {
+                yield return Assembly.GetExecutingAssembly();
+            }
```
mandel-macaque, line 46: I'm confuse by this property. Just returns the currently executing assembly. What is the goal? I'm not an android expert. But either:

1. We have on instrumentation with several test assemblies (each one of them with several tests).
2. We have an instrumentation class per assembly, with a number of tests.

If the second is correct, we could  modify the property to be just Assembly, else.. I don't understand why the class returns a IEnumerable.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            Finish(Result.Ok, bundle);
+        }
+
+        class TestsEntryPoint : AndroidApplicationEntryPoint
+        {
+            readonly string _resultsPath;
+            readonly string _excludeCategoriesDir;
+            readonly string _excludeCategoriesFile;
```
mandel-macaque, line 96: I think @premun  whats a 'private' in those:
```suggestion
            private readonly string _resultsPath;
            private readonly string _excludeCategoriesDir;
            private readonly string _excludeCategoriesFile;
```

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
```diff
@@ -0,0 +1,192 @@
...
+            {
+                var testRunner = base.GetTestRunner(logWriter);
+
+                ConfigureFilters(_parsedArguments[ExcludeMethodArgumentName], testRunner.SkipMethod, true);
+                ConfigureFilters(_parsedArguments[ExcludeClassArgumentName], testRunner.SkipClass, true);
+                ConfigureFilters(_parsedArguments[IncludeMethodArgumentName], testRunner.SkipMethod, false);
+                ConfigureFilters(_parsedArguments[IncludeClassArgumentName], testRunner.SkipClass, false);
+
+                return testRunner;
+            }
```
mandel-macaque, line 189: LOVE IT!

mandel-macaque (changes requested):

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: This property was aimed to pass tests from user to `TestsEntryPoint::GetTestAssemblies`. Anyway I'm going to remove `Instrumentation` class from this pull request to adjust it with #480

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/AndroidInstrumentationBase.cs
greenEkatherine: fixed

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit.csproj
greenEkatherine: It is not preserved after update

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/DefaultAndroidEntryPoint.cs
```diff
@@ -0,0 +1,98 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using Microsoft.DotNet.XHarness.TestRunners.Common;
+using Microsoft.DotNet.XHarness.TestRunners.Xunit;
+
+namespace Microsoft.DotNet.XHarness.DefaultAndroidEntryPoint.Xunit
```
premun, line 8: Can you please add a documentation header explaining the purpose of this class?
I think it will be very important for anyone looking into this.

Also please make a note in the README, where we talk about TestRunners. Otherwise people who are not Matthew won't know about this

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/DefaultAndroidEntryPoint.cs
```diff
@@ -0,0 +1,98 @@
...
+                    filterMethod(f, isExcluded);
+                }
+            }
+        }
+
+        protected override TestRunner GetTestRunner(LogWriter logWriter)
+        {
+            var testRunner = base.GetTestRunner(logWriter);
+
+            Tuple<string, Action<string, bool>, bool>[] filters =
```
premun, line 82: FYI, in C# 8.0+, you can use tuples without the Tuple class, you can even name the items.

The initialization would be:
```csharp
(string, Action<string, bool>, bool)[] filters
```
or with named version:
```csharp
(string Filter, Action<string, bool> FilterMethod, bool IsExcluded)[] filters
```
And the initialization would then be also shorter:
```csharp
(_parsedArguments.GetValueOrDefault(ExcludeMethodArgumentName), testRunner.SkipMethod, true),
```
Below, you would then do:
```csharp
ConfigureFilters(t.Filter, t.FilterMethod, t.IsExcluded);
```

premun: Or, please also consider creating a private class/record inside of this class:
```csharp
private class FilterDefinition { ...
```

And then use it here and in the ConfigureFilters method.

premun: Although, this ConfigureFilters is so short it could be inlined as well without problems too.

premun (approved):
Looks clean!

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/DefaultAndroidEntryPoint.cs
greenEkatherine: Cool, I was looking for something like this initialization, thanks!

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/DefaultAndroidEntryPoint.cs
greenEkatherine: added

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/DefaultAndroidEntryPoint.cs
greenEkatherine: done

mandel-macaque (approved):

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/DefaultAndroidEntryPoint.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            // use default name for test results file
+            _parsedArguments.TryAdd(ResultsFileArgumentName, "TestResults.xml");
+
+            if (!Directory.Exists(_parsedArguments[ResultsFileArgumentPath]))
+            {
+                Directory.CreateDirectory(_parsedArguments[ResultsFileArgumentPath]);
+            }
+
+            _resultsPath = Path.Combine(_parsedArguments[ResultsFileArgumentPath], _parsedArguments[ResultsFileArgumentName]);
```
mattleibow, line 58: There is no checks for this, so it will just crash if not provided. I'm thinking this should just be a string file path on the ctor.

mattleibow (commented):
Quite good. Just a comment on the visibility of the args.

But I am also not sure how the filter from the CLI gets to the entry point. I added a `--method` and it never exists on any of the command lines generated. If the CLI runner is going to pass them, then maybe we need to coordinate with what the CLI will use?

greenEkatherine:
> But I am also not sure how the filter from the CLI gets to the entry point. I added a `--method` and it never exists on any of the command lines generated. If the CLI runner is going to pass them, then maybe we need to coordinate with what the CLI will use?

`--class` and `--method` parameters are filtering for apple commands only. I'm sorry for the confusion, there are some mix of arguments in `help` command now but it's going to be fixed in #431 

However, `TestRunner` has more options to work with classes and methods filtering, so I added the possibility to use it in that way.

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/DefaultAndroidEntryPoint.cs
```diff
@@ -0,0 +1,116 @@
...
+    public class DefaultAndroidEntryPoint : AndroidApplicationEntryPoint
+    {
+        private const string ResultsFileArgumentName = "results-file-name";
+        private const string ResultsFileArgumentPath = "results-file-path";
+        private const string ExcludeCategoriesDirArgumentName = "exclude-categories-dir";
+        private const string ExcludeCategoriesFileArgumentName = "exclude-categories-file";
+        private const string ExcludeMethodArgumentName = "exclude-method";
+        private const string IncludeMethodArgumentName = "include-method";
+        private const string ExcludeClassArgumentName = "exclude-class";
+        private const string IncludeClassArgumentName = "include-class";
```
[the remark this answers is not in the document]
greenEkatherine: agree, done

src/Microsoft.DotNet.XHarness.InstrumentationBase.Xunit/DefaultAndroidEntryPoint.cs
greenEkatherine: replaced

XHarness.sln
```diff
@@ -42,6 +42,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.XHarness.A
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.XHarness.TestRunners.NUnit", "src\Microsoft.DotNet.XHarness.TestRunners.NUnit\Microsoft.DotNet.XHarness.TestRunners.NUnit.csproj", "{D51E7536-E698-42F4-939A-F259D543D1E1}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.XHarness.DefaultAndroidEntryPoint.Xunit", "src\Microsoft.DotNet.XHarness.InstrumentationBase.Xunit\Microsoft.DotNet.XHarness.DefaultAndroidEntryPoint.Xunit.csproj", "{FBDF39C0-1096-4088-B057-2FF8CCDFB32D}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Xamarin.Essentials.DeviceTests.Android", "..\..\..\Downloads\APK\Xamarin.Essentials.DeviceTests.Android.csproj", "{CB2072E0-A437-4811-AE17-16CAE0DDA1B1}"
+EndProject
```
mattleibow, line 48: I think this was not supposed to be here?

greenEkatherine: yep, removing

mattleibow:
Very nice! Thanks for all of this!

Issue #462, opened 2021-02-18
Detect exit code when calling `apple run` for MacCatalyst target 


premun:
1. It seems MacCatalyst doesn't forward the exit code of the app
2. We cannot get the system logs yet (#459) so we cannot get the exit code from there neither

We should investigate 1. or wait for 2.

akoeplinger:
@premun can you please try to run the binary in the `HelloCatalyst.app/Contents/MacOS` folder directly and see whether you get the exit code that way?

premun:
@akoeplinger will try!

Meanwhile, I was able to get it from the `/var/log/system.log` the same way I get it from Simulators btw

premun:
Ah, cool, running the binary works! Also gets us the logs!

Issue #459, opened 2021-02-17
Collect MacCatalyst system logs


premun:
When running on iOS simulators, we are able to get the device system logs.

For MacCatalyst apps, we might be able to do the same from the MacOS system logs. We should investigate this and possible capture.

PR #464, opened 2021-02-18
Add support for the MacCatalyst target

### LINKED ISSUE #435: Add Support for Mac Catalyst
We are in the process of building a mac catalyst runtime.  In order to run tests against it, I would like xharness to support running catalyst apps in order to take advantage of the underlying infrastructure.  

- [ ] Add new iOS target (`maccatalyst` would be on suggestion)
- [ ] Launch MacCatalyst app
  - [ ] Support the `test` command
  - [X] Support the `run` command (#462)
- [X] Collect MacCatalyst system logs (#459)

premun:
- Adds support for the `apple test -t=maccatalyst` command (resolves #435)
- Getting system logs for the process is not working at the moment (#459)
- The `apple run` command cannot detect exit code yet (#462)

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/AppRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppRunner.cs
@@ -245,7 +245,7 @@ namespace Microsoft.DotNet.XHarness.Apple
         }
 
         /// <summary>
-        /// Runs the MacCatalyst app via `open -W path.to.app`.
+        /// Runs the MacCatalyst app via running its binary in Contents/MacOS/[binary.app].
         /// </summary>
         private async Task<ProcessExecutionResult> RunMacCatalystApp(
             AppBundleInformation appInformation,
@@ -260,15 +260,12 @@ namespace Microsoft.DotNet.XHarness.Apple
 
             await crashReporter.StartCaptureAsync();
 
-            var arguments = new List<string>
-            {
-                "-W",
-                appInformation.LaunchAppPath,
-            };
-
+            var arguments = new List<string>();
             arguments.AddRange(_appArguments);
 
-            return await _processManager.ExecuteCommandAsync("open", arguments, _mainLog, timeout, null, cancellationToken);
+            var binaryPath = Path.Combine(appInformation.AppPath, "Contents", "MacOS", appInformation.AppName);
+
+            return await _processManager.ExecuteCommandAsync(binaryPath, arguments, _mainLog, timeout, null, cancellationToken);
         }
 
         private MlaunchArguments GetCommonArguments(int verbosity)
--- a/src/Microsoft.DotNet.XHarness.Apple/AppRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppRunner.cs
@@ -244,7 +244,7 @@ namespace Microsoft.DotNet.XHarness.Apple
             }
 
             // Arguments passed to the iOS app bundle
-            args.AddRange(_appArguments.Select(arg => new SetAppArgumentArgument(arg, true)));
+            args.AddRange(_appArguments.Select(arg => new SetAppArgumentArgument(arg)));
 
             return args;
         }
--- a/src/Microsoft.DotNet.XHarness.Apple/AppRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppRunner.cs
@@ -236,16 +236,7 @@ namespace Microsoft.DotNet.XHarness.Apple
 
         private MlaunchArguments GetCommonArguments(int verbosity)
         {
-            var args = new MlaunchArguments
-            {
-                new SetAppArgumentArgument("-connection-mode"),
-                new SetAppArgumentArgument("none"), // This will prevent the app from trying to connect to any IDEs
-
-                // On macOS we can't edit the TCC database easily
-                // (it requires adding the mac has to be using MDM: https://carlashley.com/2018/09/28/tcc-round-up/)
-                // So by default ignore any tests that would pop up permission dialogs in CI.
-                new SetEnvVariableArgument(EnviromentVariables.DisableSystemPermissionTests, 1),
-            };
+            var args = new MlaunchArguments();
 
             for (var i = -1; i < verbosity; i++)
             {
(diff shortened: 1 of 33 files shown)


src/Microsoft.DotNet.XHarness.Apple/AppRunner.cs
```diff
@@ -219,7 +229,7 @@ public class AppRunner : AppRunnerBase
             {
                 await crashReporter.StartCaptureAsync();
 
-                _mainLog.WriteLine("Starting test run");
+                _mainLog.WriteLine("Starting the app");
```
mandel-macaque, line 232: It is a test run, right? Why even change it?

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
```diff
@@ -93,6 +94,31 @@ public class AppTester : AppRunnerBase
...
+                        skippedMethods,
+                        skippedTestClasses,
+                        cancellationToken);
+
+                    return ("MacCatalyst", catalystTestResult, catalystResultMessage);
+                }
+                finally
+                {
+                    deviceListener.Cancel();
+                    deviceListener.Dispose();
```
mandel-macaque, line 118: the 'using' keyword does exactly this.

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
```diff
@@ -343,54 +369,156 @@ public class AppTester : AppRunnerBase
...
+                    timeout,
+                    envVariables.ToDictionary(p => p.Key, p => p.Value is bool ? p.Value.ToString().ToLowerInvariant() : p.Value.ToString()),
+                    cancellationToken);
+
+                await testReporter.CollectSimulatorResult(result);
+            }
+            finally
+            {
+                deviceListener.Cancel();
+                deviceListener.Dispose();
```
mandel-macaque, line 452: same using does do this already.

mandel-macaque (changes requested):
Fix the use of the pattern:
```csharp
try {
} finally {
    IDisposable.Dispose();
}
```

The above is generated by the compiler. compiler > human

Ref: https://sharplab.io/#gist:4aad44262924444d74167abd07a043a7

```il
finally
        {
            // sequence point: hidden
            IL_002c: ldloc.0
            IL_002d: brfalse.s IL_0035

            IL_002f: ldloc.0
            IL_0030: callvirt instance void [System.Private.CoreLib]System.IDisposable::Dispose()

            // sequence point: hidden
            IL_0035: endfinally
        } // end handler
```

src/Microsoft.DotNet.XHarness.Apple/AppRunner.cs
premun: For the `app run` command, it's not a test run

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
premun: I am not sure I can do that actually.

We kind of need to call the `Cancel` before `Dispose` because of how the listener works so if there would be an exception, it would just call `Dispose` before it would reach `Cancel`

premun: Ideally, the listener class would do this but I cannot change this because of `xamarin-macios`

mandel-macaque: Why? It an IDisposable has a Cancel it should be called in the dispose. For example in the IO classes you have Close and you don't need to call it if you are calling Dispose. I'll take a look at the listeners because if that is the case, they should be fixed.

premun: Well, I don't know.. you wrote these 😅

premun: I can add an internal flag `_isClosed` and lock on it and do the `Cancel`

premun: I was actually half way there during this implementation

greenEkatherine (approved):

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
premun: I created a follow-up as this seems to be a bit tricky (#467)

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
premun: Will be dealt with in #467

premun:
@akoeplinger I changed the way we call the open/binary and updated the PR with this new way so technically #462 can be closed when this is merged.

src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
```diff
@@ -45,5 +53,32 @@ protected async Task<string> FindDevice(TestTargetOs target)
...
+            IEnumerable<string> appArguments,
+            Dictionary<string, object> environmentVariables,
+            CancellationToken cancellationToken)
+        {
+            var binaryPath = Path.Combine(appInfo.AppPath, "Contents", "MacOS", appInfo.AppName);
+            var arguments = new List<string>();
+
+            if (!File.Exists(binaryPath))
+            {
+                _mainLog.WriteLine($"Failed to find an executable binary at {binaryPath}. Trying to run app using `open -W`");
```
akoeplinger, line 69: We can parse the Info.plist file in the appbundle and check `CFBundleExecutable` for a reliable way to get the binary name. I'm fine if we do this in a follow-up though :)

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
```diff
@@ -343,54 +372,150 @@ public class AppTester : AppRunnerBase
             }
         }
 
-        private MlaunchArguments GetCommonArguments(
-            int verbosity,
+        /// <summary>
+        /// Runs the MacCatalyst app via `open -W path.to.app`.
```
akoeplinger, line 376: nit: fix comment since we're no longer using `open`

akoeplinger (approved):

PR #474, opened 2021-02-23
Capture System.log when running MacCatalyst apps

### LINKED ISSUE #459: Collect MacCatalyst system logs

premun:
Resolves #459


src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
```diff
@@ -76,9 +91,18 @@ protected async Task<string> FindDevice(TestTargetOs target)
...
-            return _processManager.ExecuteCommandAsync(binaryPath, arguments, _mainLog, timeout, envVars, cancellationToken);
+            systemLog.StartCapture();
+
+            try
+            {
+                return await _processManager.ExecuteCommandAsync(binaryPath, arguments, _mainLog, timeout, envVars, cancellationToken);
+            }
+            finally
+            {
+                systemLog.StopCapture();
```
greenEkatherine, line 104: Does it stop logging in any case?

premun: What do you mean "any case"? Even when there's an exception? (then the answer would be yes)

greenEkatherine: No, the opposite case - without exception

premun: Yes, finally should do it in all cases

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
```diff
@@ -289,11 +289,10 @@ public class AppTester : AppRunnerBase
 
                 _mainLog.WriteLine("Starting test run");
 
-                var result = _processManager.ExecuteCommandAsync(mlaunchArguments, _mainLog, timeout, cancellationToken: cancellationToken);
-
+                var result = await _processManager.ExecuteCommandAsync(mlaunchArguments, _mainLog, timeout, cancellationToken: cancellationToken);
                 await testReporter.CollectSimulatorResult(result);
 
-                // cleanup after us
+                // Cleanup after us
```
mandel-macaque, line 295: ein?

src/Microsoft.DotNet.XHarness.iOS.Shared/ITestReporter.cs
```diff
@@ -9,20 +9,18 @@
...
 namespace Microsoft.DotNet.XHarness.iOS.Shared
 {
-
-    // interface that represents a class that know how to parse the results from an app run.
+    /// <summary>
+    /// Interface that represents a class that knows how to parse test results.
+    /// </summary>
     public interface ITestReporter
     {
-
```
mandel-macaque, line 16: Noise

src/Microsoft.DotNet.XHarness.iOS.Shared/ITestReporter.cs
```diff
@@ -9,20 +9,18 @@
...
         ILog CallbackLog { get; }
-
```
mandel-macaque, line 18: Noise

mandel-macaque (approved):
But i'd remove the noise in the merge.

src/Microsoft.DotNet.XHarness.iOS.Shared/ITestReporter.cs
premun: I would revert it if the file didn't have any other changes and had more than 30 lines but since it's being changed already, removing a blank line doesn't hurt? Or what are your thoughts on this?

PR #475, opened 2021-02-24
Make sure the MacCatalyst app's binary is executable

premun:
I found this problem when trying to run a runtime-built app on Helix

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
@@ -79,7 +79,17 @@ namespace Microsoft.DotNet.XHarness.Apple
             var binaryPath = Path.Combine(appInfo.AppPath, "Contents", "MacOS", appInfo.BundleExecutable ?? appInfo.AppName);
             var arguments = new List<string>();
 
-            if (!File.Exists(binaryPath))
+            if (File.Exists(binaryPath))
+            {
+                // We need to make the binary executable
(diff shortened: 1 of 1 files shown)


akoeplinger (approved):

akoeplinger:
I guess this is because we package the .app in a .zip for Helix upload and that doesn't preserve the executable bit?

premun:
> I guess this is because we package the .app in a .zip for Helix upload and that doesn't preserve the executable bit?

Yeah, I did `tar -czvf` myself and maybe did something worng there. I was wondering if it's the XHarness' responsibility to make sure it is executable first but maybe it makes sense to just do it?

akoeplinger:
Hmm if we're using a `tar.gz` then that should preserve the executable bit, I'm only aware of problems when using .zip.

It's probably harmless to just always run the `chmod` but it could point to some other infra issue that we should investigate.

premun:
Right, then I can re-package these apps and get rid of this problem. This app hasn't been zipped by the Arcade SDK but by me manually and put to a storage for our integration tests here so we might be fine in Arcade actually.

So I will start using **tar.gz** and won't merge this change yet

greenEkatherine (approved):

mandel-macaque (approved):

premun:
So the issue here is that my apps are actually already **tar.gz** wit hthe `+x` but when I create the Helix jobs from Windows, the executable information gets lost when we pack the Helix payload. So I will check this in so that we don't have issues with this cross platform scenario

PR #479, opened 2021-02-25
Allow to skip device-id for android install/run/uninstall

### LINKED ISSUE #433: Allow to omit device id if there is only one suitable device
New android commands `install`, `run`, `uninstall` require `device-id` as a mandatory parameter. In case when we have only one suitable device (i.e. with required architecture or app installed) it could be useful to omit device-id.

greenEkatherine:
Resolves #433

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -436,7 +436,8 @@ namespace Microsoft.DotNet.XHarness.Android
             }
             else if (devices.Count > 1)
             {
-                logger.LogError("There are more than one device with the app installed, please provide --device-id to choose the required one");
+                logger.LogError("There is more than one device with the app installed, please provide --device-id to choose the required one");
+
                 return null;
             }
             return devices.Keys.First();
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -431,13 +431,12 @@ namespace Microsoft.DotNet.XHarness.Android
             var devices = GetAllDevicesToUse(logger, apkRequiredProperty, propertyName);
             if (devices.Count == 0)
             {
-                logger.LogError("Cannot find a device with app installed, please check that a device is attached or run install first");
+                logger.LogError($"Cannot find a device with {propertyName}={apkRequiredProperty}, please check that a device is attached");
                 return null;
             }
             else if (devices.Count > 1)
             {
-                logger.LogError("There is more than one device with the app installed, please provide --device-id to choose the required one");
-
+                logger.LogError($"There is more than one device with {propertyName}={apkRequiredProperty}, please provide --device-id to choose the required one");
                 return null;
             }
             return devices.Keys.First();
@@ -465,12 +464,12 @@ namespace Microsoft.DotNet.XHarness.Android
 
             var result = allDevicesAndTheirProperties
                 .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Contains(apkRequiredProperty))
-                .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+                .ToDictionary(kvp => kvp.Key, kvp => kvp.Value) as Dictionary<string, string>;
             
             if (result.Count() == 0)
             {
                 // In this case, the enumeration worked, we found one or more devices, but nothing matched the APK's architecture; fail out.
-                logger.LogError($"No devices found with {propertyName} '{apkRequiredProperty}'.");
+                logger.LogError($"No devices with {propertyName} '{apkRequiredProperty}' was found among attached devices.");
             }
             return result;
         }
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -413,8 +413,38 @@ namespace Microsoft.DotNet.XHarness.Android
         }
 
 
+
         public string? GetDeviceToUse(ILogger logger, string apkRequiredProperty, string propertyName)
         {
+            var allDevicesAndTheirProperties = GetAllDevicesToUse(logger, apkRequiredProperty, propertyName);
+            if (allDevicesAndTheirProperties.Count > 0)
+            {
+                var firstAvailableCompatible = allDevicesAndTheirProperties.First();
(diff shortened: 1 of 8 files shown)


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -413,8 +413,38 @@ public List<string> PullFiles(string devicePath, string localPath)
...
+                logger.LogError("There are more than one device with the app installed, please provide --device-id to choose the required one");
+                return null;
+            }
+            return devices.Keys.First();
+        }
+
+        public Dictionary<string, string> GetAllDevicesToUse(ILogger logger, string apkRequiredProperty, string propertyName)
+        {
+
             Dictionary<string, string?> allDevicesAndTheirProperties = new Dictionary<string, string?>();
```
premun, line 448: I think there's a mismatch - `result` is `Dictionary<string, string?>` but the return type is `Dictionary<string, string>`.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -423,28 +453,25 @@ public List<string> PullFiles(string devicePath, string localPath)
...
-            }
-            else
+            var result = allDevicesAndTheirProperties
+                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Contains(apkRequiredProperty))
+                .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+            
+            if (result.Count() == 0)
             {
                 // In this case, the enumeration worked, we found one or more devices, but nothing matched the APK's architecture; fail out.
                 logger.LogError($"No devices found with {propertyName} '{apkRequiredProperty}'.");
```
premun, line 472: I would be more explicit here and say that we found devices but none with the property. I know it means that but I would be overexplicit in these cases

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
premun: I think maybe you could use a different variable when you re-create the `string, string` version in the `ToDictionary`

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -423,28 +453,25 @@ public List<string> PullFiles(string devicePath, string localPath)
...
-            if (allDevicesAndTheirProperties.Any(kvp => apkRequiredProperty.Equals(kvp.Value, StringComparison.OrdinalIgnoreCase)))
-            {
-                // Key-value tuples here are of the form <device serial number, device property>
-                KeyValuePair<string, string?> firstAvailableCompatible = allDevicesAndTheirProperties.FirstOrDefault(kvp => apkRequiredProperty.Equals(kvp.Value, StringComparison.OrdinalIgnoreCase));
-                logger.LogDebug($"Using first-found compatible device of {allDevicesAndTheirProperties.Count} total- serial: '{firstAvailableCompatible.Key}' - {propertyName}: {firstAvailableCompatible.Value}");
-                return firstAvailableCompatible.Key;
-            }
-            else
+            var result = allDevicesAndTheirProperties
+                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Contains(apkRequiredProperty))
```
premun, line 466: What characters are we splitting here by?
(I don't know what the method does called like this, so asking)

Can it happen that the default set of splitters contains a character that is also in the `apkRequiredProperty`? Because if for example a '-' or ':' was there, it wouldn't work

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -413,8 +413,38 @@ public List<string> PullFiles(string devicePath, string localPath)
...
+            }
+            return null;
+        }
+
+        public string? GetUniqueDeviceToUse(ILogger logger, string apkRequiredProperty, string propertyName)
+        {
+            var devices = GetAllDevicesToUse(logger, apkRequiredProperty, propertyName);
+            if (devices.Count == 0)
+            {
+                logger.LogError("Cannot find a device with app installed, please check that a device is attached or run install first");
```
premun, line 434: I think this error is too custom for the purpose of this method. I think the calling method knows it's looking for the app installed but this one only looks for any property?

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -413,8 +413,38 @@ public List<string> PullFiles(string devicePath, string localPath)
...
+                return null;
+            }
+            else if (devices.Count > 1)
+            {
+                logger.LogError("There are more than one device with the app installed, please provide --device-id to choose the required one");
```
premun, line 439:
```suggestion
                logger.LogError("There is more than one device with the app installed, please provide --device-id to choose the required one");

```

Again, could be too specific error. But maybe I understand it wrong?

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
premun: Yeah, looking lower at the usage - it's the caller of this who know we're looking for a "device with a specific app". I would consider removing this method and just doing the logic there (find all devices and if there's >1, throw the error from there). If it's used in multiple places, you could have an `AndroidCommandBase` that has this logic. But this is just my opinion and I will leave it on your decision. Fixing the error message is also fine

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
greenEkatherine: For the list of the installed app for example. It is splitting by whitespace-like characters, isn't it? At least I check for arm64-v8a arch and it wasn't splitted that way

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
greenEkatherine: Actually I filter out all null values, so I'm going to convert it to `Dictionary<string, string>` explicitely

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
greenEkatherine: I choose fixing the message 😃

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
premun: Yup, should be good then

premun (approved):

PR #484, opened 2021-02-26
Run MacCatalyst via `open w` and detect exit code from logs

### LINKED ISSUE #462: Detect exit code when calling `apple run` for MacCatalyst target

premun:
Before I dive into the investigations of why we get hangs when running the app's binary directly, we can have this solution that uses `open -w` and works in the Helix environment

This solution requires getting the exit code from the `system.log`

Resolves #462

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
@@ -76,26 +76,18 @@ namespace Microsoft.DotNet.XHarness.Apple
                 entireFile: false,
                 LogType.SystemLog);
 
+            // We need to make the binary executable
             var binaryPath = Path.Combine(appInfo.AppPath, "Contents", "MacOS", appInfo.BundleExecutable ?? appInfo.AppName);
-            var arguments = new List<string>();
-
             if (File.Exists(binaryPath))
             {
-                // We need to make the binary executable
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
```diff
@@ -76,26 +76,18 @@ protected async Task<string> FindDevice(TestTargetOs target)
...
             if (File.Exists(binaryPath))
             {
-                // We need to make the binary executable
-                var result = await _processManager.ExecuteCommandAsync("chmod", new[] { "+x", binaryPath }, _mainLog, TimeSpan.FromSeconds(10), cancellationToken: cancellationToken);
-
-                if (!result.Succeeded)
-                {
-                    _mainLog.WriteLine($"Failed to make the binary at {binaryPath} executable, the run might fail");
-                }
+                await _processManager.ExecuteCommandAsync("chmod", new[] { "+x", binaryPath }, _mainLog, TimeSpan.FromSeconds(10), cancellationToken: cancellationToken);
```
mandel-macaque, line 83: I'd do a 'a+x'

mandel-macaque (approved):

src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
premun: I think the smallest possible change to the file system that will make it is better?

Issue #473, opened 2021-02-22
Wasm browser test runs intermittently fail to read test results


steveisok:
We've recently noticed xharness fail to read results from in the browser wasm test runs.  This has resulted in builds being flagged in dotnet/runtime as having failures, but the logs say otherwise.  

If you notice on [build](https://github.com/dotnet/runtime/pull/48593) in question, the log below does _looks_ ok, but is still flagged as in error:

https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-48593-merge-840f229ea75c4f63ad/System.IO.Pipelines.Tests/console.318c6533.log?sv=2019-07-07&se=2021-03-14T14%3A15%3A06Z&sr=c&sp=rl&sig=V3V1rKZcGBWXTNCyBWw3cnrLshSmLfjGA96KBU9nE8c%3D

It might have something to do with either an error in xml processing or some interference in finding the special log message we need from stdout.  

Here's the xml error message we've seen in some cases:

`xml.etree.ElementTree.ParseError: mismatched tag: line 21883, column 6`

radical:
I'm not being able to see how to get the `xharness-output` contents for this. That should have the console log, which might be helpful.

radical:
Ok got the log. `wast-console.log` ends with:

```
=== TEST EXECUTION SUMMARY ===
Total: 340, Errors: 0, Failed: 0, Skipped: 43, Time: 24.1354s

IT 0
```

I think that last line is part of `EXIT 0`.

radical:
@steveisok do you have links to any builds with the xml error?
Meanwhile, I'm adding a fallback for the `IT 0` issue.

PR #489, opened 2021-03-02
Don't reset the stream if we got a partial message

lewing:
We were (accidentally I presume) clearing the message stream on every receipt instead of only when EndOfMessage was set, that means if the message was sent in multiple parts (or if it was too long to fit in buff) we'd only process the final part.  This moves the stream resetting to the EndOfMessage case so that we'll get the whole message.

This would show up in the logs as corrupted messages where only the trailing part of the string was present.

Fixes https://github.com/dotnet/xharness/issues/473


premun:
@lewing you will have to merge main in order for the CI legs to match the branch policies. Sorry for the inconvenience

lewing:
@premun merged master, let me know if you need anything else.

akoeplinger (approved):

PR #491, opened 2021-03-02
Do not report LaunchFailure when the app connected over TCP

### LINKED ISSUE #483: Infrastructure retry/reboot happens even when a test times out
**Context**
Originally, we were returning the exit code 83 when we failed to install the app (the simulator would hang when booting).
The 83 then results in an infra retry + machine reboot which would fix the hanging simulator.

**Problem**
I noticed that when we succeed in installing the app and the tests start running but they hang themselves (endless running test), we also return 83 (app launch failure) and we also retry+reboot which is unnecessary because the same will happen on the next run.
We are already detecting when the app connects to XHarness over TCP so we are probably just merging two results in one exit code or something.

**Notes**
Example logs:
https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-heads-master-ae9502d98e3a486aa0/System.Net.WebSockets.Client.Tests.Attempt.3/console.6d14f56c.log?sv=2019-07-07&se=2021-03-15T13%3A38%3A14Z&sr=c&sp=rl&sig=bj0C2RJFi4Y5Ogl76ueW0vNjnnLdyJBfp7Wx93D19VI%3D

```
[09:40:48] dbug: Xamarin.Hosting: Launching net.dot.System.Net.WebSockets.Client.Tests async on 'iOS 13.5 (17F61) - iPhone X'
[09:40:48] dbug: Xamarin.Hosting: Launched net.dot.System.Net.WebSockets.Client.Tests with pid 15781
[09:41:00] dbug: Connection from 127.0.0.1:63509 saving logs to /tmp/helix/working/AEBC094A/w/BF8C0A69/uploads/test-ios-simulator-64-20210223_094044.log
[09:41:00] dbug: Test execution started
[10:10:44] dbug: Killing process 15778 as it was cancelled
[10:10:44] dbug: Process mlaunch exited with 137
[10:10:44] dbug: Test run failed
[10:10:44] dbug: Completion timed out after 1797.7718879 seconds
[10:10:49] dbug: Test log server listening on: 0.0.0.0:63508
[10:10:50] dbug: Test run failed to launch
[10:10:50] dbug: No crash reports, waiting 5 seconds for the crash report service...
[10:10:56] fail: Failed to launch the application:
                 Launch failure
                 
                 Check logs for more information.
XHarness exit code: 83
```

Or check Kusto for failed work items and see the logs. It usually happens for `System.Net.WebSockets.Client.Tests` or `System.Net.WebClient.Tests`:
```
WorkItems
| where QueueName == "osx.1015.amd64.open" and Started > now()-1d and (ExitCode == 83 or ExitCode == 123)
| project JobName, MachineName, FriendlyName, Status, ExitCode, ConsoleUri, Started
| order by Started desc
```

premun:
Resolves #483

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
@@ -112,7 +112,8 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared
 
             CallbackLog = new CallbackLog(line =>
             {
-                // MT1111: Application launched successfully, but it's not possible to wait for the app to exit as requested because it's not possible to detect app termination when launching using gdbserver
+                // MT1111: Application launched successfully, but it's not possible to wait for the app to exit as
+                // requested because it's not possible to detect app termination when launching using gdbserver
                 _waitedForExit &= line?.Contains("MT1111: ") != true;
                 if (line?.Contains("error MT1007") == true)
                 {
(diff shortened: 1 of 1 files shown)


premun:
@rolfbjarne FYI thre is technically a change of behavior in this PR but - report crash instead of launch failure

mandel-macaque (approved):
Looks cleaner

Issue #493, opened 2021-03-03
Support passing environment variables to iOS apps launched by xharness, and support separate install, run-test, and uninstall subcommands


imhameed:
I'm working on getting CoreCLR's runtime tests working with Mono on iOS devices. In https://github.com/dotnet/runtime/pull/43954, I'm bundling multiple runtime test assemblies (with one entry point each) together into a single iOS app bundle, similar to what we currently do on Android. I'm passing the entry point assembly to run through an environment variable (unlike Android, which uses a command line argument): https://github.com/dotnet/runtime/pull/43954/files#diff-6cef647106df3b37e45599e150f97a7dc3678f2606723b1967dbda7d1f0a42b9R268-R273

I've verified that doing this works locally when launching runtime test assemblies (via their containing iOS app bundle) via mlaunch. To make this run for real on CI I'll need to do this with xharness.

xharness doesn't appear to have a way to add passthrough environment variables, although it does appear to support mlaunch's `--setenv` option internally:
https://github.com/dotnet/xharness/blob/main/src/Microsoft.DotNet.XHarness.iOS.Shared/Execution/Arguments.cs#L199-L222

There was also a request in https://github.com/dotnet/xharness/issues/397 to add separate install, run-test, and uninstall commands to xharness for Android. That will also be necessary for iOS, because a single iOS runtime test app bundle will contain multiple entry points.

premun:
@imhameed sure, this should be a piece of cake!

Are env variables fine or do you need to pass arguments to the app directly as well (we support this already but not sure it works on the mlaunch side)?

premun:
@imhameed can you also please provide me with a test application and some instructions on how you would like to run the tests using it so that I have a good picture of things should work?

imhameed:
> @imhameed sure, this should be a piece of cake!

Great!

> I have some questions:
> 
> 1. Are env variables fine or do you need to pass arguments to the app directly as well (we support this already but not sure it works on the mlaunch side)?
> 
> 2. For the `run-test`, `install`/`uninstall` flows, I can follow the flow from Android?
> 
> 3. I know that for Simulators there's no install/uninstall, we just run the app so we won't need to get the device ID but I need to look into whether we can just run the app without having to have to copy it there repeatedly. I imagine we might keep the same flow for device/Simulator though so that it's not confusing where in/uninstall will be a no-op

1. I don't need to pass arguments to the app right now. And I think mlaunch does support this.
2. Yes.
[...]

PR #496, opened 2021-03-05
Pass extra arguments as environmental variables too

premun:
The extra arguments added after the `--` verbatim string are now also passed as environmental variables:
```
xharness apple test -a=some.app -o=o -t=ios-simulator-64 --set-env=bar=xyz -- --foo=bar
```
ends up as (mlaunch output):
```
Launching net.dot.System.Numerics.Vectors.Tests async on 'iOS 14.2 (18B79) - iPhone X' with: {
    arguments =     (
        "--foo=bar",
        foo3
    );
    environment =     {
        NSUnbufferedIO = YES;
        "NUNIT_AUTOEXIT" = true;
        "NUNIT_ENABLE_XML_OUTPUT" = true;
        "NUNIT_HOSTNAME" = "127.0.0.1";
        "NUNIT_HOSTPORT" = 53423;
        "NUNIT_XML_VERSION" = xUnit;
        "OS_ACTIVITY_DT_MODE" = YES;
        bar = xyz;
    };
}
```

Same for `apple run` and for MacCatalyst apps.

Related to #493


src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
```diff
@@ -89,22 +87,65 @@ protected async Task<string> FindDevice(TestTargetOs target)
...
+
+                    // Remove initial --
+                    while (name.StartsWith("-"))
+                    {
+                        name = name.Substring(1);
+                    }
+
+                    if (envVariables.ContainsKey(name))
+                    {
+                        _mainLog.WriteLine($"Environmental variable {name} is already passed to the application to drive test run, skipping..");
```
greenEkatherine, line 143: I guess you should `continue` here. Otherwise logging could be wrong because value was updated and not skipped

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
```diff
@@ -428,17 +435,17 @@ public class AppTester : AppRunnerBase
 
                 envVariables[EnviromentVariables.HostName] = "127.0.0.1";
 
+                AddExtraEnvVars(envVariables, _appArguments);
+
                 var arguments = new List<string>
```
greenEkatherine, line 440: It seems that `arguments` is unused variable

premun: Nice catch!

src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
premun: Even nice-er catch!

premun:
@imhameed can you please check the behavior in the description and let me know if this will resolve some of the issues from #493?

greenEkatherine (approved):

mandel-macaque (approved):

akoeplinger:
I don't like this, why are we not adding a separate `--setenv` option instead of passing all command line arguments?

premun:
@akoeplinger we are, the output in the description is from mlaunch already (which received bunch of setenvs) and this is the log statement from when it starts the iOS app

premun:
@akoeplinger oh, do you mean not passing all of the extra arguments as env variables? But have `--setenv` on XHarness too?

akoeplinger:
@premun yes

premun:
@akoeplinger done, updated the description and added tests

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
```diff
@@ -450,28 +458,29 @@ public class AppTester : AppRunnerBase
...
-            string listenerTmpFile)
+            string listenerTmpFile,
+            IEnumerable<(string, string)> extraEnvVariables)
         {
-            var variables = new Dictionary<string, object>
+            var variables = new Dictionary<string, string>
             {
-                { EnviromentVariables.AutoExit, true },
-                { EnviromentVariables.HostPort, listenerPort },
+                { EnviromentVariables.AutoExit, "true" },
```
mandel-macaque, line 472: Why did we move to strings? I fear parsing annoying code.

mandel-macaque (approved):

src/Microsoft.DotNet.XHarness.Apple/AppTester.cs
premun: @mandel-macaque they become strings eventually, only that I found that booleans become capitalized (`True`/`False`) so it seems to have more control over this from the beginning

imhameed:
Sorry for not reviewing this earlier. I'm working on the iOS x64 simulator stuff again. I just checked out a copy of xharness (3f5f7deef8ceb1e624adce3ed537cdf392d3ada5) and built it locally. Here's what happens when I try to run a runtime test using this locally-built copy of xharness and `--set-env`:

```
imhameed@ms-ih-mbp ~/ms/dnr-9.x-ios/runtime/artifacts/tests/coreclr/obj/iOS.x64.Release/tests/run/iOSApps/baseservices_mono/AppBundle/baseservices_mono/Release-iphonesimulator$ xharness.sh apple run --app=baseservices_mono.app --output-directory=out --targets=ios-simulator-64 --mlaunch=/Library/Frameworks/Xamarin.iOS.framework/Versions/14.10.0.4/bin/mlaunch --set-env=MONO_APPLE_APP_ENTRY_POINT_LIB_NAME=runningmono.dll
XHarness command issued: apple run --app=baseservices_mono.app --output-directory=out --targets=ios-simulator-64 --mlaunch=/Library/Frameworks/Xamarin.iOS.framework/Versions/14.10.0.4/bin/mlaunch --set-env=MONO_APPLE_APP_ENTRY_POINT_LIB_NAME=runningmono.dll
fail: The set-env argument set-env must be in the key=value format
XHarness exit code: 3 (INVALID_ARGUMENTS)
```

premun:
Oops, sorry! There was a missing `=`.

Fix is here: https://github.com/dotnet/xharness/pull/521

Issue #478, opened 2021-02-25
xharness should also download/install the versions of the simulators


mattleibow:
It would be nice if there was a way to install simulators for iOS as all the info to the URL is there:

https://github.com/xamarin/xamarin-macios/tree/main/tools/siminstaller

I would maybe do something like:

```
xharness apple simulators install ios-simulator-64_10.3.1
```

When doing this for Android, it sort of fits in with #380 and #393 but this just calls upon the sdkmanager to also download.

premun:
There is some code for this in this repo already which is packaged into another .NET tool and can do this:
- https://github.com/dotnet/xharness/tree/main/tools/Microsoft.DotNet.XHarness.SimulatorInstaller
- https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet-eng&package=Microsoft.DotNet.XHarness.SimulatorInstaller&protocolType=NuGet&version=1.0.0-prerelease.21124.1

This tool is actually originally coming from the siminstaller you've linked, just it was re-written.

We could easily move these commands into XHarness CLI itself. @tkapin thoughts?

PR #497, opened 2021-03-05
Merge SimulatorInstaller commands to main CLI

premun:
This removes the SimulatorInstaller project and moves the file to CLI so that you can do:
```
xharness apple simulators list
```

When installing the simulators, it doesn't accept the same format e.g. `ios-simulator-64` of targets as the CLI, this will be a follow-up.

#478


src/Microsoft.DotNet.XHarness.CLI/Program.cs
```diff
@@ -28,27 +28,18 @@ public static int Main(string[] args)
...
 
-            // Root command: will use the platform specific commands to perform the appropriate action.
-            var commands = new CommandSet("xharness");
-
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
-            {
-                commands.Add(new AppleCommandSet());
-            }
-            else if (args[0] == "apple")
+            if (args[0] == "apple" && RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
mattleibow, line 31: This throws when it is just `dotnet xharness`:

```suggestion
            if (args.Length > 0 && args[0] == "apple" && !RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```

src/Microsoft.DotNet.XHarness.CLI/Commands/XHarnessHelpCommand.cs
```diff
@@ -20,6 +21,13 @@ public override int Invoke(IEnumerable<string> arguments)
         {
             string[] args = arguments.ToArray();
 
+            // TODO (#400): We can remove this after some time when users get used to the new commands
+            if (args[0] == "ios")
```
mattleibow, line 25:
```suggestion
            if (args.Length > 0 && args[0] == "ios")
```

src/Microsoft.DotNet.XHarness.CLI/Program.cs
```diff
@@ -28,27 +28,18 @@ public static int Main(string[] args)
                 args[0] = "apple";
```
mattleibow, line 28: The line above this also needs to be a Length > 0 as `dotnet ios` will fail without a warning as to the change.

mattleibow (commented):
There are some argument testing that seems a little off, but otherwise seems to be working.

Thanks for this as we have been using all sorts of scripts and things to avoid having to add more unrelated code to repos. This is really going to help keep things clean.

premun:
@greenEkatherine we would need this PR in because it changes names of PR legs and the branch policies had to change.

Can you please have a look at this PR?

greenEkatherine (approved):

PR #509, opened 2021-03-18
[wasm] Search js engines in PATH

radekdoulik:
The `Process.Start` doesn't find our JS engines on Windows when

    process.StartInfo.UseShellExecute = false;

is set. So we need to pass full path to the js engine cmd script on Windows.

Otherwise we endup with errors like this one:

    XHarness command issued: wasm test --app=. --engine=V8 --engine-arg=--stack-trace-limit=1000 --js-file=runtime.js --output-directory=C:\Users\rodo\git\runtime\artifacts\bin\System.Collections.Concurrent.Tests\net6.0-Debug\browser-wasm\AppBundle\xharness-output -- --run WasmTestRunner.dll System.Collections.Concurrent.Tests.dll -notrait category=OuterLoop -notrait category=failing
    info: 22:49:48.1379578 Running v8 --expose_wasm --stack-trace-limit=1000 runtime.js -- --run WasmTestRunner.dll System.Collections.Concurrent.Tests.dll -notrait category=OuterLoop -notrait category=failing
    crit: The engine binary `v8` was not found
    XHarness exit code: 83 (APP_LAUNCH_FAILURE)

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Runtime.InteropServices;
 using System.IO;
 using System.Threading.Tasks;
 using Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm;
@@ -31,6 +32,30 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
         {
         }
 
+        static readonly string[] s_extensions = { ".cmd", ".exe" };
+
+        private static string GetFullPathTo(string engineBinary)
+        {
+            var path = Environment.GetEnvironmentVariable("PATH");
+            if (path == null)
+                return engineBinary;
+
+            foreach (var folder in path.Split(";"))
+            {
+                foreach (var ext in s_extensions)
+                {
+                    var fullPath = Path.Combine(folder, engineBinary + ext);
+                    if (File.Exists(fullPath))
+                    {
+                        engineBinary = fullPath;
+                        break;
+                    }
+                }
+            }
+
+            return engineBinary;
+        }
+
         protected override async Task<ExitCode> InvokeInternal(ILogger logger)
         {
             var processManager = ProcessManagerFactory.CreateProcessManager();
@@ -42,6 +67,9 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
                 JavaScriptEngine.SpiderMonkey => "sm",
                 _ => throw new ArgumentException()
             };
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                engineBinary = GetFullPathTo(engineBinary);
 
             var engineArgs = new List<string>();
 
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
@@ -32,11 +32,9 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
         {
         }
 
-        static readonly string[] s_extensions = { ".cmd", ".exe" };
-
-        private static string GetFullPathTo(string engineBinary)
+        private static string FindEngineInPath(string engineBinary)
         {
-            if (Path.IsPathRooted(engineBinary))
+            if (File.Exists (engineBinary) || Path.IsPathRooted(engineBinary))
                 return engineBinary;
 
             var path = Environment.GetEnvironmentVariable("PATH");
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -43,6 +68,9 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
                 _ => throw new ArgumentException()
             };
 
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                engineBinary = GetFullPathTo(engineBinary);
```
radical, line 72: This should probably not do this if the path is already rooted

radekdoulik: The `engineBinary` is set few lines above and is not rooted. So hopefully we don't need that?

radical: hah! ignore me!

radekdoulik: OTOH, it is easy to do, so I will handle it just in case.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -31,6 +32,34 @@ public WasmTestCommand() : base("test", true, CommandHelp)
...
+        {
+            if (Path.IsPathRooted(engineBinary))
+                return engineBinary;
+
+            var path = Environment.GetEnvironmentVariable("PATH");
+
+            if (path == null)
+                return engineBinary;
+
+            foreach (var folder in path.Split(";"))
```
premun, line 47: I believe this separator differs per platform. There should be something like `Path.PathSeparator` or similar that would give you the right one

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -31,6 +32,34 @@ public WasmTestCommand() : base("test", true, CommandHelp)
         {
         }
 
+        static readonly string[] s_extensions = { ".cmd", ".exe" };
+
+        private static string GetFullPathTo(string engineBinary)
+        {
+            if (Path.IsPathRooted(engineBinary))
```
premun, line 39: Should we also check `File.Exists(engineBinary)` for cases where the binary is in the current working directory or we sent a relative path to the binary?
(and root the path in that case and return)

premun (changes requested):

akoeplinger:
Hmm according to https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.useshellexecute?view=net-5.0#workingdirectory we should _already_ be looking in PATH for the executable:

> When UseShellExecute is false, the FileName property can be either a fully qualified path to the executable, or a simple executable name that the system will attempt to find within folders specified by the PATH environment variable.

I wonder if this is because we specify `v8` but the actual executable is called `v8.exe` or `v8.cmd`? would you mind testing if it works if you append the extension?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
radekdoulik: This is only used on windows, it will be cleaner though.

radekdoulik:
> 
> 
> Hmm according to https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.useshellexecute?view=net-5.0#workingdirectory we should _already_ be looking in PATH for the executable:
> 
> > When UseShellExecute is false, the FileName property can be either a fully qualified path to the executable, or a simple executable name that the system will attempt to find within folders specified by the PATH environment variable.
> 
> I wonder if this is because we specify `v8` but the actual executable is called `v8.exe` or `v8.cmd`? would you mind testing if it works if you append the extension?

Tried that and it doesn't work. I guess it would work with binaries like `.exe`, but doesn't with `.cmd`?

```
    XHarness command issued: wasm test --app=. --engine=SpiderMonkey --js-file=runtime.js --output-directory=C:\Users\rodo\git\runtime\artifacts\bin\System.Collec
  tions.Concurrent.Tests\net6.0-Debug\browser-wasm\AppBundle\xharness-output -- --run WasmTestRunner.dll System.Collections.Concurrent.Tests.dll -notrait category
  =OuterLoop -notrait category=failing
    info: 14:24:31.8300817 Running sm.cmd runtime.js --run WasmTestRunner.dll System.Collections.Concurrent.Tests.dll -notrait category=OuterLoop -notrait categor
  y=failing


    fail: 14:24:31.8647697 The system cannot find the path specified.

    info: 14:24:31.8662239 Process sm.cmd exited with 1

    fail: Application has finished with exit code 1 but 0 was expected
    XHarness exit code: 71 (GENERAL_FAILURE)
    ----- end Fri 03/19/2021 14:24:31.88 ----- exit code  ----------------------------------------------------------
    XHarness artifacts: "C:\Users\rodo\git\runtime\artifacts\bin\System.Collections.Concurrent.Tests\net6.0-Debug\browser-wasm\AppBundle\xharness-output"
```

radical (approved):

premun (approved):

Issue #456, opened 2021-02-17
Move XHarness onto .NET 6


premun:
Currently, it is a .NET 3.1 tool. In order to update this, we need:
- [ ] Update the projects themselves
- [x] Update the Helix SDK
- [ ] Update the bootstrap script

It is unclear how we're going to do this since the PR build will fail most likely because the XHarness SDK will try to install it with 3.1 so we might as well ignore the failed build and merge.

PR #516, opened 2021-03-22
Multi-target XHarness at .NET 3.1 and 6.0

premun:
- Adds the .NET 6.0 target to XHarness CLI
  - We need both first so that Helix SDK doesn't break, switch it there and remove 3.1 after
- Changes libraries from `netstandard2.1` to `net6.0` and multitargets to both where needed (Apple, Android)
- Fixes warnings that would break the build
  - WebClient is deprecated
  - Nullability issues
- Resolves most of the build messages to clean the solution
  - Naming
  - Static methods where possible
  - Calling `GC.SuppressFinalize`
  - using `new ()`

#456

### DIFF
--- a/src/Microsoft.DotNet.XHarness.TestRunners.Common/Microsoft.DotNet.XHarness.TestRunners.Common.csproj
+++ b/src/Microsoft.DotNet.XHarness.TestRunners.Common/Microsoft.DotNet.XHarness.TestRunners.Common.csproj
@@ -6,6 +6,7 @@
     <!-- Mono.Options is apparently not strong-name signed -->
     <NoWarn>CS8002;</NoWarn>
     <RootNamespace>Microsoft.DotNet.XHarness.TestRunners.Common</RootNamespace>
+    <LangVersion>9.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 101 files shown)


src/Microsoft.DotNet.XHarness.TestRunners.Common/Microsoft.DotNet.XHarness.TestRunners.Common.csproj
```diff
@@ -1,11 +1,13 @@
...
 
   <PropertyGroup>
+    <TargetFrameworks>netstandard2.1;net6.0</TargetFrameworks>
     <IsPackable>true</IsPackable>
-    <TargetFrameworks>netstandard2.1;netcoreapp3.1</TargetFrameworks>
+    <LangVersion>9.0</LangVersion>
     <!-- Mono.Options is apparently not strong-name signed -->
     <NoWarn>CS8002;</NoWarn>
     <RootNamespace>Microsoft.DotNet.XHarness.TestRunners.Common</RootNamespace>
+    <LangVersion>9.0</LangVersion>
```
greenEkatherine, line 10: duplication?

greenEkatherine (approved):

PR #517, opened 2021-03-23
Accept test targets when working with the simulators command

### LINKED ISSUE #478: xharness should also download/install the versions of the simulators

premun:
XHarness simulators command now also accepts target runtimes in the same format as test targets. These formats are now equivalent:
- `com.apple.pkg.iPhoneSimulatorSDK14_3`
- `ios-simulator_14.3`

There is also a breaking change for the `install` and `find` commands - we now accept the list of simulators as positional arguments instead of `--simulator` arguments. The new usage is this:
```
xharness apple simulators install ios-simulator_14.3 tvos-simulator_11.6
```

Resolves #478

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Apple/Simulators/FindCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Apple/Simulators/FindCommandArguments.cs
@@ -2,30 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
 using Mono.Options;
 
 namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Apple.Simulators
 {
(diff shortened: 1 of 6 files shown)


mattleibow:
The other commands also allowed for the bit size, so not sure if that is relevant or needs consistency :

ios-simulator-64_14.3

This might be relevant for watch things since I am not sure if any other architectures exist other than 64

premun:
@mattleibow it doesn't matter for the package names, I think you just need to install old enough version that are 32-bit if I am not mistaken.

However, I also allowed to enter `ios-simulator-64` and `-32` so that you can have it easy in the build and call both commands the same way:
```
xharness apple simulators install ios-simulator-64_14.2
xharness apple test --target=ios-simulator-64_14.2 ...
```

greenEkatherine (approved):

PR #520, opened 2021-03-24
Fix configuration of test filtering in Android application test runner

elinor-fung:
- Application entry points: configure test runner with current application options
  - Android and iOS application entry points were running the tests and then trying to configure filters for the test run afterwards.
- XUnit test case filtering:
  - Handle different test case filter types separately
  - Use added filters to determine the default for include/exclude
- Add tests for XUnit test case filtering

cc @steveisok @jkoritzinsky

### DIFF
--- a/src/Microsoft.DotNet.XHarness.TestRunners.Common/ApplicationEntryPoint.cs
+++ b/src/Microsoft.DotNet.XHarness.TestRunners.Common/ApplicationEntryPoint.cs
@@ -5,7 +5,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Threading.Tasks;
 
 namespace Microsoft.DotNet.XHarness.TestRunners.Common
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.XHarness.TestRunners.Common/ApplicationEntryPoint.cs
```diff
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
```
akoeplinger, line 8: it would be nice to avoid the System.Linq dependency

elinor-fung: Switched the properties on the internal `ApplicationOptions` to be `ICollection` instead of `IEnumerable` and removed the System.Linq dependency.

premun (approved):

elinor-fung:
@premun do you think this is ready to merge, or do we need to wait for other reviews? I'd like to get this in soon, since it really helps with the inner dev loop for running Android tests. (I don't have merge permissions)

PR #523, opened 2021-03-26
Accept mlaunch path in env variable

premun:
This improves the dev inner loop when using XHarness CLI binary directly (not through the tool). You can specify the mlaunch path easily.

Also adds documentation for XHarness developers.


premun:
FYI @akoeplinger @imhameed

greenEkatherine (approved):

PR #527, opened 2021-03-29
Remove the .NET 3.1 target

### LINKED ISSUE #456: Move XHarness onto .NET 6

premun:
Also bump Arcade

Resolves #456

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,19 +157,18 @@ You can also use Visual Studio 2019+ and just F5 the `Microsoft.DotNet.XHarness.
 In order for XHarness to work, you will need ADB (for Android) and mlaunch (for anything Apple).
 These are executables that go with the packaged .NET xharness tool.
 
-The easiest way to get these at the moment for development purposes is to run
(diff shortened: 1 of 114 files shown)


greenEkatherine (approved):

premun:
@mattleibow @PureWeen XHarness currently targets both 3.1 and 6.0. When I merge this, you will need .NET 6 (preview 2) to run XHarness.
I don't want to merge this before I know that I won't break your pipelines. So let me know and we can do the switch.
Or let me know if you want to keep 3.1 for some reason too.

The exact (released) version of the .NET SDK is `6.0.100-preview.2.21155.3`

PureWeen:
@premun that's fine

We're currently running all our MAUI tests using NET6 bits

PR #530, opened 2021-03-30
Add `apple device` command

premun:
Adds a command that searches for a suitable simulator/device and returns its UDID

#493

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
@@ -52,7 +52,7 @@ namespace Microsoft.DotNet.XHarness.Apple
         protected async Task<string> FindDevice(TestTargetOs target)
         {
             IHardwareDevice? companionDevice = null;
-            IHardwareDevice device = await _hardwareDeviceLoader.FindDevice(target.Platform.ToRunMode(), _mainLog, includeLocked: false, force: false);
+            IHardwareDevice device = await _hardwareDeviceLoader.FindDevice(target.Platform.ToRunMode(), _mainLog, includeLocked: false);
 
             if (target.Platform.IsWatchOSTarget())
             {
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleGetDeviceCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleGetDeviceCommand.cs
@@ -1,0 +1,104 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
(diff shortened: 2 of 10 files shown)


src/Microsoft.DotNet.XHarness.CLI/Program.cs
```diff
@@ -85,5 +94,26 @@ public static CommandSet GetXHarnessCommandSet()
...
+            if (args.Length < 2 || args.Contains("--help") || args.Contains("-h"))
+            {
+                return false;
+            }
+
+            switch (args[0])
+            {
+                case "ios":
+                case "apple":
+                case "android":
```
premun, line 112: @greenEkatherine this will change the behaviour of the `android device` command so that the output is easier to parse. But we should probably let the users (Fan Yang) know about this.

Do you want this change for Android?

greenEkatherine: yes, I think it would be great, I will simplify the command in unit tests as well. Do we have a channel for such announcements or should I notify her directly?

premun: I don't think they are aware of the Teams channel we created for mono/Xamarin, so maybe we would need to reach out to her. I would merge this then so that you can use this change in your integration tests?

src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
```diff
@@ -52,7 +52,7 @@ public abstract class AppRunnerBase
         protected async Task<string> FindDevice(TestTargetOs target)
         {
             IHardwareDevice? companionDevice = null;
-            IHardwareDevice device = await _hardwareDeviceLoader.FindDevice(target.Platform.ToRunMode(), _mainLog, includeLocked: false, force: false);
```
greenEkatherine, line 55: why do you remove this please?

src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleGetDeviceCommand.cs
```diff
@@ -0,0 +1,116 @@
...
+                logger.LogError(e.ToString());
+                return ExitCode.DEVICE_NOT_FOUND;
+            }
+
+            return ExitCode.SUCCESS;
+        }
+
+        private TestTargetOs ParseTarget()
+        {
+            if (ExtraArguments.Count() != 1)
```
greenEkatherine, line 85: probably `<1` would be better? or you want to validate that there is no more other arguments?

premun: Yeah, I only want 1. I will improve the message

src/Microsoft.DotNet.XHarness.Apple/AppRunnerBase.cs
premun: I noticed it was not doing anything in the implementation so it's probably a leftover from the old XHarness code

greenEkatherine (approved):

PR #538, opened 2021-04-06
Separate install and run operations for iOS apps

premun:
In order to introduce the new `install` / `run-test` / `uninstall` commands (#493), we need to separate installation from running the app which is currently happening via 1 mlaunch command for Simulators.

We unify the approach for Simulators and devices and do the install separately. This will also help with diagnosis of issues when simulators freeze and we don't know at which stage.

The code was refactored out of the `Command` classes into the `Orchestrator` classes so that it can be tested later too.
We can also now test passing args to the `Commands` and how they get parsed.


premun:
The new command for launching Simulator app without installing will need a new mlaunch version so the CI will fail before we get it.
The PR for mlaunch fix was opened already.

src/Microsoft.DotNet.XHarness.iOS.Shared/Execution/MlaunchProcessManager.cs
```diff
@@ -85,6 +85,11 @@ public MlaunchProcessManager(string? xcodeRoot = null, string mlaunchPath = "/Li
                 args = new MlaunchArguments(args.Prepend(new SdkRootArgument(XcodeRoot)).ToArray());
             }
 
+            for (var i = 0; i < 5; i++)
```
greenEkatherine, line 88: looks like magic number, could you please replace it with verbose constant?

src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleOrchestrator.cs
```diff
@@ -0,0 +1,366 @@
...
+            }
+
+            exitCode = await InstallApp(appBundleInfo, device, target, cancellationToken);
+
+            if (exitCode != ExitCode.SUCCESS)
+            {
+                if (!target.Platform.IsSimulator())
+                {
+                    _logger.LogInformation($"Cleaning up the failed installation from '{device.Name}'");
+                    await UninstallApp(appBundleInfo, device, new CancellationToken());
```
greenEkatherine: why not uninstalling for simulators?

premun: mlaunch just cannot do it, I am not sure what is the reason behind that

src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleOrchestrator.cs
```diff
@@ -0,0 +1,366 @@
...
+                {
+                    _logger.LogInformation($"Starting '{appBundleInfo.AppName}' on MacCatalyst");
+
+                    return await ExecuteMacCatalystApp(arguments, passthroughArguments, appBundleInfo, cancellationToken);
+                }
+                catch (Exception e)
+                {
+                    var message = new StringBuilder().AppendLine("Application run failed:");
+
+                    if (_errorKnowledgeBase.IsKnownTestIssue(_mainLog, out var failureMessage))
```
greenEkatherine: would it be better to add `LogProblem` method as for `TestOrchestrator`?

premun: It sort of works nicely as a local function since it captures some local variables that I would have to re-declare again

greenEkatherine (approved):

Issue #531, opened 2021-03-31
Improve logging for android


greenEkatherine:
Consider increase logging by following:

- Run `logcat -c` at the beginning of a test execution to clear the log
- Capture the pid from `shell am` instrument (which may not be possible?) and filter by `logcat --pid`. This is probably worse as there may be other things in the logcat log which are interesting outside the context of the test process

cc @billwert

greenEkatherine:
The first part was done:
> Run logcat -c at the beginning of a test execution to clear the log

For the second part I would like to estimate how many log lines we would filter out. On my tests I got about 1%. Probably logs are now better because of the proper cleaning. @billwert could you please also check how logs are look like now and is there still necessity to add the filter by pid?

greenEkatherine:
Confirmed offline with Bill that it can be closed now

PR #542, opened 2021-04-08
Fix Android logging for test and run commands

greenEkatherine:
First part of #531

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidRunCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidRunCommand.cs
@@ -73,9 +73,6 @@ Arguments:
             // Wait til at least device(s) are ready
             runner.WaitForDevice();
 
-            // Empty log as we'll be uploading the full logcat for this execution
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidRunCommand.cs
```diff
@@ -104,6 +101,9 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
 
             int instrumentationExitCode = (int)ExitCode.GENERAL_FAILURE;
 
+            // Empty log as we'll be uploading the full logcat for this execution
```
premun, line 104: This `InvokeHelper` is called from within different commands and that's why it was moved?

premun (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidRunCommand.cs
greenEkatherine: yes, from `test` as well

PR #551, opened 2021-04-15
Add new apple commands for separate app un/installation

premun:
Related to #493


imhameed (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleInstallCommand.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Microsoft.DotNet.XHarness.CLI.Commands.Apple
+{
+    internal class AppleInstallCommand : AppleAppCommand<AppleInstallCommandArguments>
+    {
+        private const string CommandHelp = "Installs a given iOS/tvOS/watchOS/MacCatalyst application bundle in a target device/simulator";
+
+        protected override AppleInstallCommandArguments AppleAppArguments { get; } = new();
+
+        protected override string CommandUsage { get; } = "apple install [OPTIONS]";
```
greenEkatherine, line 24: If I remember correctly we agreed to explicitly put here mandatory parameters. Could you please add them in some upcoming PR?

src/Microsoft.DotNet.XHarness.Apple/Orchestration/AppUninstallOrchestrator.cs
```diff
@@ -0,0 +1,95 @@
...
+                cancellationToken);
+        }
+
+        protected override Task CleanUpSimulators(IDevice device, IDevice? companionDevice)
+            => Task.CompletedTask; // no-op so that we don't remove the app after (reset will only clean it up before)
+
+        protected override Task<ExitCode> InstallApp(AppBundleInformation appBundleInfo, IDevice device, TestTargetOs target, CancellationToken cancellationToken)
+            => Task.FromResult(ExitCode.SUCCESS); // no-op for obvious reasons
+
+        private class FakeAppBundleInformationParser : IAppBundleInformationParser
```
greenEkatherine, line 66: Does it needed?

greenEkatherine:
Please add tests for these new commands

greenEkatherine:
@premun  I'm going to merge this pull request now to unblock @imhameed , please apply comments from it in separate pull request

greenEkatherine (approved):

premun:
@greenEkatherine there will be one more command needed for @imhameed though. But I will address the feedback there.

SamMonoRT:
Thank you for the quick turnaround

src/Microsoft.DotNet.XHarness.Apple/Orchestration/AppUninstallOrchestrator.cs
premun: For this scenario, we don't actually parse anything because we don't have an app but we still need the identifier to run the app so this is a bit of a hack to get there

Issue #552, opened 2021-04-19
Add unit tests for apple commands


premun:
**Context**
There was refactoring of code when we were introducing the new commands (`install`, `uninstall`, `just-run` and just-test`).
All of the logic was moved out of the CLI project into the Apple project so that it can be tested.

**Goal**
- [ ] Add unit tests around this new layer of logic (mostly for the `BaseOrchestrator` where most of the code is)
- [x] Add integration tests for the install/just-test/uninstall flows

PR #553, opened 2021-04-19
Add new apple commands `just-run` and `just-test`

premun:
Adds new commands that only spawn the application and skip installation/uninstallation (#493)

This is to unblock @imhameed. Tests will be added as part of #552.


src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleUninstallCommand.cs
```diff
@@ -21,7 +22,7 @@ internal class AppleUninstallCommand : AppleAppCommand<AppleUninstallCommandArgu
         private const string CommandHelp = "Uninstalls a given iOS/tvOS/watchOS/MacCatalyst application bundle from a target device/simulator";
 
         protected override AppleUninstallCommandArguments AppleAppArguments { get; } = new();
-        protected override string CommandUsage { get; } = "apple uninstall [OPTIONS]";
+        protected override string CommandUsage { get; } = "apple uninstall --app=... --output-directory=... --targets=... [OPTIONS] [-- [RUNTIME ARGUMENTS]]";
```
greenEkatherine, line 25: Do we really need `output-directory` for the `uninstall` command?

premun: There's various log files (device logs, mlaunch logs) that we capture and put in this directory (for Helix, this all is uploaded then with the results).

greenEkatherine (approved):

PR #566, opened 2021-04-27
centrally control csharp version

bruno-garcia:
Each project had its own property for the C# language.
It was a mix of C# 8 and 9. With a few projects not specifying the version at all (and hence taking the default for the SDK version).

This PR sets the C# version to 9.0 on the project's root `Directory.Build.props` and removes any other definition. This way the version can be controlled at a single location, and all projects are aligned.


premun:
Oh, cool! I never thought of this! Thanks!

premun (approved):
Oh, cool! I never thought of this! Thanks!

mandel-macaque:
@bruno-garcia thanks for the contribution!

PR #567, opened 2021-04-27
file exception has filename

bruno-garcia:
Adding the file name as a second argument to `FIleNotFoundException`.

I.e, before:

```
Unhandled exception. System.IO.FileNotFoundException: message
   at Program.Main()
```

After:
```
Unhandled exception. System.IO.FileNotFoundException: message
File name: 'file_name'
   at Program.Main()
```

It's helpful if the code handling the exception wants to deal with the file name. Otherwise it would nee to rely on the format of the message and parse it.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -29,7 +29,7 @@ namespace Microsoft.DotNet.XHarness.Android
         {
             { "architecture", "shell getprop ro.product.cpu.abi"},
             { "app", "shell pm list packages -3"}
-        }; 
+        };
 
 
         public AdbRunner(ILogger log, string adbExePath = "") : this(log, new AdbProcessManager(log), adbExePath) { }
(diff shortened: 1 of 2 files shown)


premun (approved):
Thanks! Appreciated!

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -471,7 +471,7 @@ public List<string> PullFiles(string devicePath, string localPath)
             var result = allDevicesAndTheirProperties
                 .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Contains(apkRequiredProperty))
                 .ToDictionary(kvp => kvp.Key, kvp => kvp.Value) as Dictionary<string, string>;
-
```
mandel-macaque, line 474: @premun can the .editoconfig avoid this empty space changes?

premun: I don't think `.editorconfig` has anything to do with trimming trailing whitespace. I personally don't mind these changes. It helps though when they happen on empty lines since you won't ever git blame those

PR #575, opened 2021-04-28
Run Apple apps by their bundle ID only

premun:
When we run the app that we just installed, we pass path to the bundle and it gets parsed again by mlaunch so that it knows the ID of the app it should launch. mlaunch also accepts passing just the bundle ID which just runs the app.

This change does two things:
- Runs apps by their ID only (`test`, `run`, `just-test`, `just-run` commands)
- Changes the `-a` argument for for just-run and just-test commands to only bundle ID

This makes things faster for regular runs as well as for `just-run` and `just-test` for which you can then run the installed app even if you don't already have the original app bundle.


premun:
@imhameed fyi, I had to make this change and it changes the `just-` commands a little and requiring you to pass thebundle ID instead of the package path in the `-app` parameter

mandel-macaque (approved):

greenEkatherine (approved):

PR #576, opened 2021-04-28
return TESTS_FAILED(1) when instrumentation exit code != expected one

### LINKED ISSUE #446: Add an option to return an error code if the tests failed

### LINKED ISSUE #548: Failed Instrumentation exit code is being coerced to 71 (General error)
Noted in logs like this:
https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-heads-main-65d2e387b3c9414e92/System.Xml.Xsl.XslCompiledTransformApi.Tests/console.8303b34d.log?sv=2019-07-07&se=2021-05-03T20%3A41%3A50Z&sr=c&sp=rl&sig=elT9SgYO477h1mNFm9Nc74Lzxjxyh3ybm6oPAtWIMxE%3D

Condensed version if the above log gets lost:
```
[13:54:08] info: Running instrumentation class net.dot.MonoRunner took 18.3996502 seconds
[13:54:08] dbug: Exit code: 0
                 Standard Output:
                 INSTRUMENTATION_RESULT: return-code=1
                 INSTRUMENTATION_RESULT: test-results-path=/storage/emulated/0/Android/data/net.dot.System.Xml.Xsl.XslCompiledTransformApi.Tests/files/Documents/testResults.xml
                 INSTRUMENTATION_CODE: 1
[13:54:11] info: Wrote current ADB log to C:\h\w\A92F094E\w\ABE60973\uploads\adb-logcat-net.dot.System.Xml.Xsl.XslCompiledTransformApi.Tests-net.dot.MonoRunner.log
[13:54:11] fail: Non-success instrumentation exit code: 1, expected: 0
[13:54:11] info: Attempting to remove apk 'net.dot.System.Xml.Xsl.XslCompiledTransformApi.Tests': 
[13:54:11] dbug: Executing command: 'C:\h\w\A92F094E\p\microsoft.dotnet.xharness.cli\1.0.0-prerelease.21209.2\runtimes\any\native\adb\windows\adb.exe -s 09071JEC210218 uninstall net.dot.System.Xml.Xsl.XslCompiledTransformApi.Tests'
[13:54:12] info: Successfully uninstalled net.dot.System.Xml.Xsl.XslCompiledTransformApi.Tests.
XHarness exit code: 71 (GENERAL_FAILURE)
 Tue 04/13/2021-13:54:12.45
 Tue 04/13/2021-13:54:12.84
2021-04-13 13:54:13,396: INFO: 1496: run(82): main: Main thread starting 10 workers
```

... this may have been intentional, but at least initially we intended for the instrumentation exit code (1 here because XUnit tests failed) to be the exit code, and not 71 which is a catch-all "things went bad" exit code.

greenEkatherine:
Resolved #446 
Resolved #548


src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidRunCommand.cs
```diff
@@ -205,6 +205,7 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
                 if (instrumentationExitCode != expectedExitCode)
                 {
                     logger.LogError($"Non-success instrumentation exit code: {instrumentationExitCode}, expected: {expectedExitCode}");
+                    return ExitCode.TESTS_FAILED;
```
MattGal, line 208: I am not an Android testing expert (I'd ping @akoeplinger  / @steveisok ?) but if this is like regular xunit, I don't think we want to coerce every value to mean "tests failed", just the "1" case from Xunit tests.  

What I'm curious of from the experts is that there could be crash scenarios, for instance, that return some kind of HResult.  Would it be OK just to return the instrumentationExitCode in this case?

premun: Another option would be to add an option to XHarness that would just forward the instrumentation exit code and then the person who owns the app can decide how it should behave?

premun, line 208: FWIW, when expected doesn't match actual exit code, we return `ExitCode.GENERAL_FAILURE` in WASM/iOS

greenEkatherine: > I don't think we want to coerce every value to mean "tests failed", just the "1" case from Xunit tests.

But before this change we returned `GENERAL_FAILURE` in that case, it doesn't matter one test failed or more.

greenEkatherine: how do you differ failed tests from `GENERAL_FAILURE` in that case? When expected doesn't match actual exit code isn't it a failed test?

premun: It's a bit different in iOS/WASM - for `test` command we don't try to detect exit code at all and only decide based on how the tests went. For `run` command, we don't expect tests inside and try to detect the exit code and compare it with expected.

So it's a little bit different situation and might not apply here. I only meant that as "let's behave similar in Android/iOS" but if that doesn't make sense, it's not a big deal.

greenEkatherine: @premun it's good idea, but in that case our exit code is the same as the one from instrumentation. We probably should use separate values for them, so the app owner can understand was is from instrumentation or not

MattGal: > > I don't think we want to coerce every value to mean "tests failed", just the "1" case from Xunit tests.
> 
> But before this change we returned `GENERAL_FAILURE` in that case, it doesn't matter one test failed or more.

I definitely don't think general failure describes the scenario of "everything worked as expected except one or more tests failed".  That's quite a specific failure.  To give context about the problem here, especially while standing up new machines but also randomly and when upgrading existing ones, we end up with machines where the device is missing, the device connection times out repeatedly, etc.  Having as many different exit codes for these states as possible, as well as not having a shared exit code that means this non-exceptional circumstance, helps find broken machines faster so DDFUN can offline them.

akoeplinger: So I think in the case of Android we _can_ detect if things went wrong horribly (crash etc) since then we don't get the `return-code` instrumentation key.

I think it's fine to return `TESTS_FAILED` if we did get a `return-code` but it didn't match what we were expecting, and return `GENERAL_FAILURE` in other cases.

MattGal (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidRunCommand.cs
MattGal: Works for me.  We need this change or similar soon because right now it's causing "test failure / test crash" to return the same as "device lost its auth" and making it nearly impossible to broadly catch when newly online hardware is setup wrong or when hardware loses its auth.

premun:
@greenEkatherine you will need to merge main since we changed the stages after this PR was already opened

PR #579, opened 2021-04-29
Stop targeting netstandard2.0 in TestRunner projects

premun:
This will help with some nullability issues in the project.


premun:
@mandel-macaque can we do this or I missed something?
@mattleibow I guess this won't be an issue for Xamarin?

PR #583, opened 2021-04-30
Redirect app output into a separate file for Apple device

premun:
There is a bug in mlaunch and the `--stdout` argument doesn't do anything but I will fix it later.
Once it's fixed, we will be ready.


premun:
Ah, actually, I will need to do this for `AppRunner` as well. And fix unit tests

greenEkatherine (approved):

Issue #584, opened 2021-04-30
Allow Android arm32 apps to run on Android arm64 devices


fanyang-mono:
Currently, if I try to run Android arm32 test apps on Android arm64 devices, I ran into the following error
```
[08:57:43] fail: No devices with architecture 'armeabi-v7a' was found among attached devices.
      [08:57:43] crit: Failed to run test package: Failed to find compatible device: armeabi-v7a
                       System.Exception: Failed to run test package: Failed to find compatible device: armeabi-v7a
                          at Microsoft.DotNet.XHarness.CLI.Commands.Android.AndroidInstallCommand.InvokeHelper(ILogger logger, String apkPackageName, String appPackagePath, String apkRequiredArchitecture, String deviceId, TimeSpan bootTimeoutSeconds, AdbRunner runner) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidInstallCommand.cs:line 120
                          at Microsoft.DotNet.XHarness.CLI.Commands.Android.AndroidTestCommand.InvokeInternal(ILogger logger) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs:line 77
```
However, theoretically, it should be allowed, according to android docs:
>  64-bit devices also support their 32-bit variants. Using arm64-v8a devices as an example, the device can also run armeabi and armeabi-v7a code

SamMonoRT:
This is blocking enabling test validation in CI on Arm32 apps.

greenEkatherine:
Looking into it now

greenEkatherine:
@fanyang-mono meanwhile as a workaround you may put explicit parameter `--device-arch="arm64-v8a"` to force running on these devices

greenEkatherine:
@fanyang-mono @SamMonoRT could you please provide test APK for this?

greenEkatherine:
This problem might require two changes. The first one - and I started with it - continue looking for compatible devices if we detect more then one architecture from APK. As a bonus of this change there will be possible to provide `--device-arch` more than once. I think it makes total sense because application may target more than one architecture.

Then I need APK to test if the first part would be enough. In other case the second part requires to check not only device architecture itself but its compatible architectures as well. In my opinion it should take effect only when `--device-arch` isn't set explicitly. Otherwise it might mean that user want exact that architecture and not some compatible one.

greenEkatherine:
[...]

PR #591, opened 2021-05-06
Look for devices with any of detected from apk architectures

greenEkatherine:
The first part of changes for #584. Before we choose only one detected architecture from .apk (if it's not provided explicitely). This PR will allow to continue looking for compatible devices if we detect more then one architecture from APK. Also I make it possible to provide `--device-arch` more than once.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -420,7 +420,7 @@ namespace Microsoft.DotNet.XHarness.Android
 
 
 
-        public string? GetDeviceToUse(ILogger logger, string apkRequiredProperty, string propertyName)
+        public string? GetDeviceToUse(ILogger logger, IEnumerable<string> apkRequiredProperty, string propertyName)
         {
             var allDevicesAndTheirProperties = GetAllDevicesToUse(logger, apkRequiredProperty, propertyName);
             if (allDevicesAndTheirProperties.Count > 0)
@@ -434,7 +434,7 @@ namespace Microsoft.DotNet.XHarness.Android
 
         public string? GetUniqueDeviceToUse(ILogger logger, string apkRequiredProperty, string propertyName)
         {
-            var devices = GetAllDevicesToUse(logger, apkRequiredProperty, propertyName);
+            var devices = GetAllDevicesToUse(logger, new[]{ apkRequiredProperty}, propertyName);
             if (devices.Count == 0)
             {
                 logger.LogError($"Cannot find a device with {propertyName}={apkRequiredProperty}, please check that a device is attached");
@@ -448,7 +448,7 @@ namespace Microsoft.DotNet.XHarness.Android
             return devices.Keys.First();
         }
 
-        public Dictionary<string, string> GetAllDevicesToUse(ILogger logger, string apkRequiredProperty, string propertyName)
+        public Dictionary<string, string> GetAllDevicesToUse(ILogger logger, IEnumerable<string> apkRequiredProperty, string propertyName)
         {
 
             var allDevicesAndTheirProperties = new Dictionary<string, string?>();
@@ -469,13 +469,13 @@ namespace Microsoft.DotNet.XHarness.Android
             }
 
             var result = allDevicesAndTheirProperties
-                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Contains(apkRequiredProperty))
+                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Intersect(apkRequiredProperty).Count() != 0)
                 .ToDictionary(kvp => kvp.Key, kvp => kvp.Value) as Dictionary<string, string>;
 
             if (result.Count == 0)
             {
                 // In this case, the enumeration worked, we found one or more devices, but nothing matched the APK's architecture; fail out.
-                logger.LogError($"No devices with {propertyName} '{apkRequiredProperty}' was found among attached devices.");
+                logger.LogError($"No devices with {propertyName} '" + string.Join(", ", apkRequiredProperty) + "' was found among attached devices.");
             }
 
             return result;
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -469,13 +469,15 @@ namespace Microsoft.DotNet.XHarness.Android
             }
 
             var result = allDevicesAndTheirProperties
-                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Intersect(apkRequiredProperty).Count() != 0)
+                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Intersect(apkRequiredProperty).Any())
+
                 .ToDictionary(kvp => kvp.Key, kvp => kvp.Value) as Dictionary<string, string>;
 
             if (result.Count == 0)
             {
                 // In this case, the enumeration worked, we found one or more devices, but nothing matched the APK's architecture; fail out.
-                logger.LogError($"No devices with {propertyName} '" + string.Join(", ", apkRequiredProperty) + "' was found among attached devices.");
+                logger.LogError($"No devices with {propertyName} '{ string.Join("', '", apkRequiredProperty) }' was found among attached devices.");
+
             }
 
             return result;
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/AndroidGetDeviceCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/AndroidGetDeviceCommandArguments.cs
@@ -44,6 +44,20 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Android
 
         public override void Validate()
         {
+
+            foreach (var archName in _deviceArchitecture ?? throw new ArgumentException("architecture cannot be empty"))
+            {
+                try
+                {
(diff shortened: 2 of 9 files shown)


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -469,13 +469,13 @@ public List<string> PullFiles(string devicePath, string localPath)
             }
 
             var result = allDevicesAndTheirProperties
-                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Contains(apkRequiredProperty))
+                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Intersect(apkRequiredProperty).Count() != 0)
```
premun, line 472:
```suggestion
                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Intersect(apkRequiredProperty).Any())

```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -469,13 +469,13 @@ public List<string> PullFiles(string devicePath, string localPath)
...
             var result = allDevicesAndTheirProperties
-                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Contains(apkRequiredProperty))
+                .Where(kvp => !string.IsNullOrEmpty(kvp.Value) && kvp.Value.Split().Intersect(apkRequiredProperty).Count() != 0)
                 .ToDictionary(kvp => kvp.Key, kvp => kvp.Value) as Dictionary<string, string>;
 
             if (result.Count == 0)
             {
                 // In this case, the enumeration worked, we found one or more devices, but nothing matched the APK's architecture; fail out.
-                logger.LogError($"No devices with {propertyName} '{apkRequiredProperty}' was found among attached devices.");
+                logger.LogError($"No devices with {propertyName} '" + string.Join(", ", apkRequiredProperty) + "' was found among attached devices.");
```
premun, line 478:
```suggestion
                logger.LogError($"No devices with {propertyName} '{ string.Join("', '", apkRequiredProperty) }' was found among attached devices.");

```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/AndroidGetDeviceCommandArguments.cs
```diff
@@ -25,15 +27,18 @@ public string AppPackagePath
...
             { "app|a=", "Path to already-packaged app",
                 v => AppPackagePath = RootPath(v)
             },
-            { "device-arch=", "If specified, forces running on a device with given architecture (x86, x86_64, arm64-v8a or armeabi-v7a). Otherwise inferred from supplied APK",
-                v => DeviceArchitecture = v
+            {
+                "device-arch=",
+                "If specified, forces running on a device with given architecture (x86, x86_64, arm64-v8a or armeabi-v7a). Otherwise inferred from supplied APK. " +
+                "Can be used more than once.",
+                v => _deviceArchitecture.Add(v)
```
premun, line 41: If we can only have these values, should we validate what we received to be one of them?

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/AndroidInstallCommandArguments.cs
```diff
@@ -43,6 +45,12 @@ protected override OptionSet GetTestCommandOptions() => new()
                 "device-id=", "Device where APK should be installed",
                 v => DeviceId = v
             },
+            {
+                "device-arch=",
+                "If specified, forces running on a device with given architecture (x86, x86_64, arm64-v8a or armeabi-v7a). Otherwise inferred from supplied APK. " +
+                "Can be used more than once.",
+                v => _deviceArchitecture.Add(v)
```
premun, line 52: Same comment about validation

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/AndroidTestCommandArguments.cs
```diff
@@ -67,8 +68,11 @@ public string PackageName
 
         protected override OptionSet GetTestCommandOptions() => new()
         {
-            { "device-arch=", "If specified, only run on a device with the listed architecture (x86, x86_64, arm64-v8a or armeabi-v7a).  Otherwise infer from supplied APK",
-                v => DeviceArchitecture = v
+            {
+                "device-arch=",
+                "If specified, forces running on a device with given architecture (x86, x86_64, arm64-v8a or armeabi-v7a). Otherwise inferred from supplied APK. " +
+                "Can be used more than once.",
+                v => _deviceArchitecture.Add(v)
```
premun, line 75: Same comment about validation

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetDeviceCommand.cs
```diff
@@ -46,15 +47,15 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
             }
 
             var runner = new AdbRunner(logger);
-            string apkRequiredArchitecture;
+            IEnumerable<string> apkRequiredArchitecture;
 
-            if (!string.IsNullOrEmpty(_arguments.DeviceArchitecture))
+            if (_arguments.DeviceArchitecture.Count() != 0)
```
premun, line 52: `Count()` would go through the whole enumeration which is more costly than just checking if there's at least 1 element

```suggestion
            if (_arguments.DeviceArchitecture.Any())
```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidInstallCommand.cs
```diff
@@ -44,20 +45,20 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
             var runner = new AdbRunner(logger);
 
-            string? apkRequiredArchitecture = null;
+            List<string> apkRequiredArchitecture = new();
 
             if (string.IsNullOrEmpty(_arguments.DeviceId))
             {
                 // trying to choose suitable device
-                if (!string.IsNullOrEmpty(_arguments.DeviceArchitecture))
+                if (_arguments.DeviceArchitecture.Count() != 0)
```
premun, line 53:
```suggestion
                if (_arguments.DeviceArchitecture.Any())
```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidInstallCommand.cs
```diff
@@ -83,7 +84,7 @@ public static ExitCode InvokeHelper(ILogger logger, string apkPackageName, strin
 
                     // if call via install command device id must be set
                     // otherwise - from test command - apkRequiredArchitecture was set by user or .apk architecture
-                    deviceId ??= apkRequiredArchitecture != null
+                    deviceId ??= apkRequiredArchitecture.Count() != 0
```
premun, line 87:
```suggestion
                    deviceId ??= apkRequiredArchitecture.Any()
```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -55,17 +56,17 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
             var runner = new AdbRunner(logger);
 
             // Assumption: APKs we test will only have one arch for now
-            string apkRequiredArchitecture;
+            IEnumerable<string> apkRequiredArchitecture;
 
-            if (!string.IsNullOrEmpty(_arguments.DeviceArchitecture))
+            if (_arguments.DeviceArchitecture.Count() != 0)
```
premun, line 61:
```suggestion
            if (_arguments.DeviceArchitecture.Any())
```

premun (changes requested):

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetDeviceCommand.cs
greenEkatherine: Interesting, in C++ it is const so I used to use it 😃

premun: So here `.Count()` is an extension method for all `Enumerable` types, no matter what they are. For many of them, it cannot be const (such as linked list) since you have to traverse the whole data structure to count the elements.

So I guess `Count()` in C++ is only defined for data structures that keep their count?

greenEkatherine: For linked list it's const as well https://en.cppreference.com/w/cpp/container/list/size There is singly-linked list and it's an exception of that rule

premun: Ah, this gives me flashbacks to university times when I was looking up complexities for std 😄 

Anyway, this `Count()` from `System.Linq` is a literally doing `foreach` + `++count;` and `return count;` so beware.

Also another thing to be aware of is that enumerations in C# constructed using Linq are sort of like "queries" and every time you want their values, they evaluate again (all `Where()`, all `Select()`). Sort of like SQL views. The only way to make them evaluate once is do `.ToArray()` or `.ToList()` or similar which actually turn them into the data structure.

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/AndroidGetDeviceCommandArguments.cs
greenEkatherine: validation added

premun (approved):

PR #593, opened 2021-05-10
[wasm][testing] hosting webSocket echo server in xharness process

pavelsavara:
In case that we are running unit tests in xharness, we would like to be able to host webserver in the xharness process rather than to call external servers in azure. 

That would allow us to move some unit tests from outer loop to inner loop. See https://github.com/dotnet/runtime/pull/52923

Part of overall epic https://github.com/dotnet/runtime/issues/42852

```
    <WasmXHarnessArgs>$(WasmXHarnessArgs) --set-web-server-http-env=DOTNET_TEST_WEBSOCKETHOST,DOTNET_TEST_HTTPHOST</WasmXHarnessArgs>
    <WasmXHarnessArgs>$(WasmXHarnessArgs) --set-web-server-https-env=DOTNET_TEST_SECUREWEBSOCKETHOST,DOTNET_TEST_SECUREHTTPHOST</WasmXHarnessArgs>
    <WasmXHarnessArgs>$(WasmXHarnessArgs) --web-server-middleware=$(ArtifactsDir)bin/MonoNetTestServer/$(NetCoreAppCurrent)-$(Configuration)/MonoNetTestServer.dll,GenericHandler</WasmXHarnessArgs>
    <WasmXHarnessArgs>$(WasmXHarnessArgs) --web-server-middleware=$(ArtifactsDir)bin/NetCoreServer/$(NetCoreAppCurrent)-$(Configuration)/NetCoreServer.dll,GenericHandler</WasmXHarnessArgs>


set XHARNESS_CLI_PATH=c:\Dev\xharness\artifacts\bin\Microsoft.DotNet.XHarness.CLI\Debug\net6.0\Microsoft.DotNet.XHarness.CLI.dll
set XHARNESS_COMMAND=test-browser
./dotnet build /t:Test src/libraries/System.Net.WebSockets.Client/tests /p:TargetOS=Browser /p:TargetArchitecture=wasm /p:Configuration=Release
```


pavelsavara:
@lewing

pavelsavara:
@akoeplinger could you please comment on topics in the description ? Thank you.

pavelsavara:
Right now, I'm thinking that the new echo endpoints should stay in runtime repository in form of assembly, which could be offered to xharness as a binary plugin. 
That would be better than current state, because the test would be in same repo as tested code. 
For example I would like to create unit test for https://github.com/dotnet/runtime/issues/51041, which will need server side. Ideally on single PR.

It would still leave us with .NET 4.7 version of the [CoreFxNetCloudService](https://github.com/dotnet/runtime/tree/main/src/libraries/Common/tests/System/Net/Prerequisites/Servers/CoreFxNetCloudService) and another .net core version for xharness. Unless we convert whole CoreFxNetCloudService to .net core.

@lewing your thoughts ?

radical:
Could you please explain what this is doing, and why, in the PR description?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -67,33 +69,51 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
+            var cts = new CancellationTokenSource();
+            try
             {
-                // v8 needs this flag to enable WASM support
-                engineArgs.Add("--expose_wasm");
-            }
+                var serverURLs = await WebServer.Start(
+                    _arguments, logger,
+                    null,
+                    cts.Token);
```
radical, line 78: Do we need to always start the web server for `wasm test` case? Can we limit it to only when `WebServerMiddlewarePaths` is set?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -102,11 +122,14 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
                     stdoutLog: new CallbackLog(logProcessor.Invoke),
                     stderrLog: new CallbackLog(m => logger.LogError(m)),
                     _arguments.Timeout);
+                if (cts.IsCancellationRequested)
+                {
+                    return ExitCode.TIMED_OUT;
+                }
```
radical, line 128: nit: blank line before, and after the block.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -175,6 +175,7 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
             {
                 // added based on https://github.com/puppeteer/puppeteer/blob/main/src/node/Launcher.ts#L159-L181
                 "--enable-features=NetworkService,NetworkServiceInProcess",
+                "--allow-insecure-localhost",
```
radical, line 178: Why is this needed?

radical (changes requested):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
pavelsavara: Because some of the tested connections use HTTPS, but we do not install any trusted certificates. 
Since we do not own TLS layer in the Browser platform, we don't have to test it.
Just keeping it simple.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
```diff
@@ -232,14 +227,29 @@ private void RunSeleniumLogMessagePump(IWebDriver driver, CancellationToken toke
...
+
+                if (sb.Length > 0)
+                    sb.Append('&');
+                // see runtime\src\libraries\Common\tests\System\Net\Configuration.WebSockets.cs
+                // see runtime\src\libraries\Common\tests\System\Net\Configuration.Http.cs
+                sb.Append($"arg={HttpUtility.UrlEncode($"--setenv=DOTNET_TEST_WEBSOCKETHOST={hostAndPort}")}");
+                sb.Append($"&arg={HttpUtility.UrlEncode($"--setenv=DOTNET_TEST_SECUREWEBSOCKETHOST={hostAndPortSecure}")}");
+                sb.Append($"&arg={HttpUtility.UrlEncode($"--setenv=DOTNET_TEST_HTTPHOST={hostAndPort}")}");
+                sb.Append($"&arg={HttpUtility.UrlEncode($"--setenv=DOTNET_TEST_SECUREHTTPHOST={hostAndPortSecure}")}");
+            }
```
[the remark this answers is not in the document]
pavelsavara: Moving the list of variables to set into runtime repo makes good sense, thanks! 
I done it as two list (of http and https) variable names instead, because I don't want the web server middleware to have this integration responsibility.

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
```diff
@@ -0,0 +1,152 @@
...
+            .FirstOrDefault();
+
+        if (ipAddress == null || ipAddressSecure == null)
+        {
+            throw new InvalidOperationException("Failed to determine web server's IP address or port");
+        }
+        return new ServerURLs(ipAddress, ipAddressSecure);
+    }
+
+    internal class TestWebServerStartup
```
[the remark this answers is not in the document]
pavelsavara: `TestWebServerStartup` is nested class. Not really used outside of this file.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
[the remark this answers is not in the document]
pavelsavara: Also, the middleware doesn't know the address and port. And the URLencode is also complicating this.

pavelsavara:
great feedback @radical, thanks!

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmBrowserTestRunner.cs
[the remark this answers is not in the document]
radical: > Also, the middleware doesn't know the address and port. And the URLencode is also complicating this.

That's why I was suggesting having a method on that middleware type, to which you can pass the urls, and get back the `--setenv=DOTNET_TEST_WEBSOCKETHOST=http://...` string, which you can then `URLEncode` here.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -67,33 +68,55 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
+                {
+                    serverURLs = await WebServer.Start(
+                        _arguments, logger,
+                        null,
+                        cts.Token);
+                }
 
-            engineArgs.AddRange(_arguments.EngineArgs);
-            engineArgs.Add(_arguments.JSFile);
+                cts.CancelAfter(_arguments.Timeout);
```
radical, line 83: This is only used if the web server is used, IIUC.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -67,33 +68,55 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
+
+                foreach (var envVariable in _arguments.SetWebServerEnvironmentVariablesHttp)
+                {
+                    engineArgs.Add($"--setenv={envVariable}={serverURLs!.Http}");
+                }
+
+                foreach (var envVariable in _arguments.SetWebServerEnvironmentVariablesHttps)
+                {
+                    engineArgs.Add($"--setenv={envVariable}={serverURLs!.Https}");
+                }
```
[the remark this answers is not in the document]
pavelsavara: This is intentional. I start the "empty" server even when there is just `SetWebServerEnvironmentVariablesHttp` or `SetWebServerEnvironmentVariablesHttps`. That would still open connections and respond with 404.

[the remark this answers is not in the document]
radical: Hm ok, I see `if (_arguments.WebServerMiddlewarePathsAndTypes.Count > 0 || _arguments.SetWebServerEnvironmentVariablesHttp.Count > 0 || _arguments.SetWebServerEnvironmentVariablesHttps.Count > 0)`

.. but this seems side-effecty. If you want to have a web server running anyway, then I would suggest an explicit argument for that, `--start-web-server` (or `--with-web-server`).

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
```diff
@@ -0,0 +1,159 @@
...
+    {
+        public ServerURLs(string http, string https)
+        {
+            this.Http = http;
+            this.Https = https;
+        }
+
+        public string Http { get; init; }
+        public string Https { get; init; }
+    }
```
radical, line 158: This can become:
```csharp
    public record ServerURLs(string Http, string Https);
```

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -67,33 +68,54 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
-            if (_arguments.Engine == JavaScriptEngine.V8 || _arguments.Engine == JavaScriptEngine.JavaScriptCore)
-            {
-                // v8/jsc want arguments to the script separated by "--", others don't
-                engineArgs.Add("--");
-            }
+                if (_arguments.Engine == JavaScriptEngine.V8)
+                {
+                    // v8 needs this flag to enable WASM support
+                    engineArgs.Add("--expose_wasm");
+                }
```
radical, line 90: Why is this needed here? This should be added to the arguments in the runtime repo

radical: nvm, ignore this!

pavelsavara: it's not part of this change

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
```diff
@@ -0,0 +1,149 @@
...
+                .Addresses
+                .Where(a => a.StartsWith("https:"))
+                .FirstOrDefault();
+
+            if (ipAddress == null || ipAddressSecure == null)
+            {
+                throw new InvalidOperationException("Failed to determine web server's IP address or port");
+            }
+
+            return new ServerURLs(ipAddress.Substring(ipAddress.LastIndexOf('/') + 1),
```
radical, line 80: What is `ipAddress.Substring(ipAddress.LastIndexOf('/') + 1` for?

radical: You can use `new Uri(ipAddress)`, and then use `$"{uri.Host}:{uri.Port}"`

pavelsavara: done

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -102,11 +124,16 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
                     stdoutLog: new CallbackLog(logProcessor.Invoke),
                     stderrLog: new CallbackLog(m => logger.LogError(m)),
                     _arguments.Timeout);
+                
+                if (cts.IsCancellationRequested)
+                {
+                    return ExitCode.TIMED_OUT;
+                }
```
radical, line 131: Um will this ever happen? IIUC, `cts` is being used only for the web server, and we are not canceling it before this step.

radical: nit: rename `cts` to `webServerCts`, so it's clear what it is used for.

pavelsavara: renamed. 
we do `webServerCts.CancelAfter(_arguments.Timeout)` which could kill the server if the server initialization took too long.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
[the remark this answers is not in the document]
pavelsavara: ok, I stayed with your original proposal of `WebServerMiddlewarePathsAndTypes` only for both blocks.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
radical: Having the `.CancelAfter` makes sense. But by this point in the code we have finished running the jsengine, which ran with the same timeout. So, we should continue to use the `result` from that. Or if you want to check for a timeout, then you can use `result.TimedOut`.

radical (approved):
The environment var expansion stuff could maybe be done outside the wasm commands, but that can be done later when we can see how they get used for non-wasm cases.

LGTM! Thank you for the PR, and patience with the feedback 👍

akoeplinger (commented):
Missed the ping on this one, looks good to me thank you!

PR #600, opened 2021-05-14
Run arm32 apps on arm64 devices

### LINKED ISSUE #584: Allow Android arm32 apps to run on Android arm64 devices

greenEkatherine:
Resolved #584 
Before we check only architecture of a device by running `adb shell getprop ro.product.cpu.abi`
But devices can be compatible with more than 1 architecture: https://developer.android.com/ndk/guides/abis#android-platform-abi-support
 Using arm64-v8a devices as an example, the device can also run armeabi and armeabi-v7a code.
So, to get list of compatible architectures I replaced previous command with `adb shell getprop ro.product.cpu.abilist`


premun (approved):
Can you put short description in the PR why this works?

PR #607, opened 2021-05-24
Improve TcpTextWriter

premun:
Dumping some improvements I made in a working branch
- Improves the life cycle of the object so that it cannot be in an unitialized state and can be disposed better
- Added docs
- Improved exception messages


premun:
@mandel-macaque I hope this doesn't clash with what you're working on?

PR #608, opened 2021-05-24
[Xamarin] Allow to use the runners in xamarin apps by adding netstandard 2.1 as a target.

mandel-macaque:


src/Microsoft.DotNet.XHarness.TestRunners.Common/TcpTextWriter.cs
```diff
@@ -194,8 +194,9 @@ private static string SelectHostName(string[] names, int port)
 
             return result;
         }
-
+#if NET6_0
```
mandel-macaque, line 197: @premun  is this correct?

premun: @mandel-macaque please have a look at https://github.com/dotnet/xharness/pull/607 and let's merge that first. It will remove the need for all of these changes you're making here

premun (approved):

Issue #611, opened 2021-05-27
Catalyst: open -W after I/O inside .app Contents is flaky


lambdageek:
It seems like doing I/O inside the `Contents/` of a catalyst .app will sometimes make `open -W that.app` fail
with:

```
The application cannot be opened for an unexpected reason, error=Error Domain=NSOSStatusErrorDomain Code=-10827 "kLSNoExecutableErr: The executable is missing" UserInfo={_LSLine=3691, _LSFunction=_LSOpenStuffCallLocal}
```

Investigated here: https://github.com/dotnet/runtime/pull/53197#issuecomment-849800410

in my case I was doing `cp -R`, but dotnet xharness (or dotnet helix, I'm not sure) is doing `chmod +x`.  In either case, I think sticking in a `sync` before the `open -W` is a good idea.

lambdageek:
Hmm... doing some testing by modifying xharness.

Seems like `sync` is not enough.

This [stackoverflow answer](https://stackoverflow.com/a/66550760) seems to suggest that we should run
`lsregister -f`

```
/System/Library/Frameworks/CoreServices.framework/Frameworks/LaunchServices.framework/Support/lsregister -f /path/to/app.app
```

Going to give that a try

PR #613, opened 2021-05-28
Run lsregister -f after I/O before running a MacCatalyst app

lambdageek:
Without this, launching the app fails (observed on arm64 locally, but
also on x64 on dotnet/runtime CI) with

```
The application cannot be opened for an unexpected reason, error=Error Domain=NSOSStatusErrorDomain Code=-10827 "kLSNoExecutableErr: The executable is missing" UserInfo={_LSLine=3691, _LSFunction=_LSOpenStuffCallLocal}
```

Fixes https://github.com/dotnet/xharness/issues/611


lambdageek:
The `lsregister` suggestion comes from [this stackoverflow answer](https://stackoverflow.com/a/66550760)

steveisok (approved):

PR #618, opened 2021-05-31
Stop accepting multiple test targets

premun:
This is a remnant from the original XHarness and we cannot really run more targets at the moment. Android commands don't support this neither so it doesn't make sense as it is


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Apple/AppleTestCommandArguments.cs
```diff
@@ -107,13 +107,9 @@ public override void Validate()
                 throw new ArgumentException($"Failed to find the app bundle at {AppPackagePath}");
             }
 
-            if (RunTargets.Count == 0)
+            if (Target == null)
```
greenEkatherine, line 110: Didn't you accidentally hide by overriding `Target` for `AppleTestCommandArguments`?

premun: No more `Target` anywhere anymore :)

greenEkatherine (approved):

PR #624, opened 2021-06-03
Allow CORS response expose headers

pavelsavara:
Allow response headers to be visible in wasm/JS. 
To fix wasm tests https://github.com/dotnet/runtime/issues/53668

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
@@ -43,7 +43,8 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands
                         {
                             builder.AllowAnyOrigin()
                                 .AllowAnyMethod()
-                                .AllowAnyHeader();
+                                .AllowAnyHeader()
+                                .WithExposedHeaders("*");
                         }));
                     services.AddRouting();
                     services.AddSingleton<ILogger>(logger);


src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
```diff
@@ -43,7 +43,8 @@ internal static async Task<ServerURLs> Start(TestCommandArguments arguments, ILo
                         {
                             builder.AllowAnyOrigin()
                                 .AllowAnyMethod()
-                                .AllowAnyHeader();
+                                .AllowAnyHeader()
+                                .WithExposedHeaders("*");
```
mdh1418, line 47: Would this negatively impact the requests with credentials (https)? 
I'm not familiar with what ConfigureServices is intended to do and what its used for in this WebServer context.

pavelsavara: Answered here https://github.com/dotnet/runtime/issues/53668#issuecomment-853950940

mdh1418 (approved):
Looks good to me!

pavelsavara:
`Could not find any applicable devices with device class`
https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-xharness-refs-pull-624-merge-c39cebff478e401384/System.Buffers.Tests/console.52b3b14c.log?sv=2019-07-07&se=2021-06-23T14%3A58%3A20Z&sr=c&sp=rl&sig=OC9yHbRlEsAIS3GwuGkXpSRJT7uQ00M81GxQWo%2Fj4EI%3D

PR #625, opened 2021-06-04
[wasm] Add scanning test output for error patterns

radical:
In some cases, the runtime might crash, for example, due to an
assertion. But the app continues to execute, causing v8/chrome to exit
with the expected exit code. This causes xharness to treat the tests as
passing, even though the runtime had crashed.

To catch such cases, this adds support for scanning the output for some
error patterns, and adds a new argument for `wasm test`, and `wasm
test-browser`:

    `--error-patterns=<file>`

This expects a text file with one pattern per line. The first character
of the line is treated special:

    '@': the rest of the line is a substring to be used as the pattern
         (case insensitive)
    '%': the rest of the line is a .net regex string
    '#': ignore the line completely

Example patterns file:
```
%console.error: Unhandled *[Ee]xception

@== JS stack trace
```

- this ignores whitespace only lines.
- And stops scanning after the first match

Example output:

```
fail: console.error: Unhandled exception in processFetchResponseBuffer Error: FS error

<snip />

info: === TEST EXECUTION SUMMARY ===
info: Total: 119, Errors: 0, Failed: 0, Skipped: 1, Time: 0.343874s

info: Process v8 exited with 0

fail: Application exited with the expected exit code: 0. But found a line matching an error pattern: console.error: Unhandled exception in processFetchResponseBuffer Error: FS error

XHarness exit code: 80 (APP_CRASH)
XHarness artifacts: ./xharness-output
```

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestBrowserCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestBrowserCommandArguments.cs
@@ -40,6 +40,7 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm
     {
         public Browser Browser { get; set; } = Browser.Chrome;
         public string? BrowserLocation { get; set; } = null;
+        public string? ErrorPatternsFile { get; set; }
 
         public List<string> BrowserArgs { get; set; } = new List<string>();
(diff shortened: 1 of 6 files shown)


radekdoulik (approved):

radical:
I'm going to merge this, so we can try to identify some recent wasm+aot test regressions. And I can address any feedback in a follow up PR.

Issue #626, opened 2021-06-06
WebServer used for wasm, should use `https` only when needed


radical:
`https` support was added for hosting an echo server, to help some networking tests in `runtime`. For `wasm test-browser`, we use a web server to host the app itself too, which is used for library tests, wasm build tests, debugger tests etc. And those don't need `https`.

But `https` is always being used for `wasm test-browser` now, which then requires dev certs to be installed. This should be used only when needed.

PR #627, opened 2021-06-07
make HTTPS and CORS optional features

pavelsavara:
- adds web-server-use-https
- adds web-server-use-cors
Fixes https://github.com/dotnet/xharness/issues/626

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/TestCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/TestCommandArguments.cs
@@ -23,6 +23,8 @@ namespace Microsoft.DotNet.XHarness.CLI.CommandArguments
         /// Tests classes to be included in the run while all others are ignored.
         /// </summary>
         public IEnumerable<string> ClassMethodFilters => _classMethodFilters;
+        public bool WebServerUseHttps { get; set; }
+        public bool WebServerUseCors { get; set; }
         public IList<(string path, string type)> WebServerMiddlewarePathsAndTypes { get; set; } = new List<(string, string)>();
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/TestCommandArguments.cs
```diff
@@ -45,6 +47,12 @@ protected override OptionSet GetCommandOptions()
                     "ignored. Can be used more than once.",
                     v => _classMethodFilters.Add(v)
                 },
+                { "web-server-use-https", "Bind HTTPS port too",
```
premun, line 50: Can you please move the WASM only related args to a WASM command arguments?

`TestCommandArguments` is used across iOS and Android at the moment but the switches don't do anything there.

Applies for even the previously added ones

pavelsavara: We are planning to bring the same functionality to Android soon with @mdh1418

premun: Ah, ok. Then leave it as-is. We will soon be refactoring a bit how we define arguments so taht it's easy to assemble the sets (#431)

mdh1418 (approved):
Looks good to me 👍

radical (approved):

pavelsavara:
> What are the situations in which we need to use cors and when would we not need cors?
Android or iOS tests are probably not triggering CORS and therefore it's not necessary to enable it.
CORS is necessary for Browser tests because the http client sends CORS headers.

PR #629, opened 2021-06-08
Reboot device in case of installation failure

### LINKED ISSUE #619: RCA: all Pixel 4a devices were found in recovery mode
Since May 20th and following days all new Pixel 4a devices (DNCENGWIN046 and above) lost connection and were found in recovery mode https://github.com/dotnet/core-eng/issues/13183
Now they are recovered and back online but we need to find a root cause and prevent it in the future.

greenEkatherine:
Resolves #619

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -234,13 +234,16 @@ namespace Microsoft.DotNet.XHarness.Android
                 result = RunAdbCommand($"install \"{apkPath}\"");
             }
 
-            // 3. Installation timed out; restarting the ADB server helps
+            // 3. Installation timed out; restarting the ADB server, reboot the device and give more time for installation
+            // installer might hang up so we need to clean it up and free memory
             if (result.ExitCode == (int)AdbExitCodes.INSTRUMENTATION_TIMEOUT)
             {
-                _log.LogWarning($"Installation timed out; Will make one attempt to restart ADB server, then retry the install");
+                _log.LogWarning($"Installation timed out; Will make one attempt to restart ADB server and the device, then retry the install");
                 KillAdbServer();
                 StartAdbServer();
-                result = RunAdbCommand($"install \"{apkPath}\"");
+                RebootAndroidDevice();
+                WaitForDevice();
+                result = RunAdbCommand($"install \"{apkPath}\"", TimeSpan.FromMinutes(10));
             }
 
             if (result.ExitCode != 0)
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -21,6 +21,7 @@ namespace Microsoft.DotNet.XHarness.Android
         private const string AdbEnvironmentVariableName = "ADB_EXE_PATH";
         private const string AdbDeviceFullInstallFailureMessage = "INSTALL_FAILED_INSUFFICIENT_STORAGE";
         private const string AdbInstallBrokenPipeError = "Failure calling service package: Broken pipe";
+        private const string AdbInstallException = "Exception occurred while executing 'install':";
         private const string AdbShellPropertyForBootCompletion = "sys.boot_completed";
         private readonly string _absoluteAdbExePath;
         private readonly ILogger _log;
@@ -213,6 +214,18 @@ namespace Microsoft.DotNet.XHarness.Android
             }
 
             var result = RunAdbCommand($"install \"{apkPath}\"");
+            result.ExitCode = 1;
+            result.StandardError = @"adb: failed to install C:\h\w\B67409AF\w\A8C9091E\e\System.Xml.Linq.Misc.Tests.apk: 
+                 Exception occurred while executing 'install':
+                 java.lang.NullPointerException: Attempt to invoke virtual method 'void android.content.pm.PackageManagerInternal.freeStorage(java.lang.String, long, int)' on a null object reference
+
+                     at com.android.server.StorageManagerService.allocateBytes(StorageManagerService.java:3901)
+
+                     at android.os.storage.StorageManager.allocateBytes(StorageManager.java:2278)
+
+                     at android.os.storage.StorageManager.allocateBytes(StorageManager.java:2335)
+
+                     at com.android.server.pm.PackageI";
 
             // Two possible retry scenarios, theoretically both can happen on the same run:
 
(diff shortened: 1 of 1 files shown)


akoeplinger (approved):

akoeplinger:
Hmm actually, does this really solve the issue? I assume if the device is in recovery mode you can't restart it with adb?

If you only tested on DNCENGWIN-022 then this might not work since that machine actually has an older Pixel 3 device attached, although it has Android 11.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -234,13 +234,16 @@ public int InstallApk(string apkPath)
                 result = RunAdbCommand($"install \"{apkPath}\"");
```
akoeplinger, line 234: I think we should also detect the `Exception occurred while executing 'install'` string here, besides the `AdbDeviceFullInstallFailureMessage`

greenEkatherine:
> Hmm actually, does this really solve the issue? I assume if the device is in recovery mode you can't restart it with adb?

We cannot return the device from recovery mode without manual intervention. The main goal of this PR - prevent this issue by cleaning up processes in time.
 
> If you only tested on DNCENGWIN-022 then this might not work since that machine actually has an older Pixel 3 device attached, although it has Android 11.

Agree, I'm not 100% sure that this solves the issue but I need to do some assumptions because I'm not able to reproduce it for test purposes: I tried 1000 installations and short timeout, any other ideas are welcome. My main assumption based on the fact that installations failed a lot not only on Pixel 4a devices but on Pixel 3 with Android 11 as well. I suggested that installations failed due to timeout more frequently on Android 11 and probably it is worth to raise an issue for Android.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
greenEkatherine: Great point, I will add

premun:
I am a bit worried that if we being to pile up retries like this, we will end up with having to have to raise the over all timeout for all of the tests which might have bad impact on the Helix queue.

@akoeplinger @greenEkatherine do you think it's better to just cancel the run, catch the error in Helix SDK and do the remediation there and re-run the work item somewhere else?

premun (approved):

akoeplinger:
> Agree, I'm not 100% sure that this solves the issue but I need to do some assumptions because I'm not able to reproduce it for test purposes: I tried 1000 installations and short timeout, any other ideas are welcome. My main assumption based on the fact that installations failed a lot not only on Pixel 4a devices but on Pixel 3 with Android 11 as well. I suggested that installations failed due to timeout more frequently on Android 11 and probably it is worth to raise an issue for Android.

This sounds like symptom vs. cause, I think what happened is that the Pixel 4a devices ran into an issue so the package installation started failing and after some time the system got broken (like the huge memory consumption you saw in the other issue), rebooted and entered recovery mode.

At that point the package installation started to time out, because the device was in recovery mode.

@greenEkatherine would you mind checking if the Pixel 4a devices ran into the `Exception occurred while executing 'install'` / `NullPointerException` problem before they started timing out due to being in recovery mode? If they did then we might get away with just rebooting the device in that case.

greenEkatherine:
> At that point the package installation started to time out, because the device was in recovery mode.
> 
> @greenEkatherine would you mind checking if the Pixel 4a devices ran into the `Exception occurred while executing 'install'` / `NullPointerException` problem before they started timing out due to being in recovery mode? If they did then we might get away with just rebooting the device in that case.

Sorry, I should have clarify what I meant by "timeout" - I was talking about job failed with 78 exit code:
```[04:43:37] info: Attempting to install C:\h\w\B17B098E\w\A66F08EF\e\System.ComponentModel.Annotations.Tests.apk: 
[04:43:37] dbug: Executing command: 'C:\h\w\B17B098E\p\microsoft.dotnet.xharness.cli\1.0.0-prerelease.21271.1\runtimes\any\native\adb\windows\adb.exe -s 98FAY14S83 install "C:\h\w\B17B098E\w\A66F08EF\e\System.ComponentModel.Annotations.Tests.apk"'
[04:48:37] fail: Waiting for command timed out: execution may be compromised.
[04:48:37] fail: Error:
                 Exit code: -2
                 Standard Output:
```

The order of jobs degradation:
1. From time to time `install` timed out with 78 exit code, there are still some successful runs
2. In short period of time there are exceptions thrown during installation: 
```[14:49:42] fail: Error:
                 Exit code: 1
                 Standard Output:
                 Performing Streamed Install
                 
                 Standard Error:
                 adb: failed to install C:\h\w\B67409AF\w\C1BE0A3D\e\System.CodeDom.Tests.apk: 
                 Exception occurred while executing 'install':
                 java.lang.NullPointerException: Attempt to invoke virtual method 'void android.content.pm.PackageManagerInternal.freeStorage(java.lang.String, long, int)' on a null object reference
                       at com.android.server.StorageManagerService.allocateBytes(StorageManagerService.java:3901)
                       at android.os.storage.StorageManager.allocateBytes(StorageManager.java:2278)
                       at android.os.storage.StorageManager.allocateBytes(StorageManager.java:2335)
                       at com.android.server.pm.PackageI
```
and warnings about slow operations in log which might also confirm issue with RAM:
```
05-21 15:00:11.018  1634  1869 W Looper  : Slow dispatch took 6208ms android.ui h=com.android.server.am.ActivityManagerService$UiHandler c=null m=31
05-21 15:00:11.020  1634  4168 W ActivityManager: Slow operation: 54ms so far, now at startProcess: done updating battery stats
05-21 15:00:11.020  1634  4168 W ActivityManager: Slow operation: 54ms so far, now at startProcess: building log message
05-21 15:00:11.020  1634  4168 I ActivityManager: Start proc 17003:net.dot.System.Runtime.Loader.Tests/u0a1723 for added application net.dot.System.Runtime.Loader.Tests
05-21 15:00:11.020  1634  4168 W ActivityManager: Slow operation: 54ms so far, now at startProcess: starting to update pids map
05-21 15:00:11.020  1634  4168 W ActivityManager: Slow operation: 54ms so far, now at startProcess: done updating pids map
05-21 15:00:12.062  1634  4168 W ActivityManager: Slow operation: 1043ms so far, now at attachApplicationLocked: after mServices.attachApplicationLocked
05-21 15:00:12.063  1634  4168 W ActivityManager: Slow operation: 1043ms so far, now at attachApplicationLocked: after updateOomAdjLocked
```
3. Device switch off to recovery mode, connection is lost, runs exit with DEVICE_NOT_FOUND(85) exit code.

greenEkatherine:
I added exception handling

@premun I like your idea but I would do it separately, after the fix is merged and new devices get back online

PR #630, opened 2021-06-09
Refactor CLI argument definition for better usage

### LINKED ISSUE #431: Refactor arguments classes help info and common parts

premun:
Currently, it was very hard to maintain argument definitions because inheritance no longer followed the real world e.g. some commands inheriting but hiding parent members or some commands inheriting commands by mistake such as the web server ones going to almost all commands.

This new way allows easier management and re-use of argument definitions across different commands.
Arguments are defined, parsed and validated in one place and commands then pick those.

Resolves #431


src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
```diff
@@ -3,31 +3,29 @@
...
 using Microsoft.Extensions.Options;
-using Microsoft.DotNet.XHarness.CLI.CommandArguments;
+using LogLevel = Microsoft.Extensions.Logging.LogLevel;
 
 namespace Microsoft.DotNet.XHarness.CLI.Commands
 {
     public class WebServer
     {
-        internal static async Task<ServerURLs> Start(TestCommandArguments arguments, ILogger logger, Func<WebSocket, Task>? onConsoleConnected, CancellationToken token)
+        internal static async Task<ServerURLs> Start(IWebServerArguments arguments, ILogger logger, Func<WebSocket, Task>? onConsoleConnected, CancellationToken token)
```
premun, line 28: @pavelsavara this refactoring uncovered some problems in the code that were not visible before. One of them is that the `wasm test` command doesn't require the `--app` argument but the `wasm test-browser` does. Example logs from our CI:
```
XHarness command issued: wasm test --engine=V8 --js-file=runtime.js -v --output-directory=/datadisks/disk1/work/C00109F8/w/BB980A05/uploads -- --enable-gc --run WasmTestRunner.dll System.Reflection.Extensions.Tests.dll -notrait category=IgnoreForCI -notrait category=OuterLoop -notrait category=failing
[11:57:57] fail: Required argument app|a= was not supplied
```
https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-xharness-refs-pull-630-merge-da7e4aeb8cb441b7ad/System.Reflection.Extensions.Tests/console.5ea40d4b.log?sv=2019-07-07&se=2021-06-29T11%3A55%3A57Z&sr=c&sp=rl&sig=kekuKVADunM6uhxCl8YXovs0qeY9Jc4i3t2Nfxj0hmk%3D

So, before, if we tried to spawn a web server with `wasm test` here, it would blow up on line 36:
```csharp
            var host = new WebHostBuilder()
                .UseKestrel()
                .UseContentRoot(arguments.AppPackagePath)     <<<<<<<<<<<<<<<<<<<<<<<
                .UseStartup<TestWebServerStartup>()
```

Now we have more strict validation and it uncovered this.

Can you please show me the usage of these commands in the runtime repo so that I can check we're really not using `--app` in one of them?

How should we handle this?

pavelsavara: I can see multiple options:
a) make `.UseContentRoot` only part of the `wasm test-browser` path. 
b) roll back the change which added the server to `wasm test`. Downside is that 'normal' scenario tests in inner loop could rely on this.
c) fid the proper directory which need to be passed to wasm test via `--app` and do it too.

In any case we should make sure that the script doesn't die silently. I'm confused how I missed it.

premun: I will do `a)` as part of this PR

premun: https://github.com/dotnet/xharness/pull/630/commits/94bdf98cb712cf6a3dbc4947959a5430701a9df2

radical: `--app` is set [here](https://github.com/dotnet/runtime/blob/main/eng/testing/tests.wasm.targets#L21-L22). It isn't useful for `wasm test`. IIRC, it was kept around because it was difficult to separate the changes from `wasm test-browser` in `runtime`.

We probably need do some clean up on the runtime side to be able to stop passing the arg. So, (a) makes sense.

premun: Thanks for the input! I can bump the XHarness in a separate PR and watch the build

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/WebServerMiddlewarePathsAndTypes.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+        }
+
+        public override void Action(string argumentValue)
+        {
+            var split = argumentValue.Split(',');
+            var file = split[0];
+            var type = split.Length > 1 && !string.IsNullOrWhiteSpace(split[1]) ? split[1] : "GenericHandler";
+
+            if (string.IsNullOrWhiteSpace(file))
```
pavelsavara, line 25: This probably belongs to `Validate()` now

premun: Yes! Fixed

src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleUninstallCommand.cs
```diff
@@ -28,12 +28,12 @@ public AppleUninstallCommand(IServiceCollection services) : base("uninstall", fa
...
-                    AppleAppArguments.Target,
-                    AppleAppArguments.DeviceName,
-                    AppleAppArguments.Timeout,
-                    AppleAppArguments.ResetSimulator,
-                    AppleAppArguments.EnableLldb,
+                    Arguments.BundleIdentifier,
+                    Arguments.Target,
+                    Arguments.DeviceName,
+                    Arguments.Timeout,
+                    false,
```
pavelsavara, line 35: just curious, why the change to constant ?

premun: Before, we were overriding the argument value to `false` because we couldn't hide the member completely thanks to inheritance

pavelsavara:
Overall LGTM, it would be good to test it on `runtime` `runtime-staging` and maybe `runtime-libraries-mono outerloop`

premun:
> Overall LGTM, it would be good to test it on `runtime` `runtime-staging` and maybe `runtime-libraries-mono outerloop`

Yeah, it would be nice if we could eventually copy one of the browser tests over to our PR pipeline to have more coverage.

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/Arguments/LaunchTimeoutArgument.cs
```diff
@@ -0,0 +1,20 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Microsoft.DotNet.XHarness.Common.CLI.CommandArguments;
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Android
+{
+    /// <summary>
+    /// Time to wait for boot completion. Defaults to 5 minutes.
```
greenEkatherine, line 11: I think it's not correct to mention here default value since it's not set up inside this class

src/Microsoft.DotNet.XHarness.CLI/Commands/GetStateCommand.cs
```diff
@@ -2,11 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.DotNet.XHarness.Common.CLI.CommandArguments;
 using Microsoft.DotNet.XHarness.Common.CLI.Commands;
 
 namespace Microsoft.DotNet.XHarness.CLI.Commands
 {
-    internal abstract class GetStateCommand : XHarnessCommand
+    internal abstract class GetStateCommand<TArguments> : XHarnessCommand<TArguments> where TArguments : IXHarnessCommandArguments
```
greenEkatherine, line 10: Is this abstract class needed since it used only for AppleGetStateCommand?

src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleUninstallCommand.cs
greenEkatherine, line 35: I would add variables' names here

src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleInstallCommand.cs
```diff
@@ -39,12 +39,12 @@ protected override Task<ExitCode> InvokeInternal(CancellationToken cancellationT
...
-                AppleAppArguments.AppPackagePath,
-                AppleAppArguments.Timeout,
-                AppleAppArguments.ResetSimulator,
-                AppleAppArguments.EnableLldb,
+                Arguments.Target,
+                Arguments.DeviceName,
+                Arguments.AppBundlePath,
+                Arguments.Timeout,
+                Arguments.ResetSimulator,
+                false,
```
greenEkatherine, line 47: please add variable name here as well

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Apple/Arguments/BundleIdentifierArgument.cs
```diff
@@ -0,0 +1,15 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.XHarness.Common.CLI.CommandArguments;
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Apple
+{
+    internal class BundleIdentifierArgument : RequiredStringArgument
+    {
+        public BundleIdentifierArgument() : base("app|a=", "Bundle identifier of the app that should be uninstalled")
```
greenEkatherine, line 11: not only uninstalled I guess

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Apple/Arguments/ShowDevicesUUIDArgument.cs
```diff
@@ -0,0 +1,16 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.XHarness.Common.CLI.CommandArguments;
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Apple
+{
+    internal class ShowDevicesUUIDArgument : SwitchArgument
+    {
+        public ShowDevicesUUIDArgument()
+            : base("include-devices-uuid", "Include the devices UUID. Defaults to false", true)
```
greenEkatherine, line 12: Defaults to true?

src/Microsoft.DotNet.XHarness.CLI/Commands/GetStateCommand.cs
premun: Good catch. I inherited Android one too and deleted some code there

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/ExpectedExitCodeArgument.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.XHarness.Common.CLI.CommandArguments;
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments
+{
+    /// <summary>
+    /// Exit code returned by the instrumentation for a successful run. Defaults to 0.
```
greenEkatherine, line 10:
```suggestion
    /// Exit code returned by the instrumentation for a successful run.
```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Apple/Arguments/CommunicationChannelArgument.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Apple
+{
+    /// <summary>
+    /// The way the simulator/device talks back to XHarness.
+    /// </summary>
+    internal class CommunicationChannelArgument : Argument<CommunicationChannel>
+    {
+        public CommunicationChannelArgument()
+            : base("communication-channel=", "The communication channel to use to communicate with the default", CommunicationChannel.UsbTunnel)
```
greenEkatherine, line 16: And here would be great to add info about default value

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Apple/Arguments/XmlResultJargonArgument.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Apple
+{
+    /// <summary>
+    /// Allows to specify the xml format to be used in the result files.
+    /// </summary>
+    internal class XmlResultJargonArgument : Argument<XmlResultJargon>
+    {
+        public XmlResultJargonArgument()
+            : base("xml-jargon=|xj=", $"The xml format to be used in the unit test results. Can be {XmlResultJargon.TouchUnit}, {XmlResultJargon.NUnitV2}, {XmlResultJargon.NUnitV3} or {XmlResultJargon.xUnit}", XmlResultJargon.xUnit)
```
greenEkatherine, line 16: + default value into the help message

src/Microsoft.DotNet.XHarness.Common/CLI/CommandArguments/Argument.cs
```diff
@@ -0,0 +1,275 @@
...
+            : base(prototype, description)
+        {
+            Value = defaultValue;
+        }
+
+        public override void Action(string argumentValue)
+        {
+            if (string.IsNullOrEmpty(argumentValue))
+            {
+                Value = !Value;
```
greenEkatherine, line 251: what happens if I occasionally put such argument twice?

premun: Not sure what is the right behaviour. Maybe this would be better?
```suggestion
                Value = !defaultValue;
```

premun: I mean if I stored the value

greenEkatherine: I think yes. Or somehow filter out more than one parameter because it's not obvious that the value would be unchanged after even amount of the repeated parameter

greenEkatherine (approved):

PR #638, opened 2021-06-15
Allow user to specify whether to search for wireless devices

### LINKED ISSUE #623: Disregarding wireless Apple devices speeds up device scan by 18 seconds
**Context**
- By default, mlaunch scans for devices connected by cable and also by WiFi
- The default timeout for wireless detection is 20 seconds, default for wired is 2 seconds
- All Helix devices are wired and we should then drop these extra 18 seconds since we don't need it
- This will speed up the runtime tests immensely and will also speed up regular device tests

**Goals**
- We should still make it possible for users of XHarness to scan for wireless devices so a switch should be added to any commands that scan for devices
- Regardless of what the default is, we should turn off wireless scan in Helix

premun:
Not searching for wireless devices saves 18 seconds per run. Default is now to not search for them.

Resolves #623

This is a breaking change in case you connect devices wirelessly (cc @mattleibow @PureWeen @akoeplinger @steveisok)

@mandel-macaque this is **not** a breaking change for Xamarin (shared code behaves the same)

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/Orchestration/BaseOrchestrator.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/Orchestration/BaseOrchestrator.cs
@@ -60,6 +60,7 @@ namespace Microsoft.DotNet.XHarness.Apple
         protected async Task<ExitCode> OrchestrateRun(
             TestTargetOs target,
             string? deviceName,
+            bool includeWirelessDevices,
             bool resetSimulator,
             bool enableLldb,
             AppBundleInformation appBundleInfo,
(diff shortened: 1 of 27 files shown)


akoeplinger (approved):

src/Microsoft.DotNet.XHarness.iOS.Shared/Execution/Arguments.cs
```diff
@@ -28,6 +28,16 @@ public ListDevicesArgument(string outputFile) : base("listdev", outputFile)
         }
     }
 
+    /// <summary>
+    /// When listing devices, should mlaunch also scan for wireless devices.
+    /// </summary>
+    public sealed class ListWirelessDevicesArgument : SingleValueArgument
+    {
+        public ListWirelessDevicesArgument(bool wirelessEnabled) : base("list-wireless-devices", wirelessEnabled ? "true" : "false", true)
```
premun, line 36: Should be `false` actually

premun:
I should also add a warning when doing simulators + wireless devices

src/Microsoft.DotNet.XHarness.Apple/Orchestration/BaseOrchestrator.cs
```diff
@@ -85,6 +85,11 @@ public abstract class BaseOrchestrator : IDisposable
                 }
             }
 
+            if (includeWirelessDevices && target.Platform.IsSimulator())
+            {
+                _logger.LogWarning("Including wireless devices while targeting a simulator has no effect");
```
greenEkatherine, line 90: It takes no effect but doesn't it increase timeout up to 20 seconds in that case? I mean if yes probably you should force its value to false

premun: @greenEkatherine valid observation but for simulators we do an entirely different mlaunch call and it has different timeouts

PR #641, opened 2021-06-15
Fix including ADB when packing CLI on non-Windows platforms

greenEkatherine:


premun (approved):
I would call the PR "Fix including ADB when packing CLI on non-Windows platforms" as that is the real change here

PR #645, opened 2021-06-17
Move XHarnessCommand and XHarnessCommandArguments to the CLI project

premun:
We don't need this anymore since we moved the SimulatorInstaller inside of the CLI project so let's prevent confusion


mdh1418 (approved):
Looks good to me, thanks!

PR #652, opened 2021-06-24
Initialize list arguments with non-null values

premun:


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Argument.cs
```diff
@@ -158,7 +157,7 @@ public override void Action(string argumentValue)
     public abstract class StringArgument : Argument<string?>
     {
         public StringArgument(string prototype, string description)
-            : base(prototype, description)
+            : base(prototype, description, null)
```
greenEkatherine, line 160: is it ok to have no opportunity to pass default value to string argument?

premun: It's a `string?` so I think this is expected

premun: There's also a `RequiredStringArgument` which has `string` only

greenEkatherine (approved):

PR #661, opened 2021-06-28
Redirect simulator logs into a separate file

### LINKED ISSUE #653: Collect more logs when Apple apps crash
Example run:
https://dev.azure.com/dnceng/public/_build/results?buildId=1202322&view=ms.vss-test-web.build-test-results-tab

We should check what all we can collect and are not collecting in these cases.

premun:
Use mlaunch capability that can redirect the output. Do this for Simulators only so that `--signal-app-end` keeps working for devices

Fixes #653

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppRunner.cs
@@ -271,14 +271,13 @@ namespace Microsoft.DotNet.XHarness.Apple
             IEnumerable<string> extraAppArguments,
             IEnumerable<(string, string)> extraEnvVariables)
         {
-            var appLog = _logs.CreateFile(appInformation.BundleIdentifier + ".log", LogType.ApplicationLog);
-            var appErrorLog = _logs.CreateFile(appInformation.BundleIdentifier + ".err.log", LogType.ApplicationLog);
-
             var args = GetCommonArguments(extraAppArguments, extraEnvVariables, appEndTag: null);
 
             args.Add(new SimulatorUDIDArgument(simulator.UDID));
+
+            var appLog = _logs.CreateFile(appInformation.BundleIdentifier + ".log", LogType.ApplicationLog);
             args.Add(new SetStdoutArgument(appLog));
-            args.Add(new SetStderrArgument(appErrorLog));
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppRunner.cs
```diff
@@ -273,17 +265,20 @@ public class AppRunner : AppRunnerBase
...
+            IEnumerable<(string, string)> extraEnvVariables)
         {
-            var args = GetCommonArguments(extraAppArguments, extraEnvVariables, appEndTag);
+            var args = GetCommonArguments(extraAppArguments, extraEnvVariables, appEndTag: null);
 
             args.Add(new SimulatorUDIDArgument(simulator.UDID));
 
+            var appLog = _logs.CreateFile(appInformation.BundleIdentifier + ".log", LogType.ApplicationLog);
+            args.Add(new SetStdoutArgument(appLog));
+            args.Add(new SetStderrArgument(appLog)); // Seems like mlaunch only redirects stderr, stdout doesn't produce any data
```
premun, line 280:
```suggestion
            args.Add(new SetStderrArgument(appLog)); // Seems like mlaunch only redirects stderr, stdout doesn't produce any data, however stderr captures stdout of the app too
```

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
```diff
@@ -553,11 +548,15 @@ public class AppTester : AppRunnerBase
...
                 extraEnvVariables,
-                appEndTag);
+                appEndTag: null);
 
             args.Add(new SetEnvVariableArgument(EnviromentVariables.HostName, "127.0.0.1"));
             args.Add(new SimulatorUDIDArgument(simulator));
 
+            var appLog = _logs.CreateFile(appInformation.BundleIdentifier + ".log", LogType.ApplicationLog);
+            args.Add(new SetStdoutArgument(appLog));
+            args.Add(new SetStderrArgument(appLog)); // Seems like mlaunch only redirects stderr, stdout doesn't produce any data
```
premun, line 558:
```suggestion
            args.Add(new SetStderrArgument(appLog)); // Seems like mlaunch only redirects stderr, stdout doesn't produce any data, however stderr captures stdout of the app too
```

akoeplinger (approved):

PR #665, opened 2021-06-29
Added NodeJS Engine

Daniel-Genkin:
Added a `NodeJS` engine when running xharness.

## Why is this needed
I am adding NodeJS support to MONO WASM (dotnet/runtime) and so this change would facilitate the testing as well as future maintenance of this feature.

## How is it used
Exactly the same as V8 or any other JavaScript engine.
Example: `XHarness wasm test --app=. --output-directory=".\xharness-output" --engine=NodeJS  --js-file=runtime.js   --   --run WasmTestRunner.dll System.AppContext.Tests.dll  -notrait category=OuterLoop -notrait category=failing` (this command runs some of the tests in dotnet/runtime on NodeJS)

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
@@ -56,17 +56,19 @@ namespace Microsoft.DotNet.XHarness.CLI.Commands.Wasm
             var engineBinary = Arguments.Engine.Value switch
             {
                 JavaScriptEngine.V8 => "v8",
-                JavaScriptEngine.NodeJS => "node",
                 JavaScriptEngine.JavaScriptCore => "jsc",
                 JavaScriptEngine.SpiderMonkey => "sm",
+                JavaScriptEngine.NodeJS => "node",
                 _ => throw new ArgumentException("Engine not set")
             };
 
-            if (engineBinary.Equals("node"))
-                engineBinary = FindEngineInPath(engineBinary + ".exe"); // NodeJS ships as .exe rather than .cmd
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) {
+                if (engineBinary.Equals("node"))
+                    engineBinary = FindEngineInPath(engineBinary + ".exe"); // NodeJS ships as .exe rather than .cmd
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -56,12 +56,16 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
             {
                 JavaScriptEngine.V8 => "v8",
+                JavaScriptEngine.NodeJS => "node",
                 JavaScriptEngine.JavaScriptCore => "jsc",
                 JavaScriptEngine.SpiderMonkey => "sm",
                 _ => throw new ArgumentException("Engine not set")
             };
 
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            if (engineBinary.Equals("node"))
```
premun, line 65: Why not just return `node.exe` from the switch?

premun (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
pavelsavara: Let's test how it ships on linux too

pavelsavara:
NodeJs is a Host not just Engine, it brings extra APIs.

Longer term perspective is that we may need to address feature-set unique to nodeJs, such as different networking stack and filesystem stack, among others. 

I wonder if we should rather add new `WasmTestNodeCommand` instead.

Consider unit tests which are 'browser only', those which run with `WasmTestBrowserCommand` rather than `WasmTestCommand`. Is there fetch polyfill that would make or browser the networking just work on Node ?

All of that could be answered in the future and we could merge this as is.

pavelsavara (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
Daniel-Genkin: Agreed, I actually had the same idea after I pushed but it took a while before I could actually test it haha. I just tested it on linux and windows and seems like it works as expected now

PR #683, opened 2021-07-26
Shorten timeout for finding Apple devices

premun:


src/Microsoft.DotNet.XHarness.iOS.Shared/Execution/Arguments.cs
```diff
@@ -415,4 +415,14 @@ public TcpTunnelArgument(int port)
 
         public override string AsCommandLineArgument() => $"--tcp-tunnel={_port}:{_port}";
     }
+
+    /// <summary>
+    /// Specify a timeout (in seconds) for the commands that dosn't have fixed duration.
```
greenEkatherine, line 420:
```suggestion
    /// Specify a timeout (in seconds) for the commands that doesn't have fixed duration.
```

greenEkatherine (approved):

PR #685, opened 2021-07-27
Handle bugreport generation for different APIs

### LINKED ISSUE #644: Error getting ADB bugreport on API 21
After app crashed XHarness failed to get bugreport with the error message:
```
fail: Error getting ADB bugreport: Exit code: 255 Standard Output: Standard Error: Failed to get bugreportz version: 'bugreportz -v' returned '/system/bin/sh: bugreportz: not found' (code 0). If the device does not run Android 7.0 or above, try 'adb bugreport' instead. 
```

Build https://dev.azure.com/xamarin/public/_build/results?buildId=41590&view=logs&j=dd18a2d5-2e4c-5ccd-8c00-4b4d347e66ba&t=883bb964-b431-5e95-fb8b-cec2c28425c6

/cc @mattleibow

greenEkatherine:
Resolves #644

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -23,6 +23,7 @@ namespace Microsoft.DotNet.XHarness.Android
         private const string AdbInstallBrokenPipeError = "Failure calling service package: Broken pipe";
         private const string AdbInstallException = "Exception occurred while executing 'install':";
         private const string AdbShellPropertyForBootCompletion = "sys.boot_completed";
+        private int _api;
         private readonly string _absoluteAdbExePath;
         private readonly ILogger _log;
         private readonly IAdbProcessManager _processManager;
@@ -31,7 +32,6 @@ namespace Microsoft.DotNet.XHarness.Android
             { "architecture", "shell getprop ro.product.cpu.abilist"},
             { "app", "shell pm list packages -3"}
         };
-
 
         public AdbRunner(ILogger log, string adbExePath = "") : this(log, new AdbProcessManager(log), adbExePath) { }
 
@@ -70,6 +70,9 @@ namespace Microsoft.DotNet.XHarness.Android
         public void SetActiveDevice(string? deviceSerialNumber)
         {
             _processManager.DeviceSerial = deviceSerialNumber ?? string.Empty;
+
+            var output = RunAdbCommand("shell getprop ro.build.version.sdk");
+            _api = int.Parse(output.StandardOutput);
 
             _log.LogInformation($"Active Android device set to serial '{deviceSerialNumber}'");
         }
@@ -129,19 +132,8 @@ namespace Microsoft.DotNet.XHarness.Android
 
         public void DumpBugReport(string outputFilePath)
         {
-            // give some time for bug report to be available
-            Thread.Sleep(3000);
-
-            var result = RunAdbCommand($"bugreport {outputFilePath}", TimeSpan.FromMinutes(5));
-            if (result.ExitCode != 0)
-            {
-                // Could throw here, but it would tear down a possibly otherwise acceptable execution.
-                _log.LogError($"Error getting ADB bugreport:{Environment.NewLine}{result}");
-            }
-            else
-            {
-                _log.LogInformation($"Wrote ADB bugreport to {outputFilePath}");
-            }
+            var reportManager = AdbReportFactory.CreateReportManager(_log, _api);
+            reportManager.DumpBugReport(this, outputFilePath);
         }
 
         public void WaitForDevice()
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -135,10 +135,10 @@ namespace Microsoft.DotNet.XHarness.Android
             }
         }
 
-        public void DumpBugReport(string outputFilePath)
+        public string DumpBugReport(string outputFilePathWithoutFormat)
         {
             var reportManager = AdbReportFactory.CreateReportManager(_log, _api ?? GetAPIVersion());
-            reportManager.DumpBugReport(this, outputFilePath);
+            return reportManager.DumpBugReport(this, outputFilePathWithoutFormat);
         }
 
         public void WaitForDevice()
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -23,7 +23,7 @@ namespace Microsoft.DotNet.XHarness.Android
         private const string AdbInstallBrokenPipeError = "Failure calling service package: Broken pipe";
         private const string AdbInstallException = "Exception occurred while executing 'install':";
         private const string AdbShellPropertyForBootCompletion = "sys.boot_completed";
-        private int _api;
+        private int? _api;
         private readonly string _absoluteAdbExePath;
         private readonly ILogger _log;
         private readonly IAdbProcessManager _processManager;
@@ -71,10 +71,15 @@ namespace Microsoft.DotNet.XHarness.Android
         {
             _processManager.DeviceSerial = deviceSerialNumber ?? string.Empty;
 
-            var output = RunAdbCommand("shell getprop ro.build.version.sdk");
-            _api = int.Parse(output.StandardOutput);
+            _api = GetAPIVersion();
 
             _log.LogInformation($"Active Android device set to serial '{deviceSerialNumber}'");
+        }
+
+        private int GetAPIVersion()
+        {
+            var output = RunAdbCommand("shell getprop ro.build.version.sdk");
+            return int.Parse(output.StandardOutput);
         }
 
         private static string GetCliAdbExePath()
@@ -132,7 +137,7 @@ namespace Microsoft.DotNet.XHarness.Android
 
         public void DumpBugReport(string outputFilePath)
         {
-            var reportManager = AdbReportFactory.CreateReportManager(_log, _api);
+            var reportManager = AdbReportFactory.CreateReportManager(_log, _api ?? GetAPIVersion());
             reportManager.DumpBugReport(this, outputFilePath);
         }
 
--- a/src/Microsoft.DotNet.XHarness.Android/Execution/AdbReportFactory.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/Execution/AdbReportFactory.cs
@@ -1,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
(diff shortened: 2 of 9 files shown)


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -129,19 +132,8 @@ public void DumpAdbLog(string outputFilePath, string filterSpec = "")
...
-            if (result.ExitCode != 0)
-            {
-                // Could throw here, but it would tear down a possibly otherwise acceptable execution.
-                _log.LogError($"Error getting ADB bugreport:{Environment.NewLine}{result}");
-            }
-            else
-            {
-                _log.LogInformation($"Wrote ADB bugreport to {outputFilePath}");
-            }
+            var reportManager = AdbReportFactory.CreateReportManager(_log, _api);
```
premun, line 135: Should we throw when `SetActiveDevice` was not called / `_api` is not set?

Or maybe getting API could be a method and we could do something like `_api ?? GetAPIVersion()` or we can just have `GetAPIVersion` cache the result and we could call it fast many times?

src/Microsoft.DotNet.XHarness.Android/Execution/BaseReportManager.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DotNet.XHarness.Android.Execution
+{
+    class BaseReportManager
```
premun, line 11: nit: "Base" usually suggests `abstract` class and I am not sure having `Base` and `Old` as options is intuitive. Maybe `ReportManager` and `OldReportManager` would make more sense? Or maybe we could even include the 23 info in the class name so that we have this knowledge somewhere else than the factory too - `Api23AndOlderReportManager`? It should definitely be conveyed somewhere in the class too (at least a doc header) so that when it's read in isolation, we know what it's for.

src/Microsoft.DotNet.XHarness.Android/Execution/BaseReportManager.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        public ILogger Logger { get; }
```
premun, line 13: Should this be `protected`?

```suggestion
        protected ILogger Logger { get; }
```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidRunCommand.cs
```diff
@@ -194,7 +194,7 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
 
                     if (processCrashed)
                     {
-                        runner.DumpBugReport(Path.Combine(outputDirectory, $"adb-bugreport-{apkPackageName}.zip"));
+                        runner.DumpBugReport(Path.Combine(outputDirectory, $"adb-bugreport-{apkPackageName}"));
```
premun, line 197: I think it might be a bit surprising for the user that the file won't be available on the path he provided in `outputFilePath`.

If we append a suffix, I would return the file name from the method so it's clear. Maybe even think about the name of the argument to be more explicit.

tests/Microsoft.DotNet.XHarness.Android.Tests/AdbRunnerTests.cs
```diff
@@ -87,11 +87,12 @@ public void DumpAdbLog()
         public void DumpBugReport()
         {
             var runner = new AdbRunner(_mainLog.Object, _processManager.Object, s_adbPath);
-            string pathToDumpBugReport = Path.Join(s_scratchAndOutputPath, $"{Path.GetRandomFileName()}.zip");
+            runner.SetActiveDevice(string.Empty);
+            string pathToDumpBugReport = Path.Join(s_scratchAndOutputPath, $"{Path.GetRandomFileName()}");
```
premun, line 91:
```suggestion
            string pathToDumpBugReport = Path.Join(s_scratchAndOutputPath, Path.GetRandomFileName());
```

tests/Microsoft.DotNet.XHarness.Android.Tests/AdbRunnerTests.cs
```diff
@@ -294,14 +295,18 @@ private ProcessExecutionResults CallFakeProcessManager(string process, string ar
                         if (s_bootCompleteCheckTimes > 0)
                         {
                             // Tell it we've booted.
-                            stdOut = "1{Environment.NewLine}";
+                            stdOut = $"1{Environment.NewLine}";
                         }
                         else
                         {
                             stdOut = $"{Environment.NewLine}";
```
premun, line 302:
```suggestion
                            stdOut = Environment.NewLine;
```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -71,6 +71,9 @@ public void SetActiveDevice(string? deviceSerialNumber)
         {
             _processManager.DeviceSerial = deviceSerialNumber ?? string.Empty;
 
+            var output = RunAdbCommand("shell getprop ro.build.version.sdk");
```
premun, line 74: I would extract getting API version as a separate method maybe

src/Microsoft.DotNet.XHarness.Android/Execution/AdbReportFactory.cs
```diff
@@ -0,0 +1,17 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DotNet.XHarness.Android.Execution
+{
+    internal class AdbReportFactory
+    {
+        internal static BaseReportManager CreateReportManager(ILogger log, int api)
```
premun, line 11: Can you add a doc for the method and explain what is `api`?

src/Microsoft.DotNet.XHarness.Android/Execution/BaseReportManager.cs
premun: Actually.. Better pattern is not to reuse dependencies from a parent but have your own copy - the idea is that you don't know what the parent might do with dependencies and so you create your own `private` field `_log` and work with it as this is now your class' implementation.

src/Microsoft.DotNet.XHarness.Android/Execution/BaseReportManager.cs
premun: On a second thought - your classes don't share any logic or common code - so it might be better to rather not inherit them and only create `IReportManager` and force the `DumpBugReport` through that.

If they start sharing something, I would ideally keep implementations as separate leaf children, so we would have
```
         BaseReportManager
	      /            \
OldReportManager      NewReportManager
```
and share the logic.

But I leave this up to you.

premun (changes requested):
Just minor things

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
greenEkatherine: In practice we call `SetActiveDevice` at the beginning of each command like `install/test/run`. I may add `GetAPIVersion()` to be sure in case if something changes in this logic.

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidRunCommand.cs
greenEkatherine: Why won't it be provided? I put the file there in both cases, the only difference is format. The name and format wasn't provided by user as well, we always generated it by ourselves and then put notify in the output.

premun: I mean it with regards to this specific API/method - you call the method:
```
runner.DumpBugReport("output/adb-bugreport-net.dot.whatever");
```

And after you call it, you would expect a file on a path that you supplied to an argument called `outputFilePath`.
But there won't be a file like that, there will be some `adb-bugreport-net.dot.whatever.zip` or `.txt` and if you want to copy it, read it.. it will be very complicated.

I understand we just put it in some directory that gets uploaded automatically and we don't care for the specific name in our case but that might be beyond the scope of the API of the method.

So my only suggestion is to change the method to return `string` instead of `void` and give us back the name of the file it actually produced. Possibly also change the name of the parameter a little so it's clear it's just a hint where to create the file.
This way, you will have a more intuitive API of the method as when you want to use it, you will know there's some string coming back and you will look into what it is.

src/Microsoft.DotNet.XHarness.Android/Execution/BaseReportManager.cs
greenEkatherine: I added `IReportManager`

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -71,9 +71,17 @@ public void SetActiveDevice(string? deviceSerialNumber)
...
             _processManager.DeviceSerial = deviceSerialNumber ?? string.Empty;
 
+            _api = GetAPIVersion();
+
             _log.LogInformation($"Active Android device set to serial '{deviceSerialNumber}'");
         }
 
+        private int GetAPIVersion()
+        {
+            var output = RunAdbCommand("shell getprop ro.build.version.sdk");
```
premun, line 81: Should we also null `_api` when we call `SetActiveDevice` again?

greenEkatherine: Sorry, I don't understand why null - to rewrite old value? Or maybe you mean update `_api`?

premun: I guess when we switch between devices, the API can change so I would null the `_api` field so that we know we need to fetch it again

greenEkatherine: but I updated `_api` inside `SetActiveDevice` - it should help in that case

premun: Ah, sorry, my bad. Then we're all good

premun (approved):

src/Microsoft.DotNet.XHarness.Android/Execution/IReportManager.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.XHarness.Android.Execution
+{
+    internal interface IReportManager
+    {
+        public string DumpBugReport(AdbRunner runner, string outputFilePathWithoutFormat);
```
premun, line 9:
```suggestion
        string DumpBugReport(AdbRunner runner, string outputFilePathWithoutFormat);
```

Issue #687, opened 2021-08-02
Add diagnostics mode


premun:
### Context
Every command should have a diagnostics mode when it would log useful data during the execution.

We will eventually start adding more data (used RAM, available storage..) but in the first phase we will only log every command and its result (exit code).

### Goals
- New `--diagnostics` option is added to every command
- XHarness CLI also accepts turning the diag mode on using an environment variable
- When in this mode, XHarness should create a (JSON) file and fill it with useful data such as:
    - Name of the executed command (test, run, install..)
    - Platform (android/apple/wasm)
    - Test target - which device/simulator was used (it's OS/API version..)
    - Exit code
- The data should be labeled so that later we can add the other metrics mentioned above.
- The option/variable might also hold value (path to where diagnostics file will be stored)

### Notes
- I guess the idea might be having something like a JSON file with an array of objects - each object is one metric.
- The data will be most likely parsed by Python later so if there's a better format to hand them over, JSON is not a hard requirement.
- If we have more data, such as error details, we can totally start collecting it already - Python just won't forward all of it to Application Insights.
- We need to make sure diagnostics data get outputted in all cases (crashes..)
- We should keep the data as uniform as possible across platforms so that parsing it can be simple too.

PR #699, opened 2021-08-26
Add command diagnostics for Android and Apple commands

premun:
Adds a new option/env variable that will make XHarness log diagnostic data for each executed command into a JSON file

https://github.com/dotnet/xharness/issues/687


premun:
@greenEkatherine ping

src/Microsoft.DotNet.XHarness.Common/CommandDiagnostics.cs
```diff
@@ -0,0 +1,142 @@
...
+                }
+                else
+                {
+                    var data = new[]
+                    {
+                        this
+                    };
+
+                    string json = JsonSerializer.Serialize(data, options);
+                    File.WriteAllText(targetFile, JsonSerializer.Serialize(data, options));
```
greenEkatherine, line 133:
```suggestion
                    File.WriteAllText(targetFile, json);
```

greenEkatherine (approved):

src/Microsoft.DotNet.XHarness.Common/CommandDiagnostics.cs
premun: Good catch

PR #720, opened 2021-09-27
Include installation time in `--launch-timeout` and create exit code for installation timeout

### LINKED ISSUE #719: Recognize and create new exit code for installation timeout
### Context
For iOS simulators, we occasionally see installation slowing down until it takes even 30 minutes. This is usually resolved by reboot of the machine.
To catch this early and not block the machine, we have the `--launch-timeout` option which measures how long it takes to boot the app.
When this happens, we return timeout only though.

We should be able to distinguish if the installation itself takes long and return with a unique exit code so that we can react to it in Helix SDK.

### Goal
Create new exit code for installation timeout and return it if we hit `--launch-timeout`.

premun:
The `--launch-timeout` only started running after we found the device and installed the app. Timing out installation operations where then not timing out properly.

Also adds a new exit code to distinguish failing installation and timing out installation.

Resolves #719


src/Microsoft.DotNet.XHarness.Apple/Orchestration/RunOrchestrator.cs
```diff
@@ -95,30 +97,56 @@ public class RunOrchestrator : BaseOrchestrator, IRunOrchestrator
...
+                }
+            });
+
+            using var launchTimeoutCancellationToken = CancellationTokenSource.CreateLinkedTokenSource(
+                launchTimeoutCancellation.Token,
+                cancellationToken);
+
+            Func<AppBundleInformation, Task<ExitCode>> executeMacCatalystApp = (appBundleInfo) =>
+            {
+                appRunStarted = true;
```
mandel-macaque, line 122: This is going to be a thread issue, there is a better way to deal with it without the need to do a continue with and specially with a more reliable way to set that the appRunStarted in a **saver** manner.

Because the appRunStarted not thread safe we potentially could have threading issues since we have two threads at least interested int he value:

1. The thread that sets the appRunStarted to true
2. The ContinueWith which is checking the value.

We can make a more robust code doing the following:

1. Move appRunStarted to be a TaskCompletionSource<bool> TunnelHoleThrough { get; }

```csharp
 TaskCompletionSource<bool> appRunStarted =  new TaskCompletionSource<bool>();

// later we can do
appRunStarted.TrySetResult(true);
```
2. Use a Task.WhenAny with the delay and the task of the completion source.
```
if (await Task.WhenAny(appRunStarted.Task, Task.Delay(timeout, timeoutCancellationTokenSource.Token)) == appRunStarted) {
// not timeout happened
} else {
// timeout happened.
}
```

mandel-macaque (changes requested):

src/Microsoft.DotNet.XHarness.Apple/Orchestration/RunOrchestrator.cs
premun: I went over the code and I don't believe I can do what you propose because my structure of code is a little bit different.

A bit simplified, I have a method like this:
```csharp
Run(Action blockB, CancellationToken ct)
{
	// Block of code A
    blockB();
	// Block of code C
}
```

I want to timescope `block of code A` using the mixture of a cancellation token and a timeout `launchTimeout`.
Once `block of code A` finishes, I want to keep using the cancellation token but drop the `launchTimeout` requirement.

In my method, I create a new cancellation that only cancels if a boolean flag is false:
```csharp
using var launchTimeoutCancellation = new CancellationTokenSource();
var appRunStarted = false;
var task = Task.Delay(launchTimeout < timeout ? launchTimeout : timeout).ContinueWith(t =>
{
    if (!appRunStarted)
    {
        launchTimeoutCancellation.Cancel();
    }
});
```

Then I give this **new** cancellation token to the `Run()` method, together with the `Action` that represents block B and which sets this flag:

```csharp
var blockB = () =>
{
    flag = true;
    // do some other work relying on the original cancellation token
};

Run(blockB, newCancellationToken);
```

The high-level idea I need is to create a `CancellationToken` and then revoke it, if it doesn't expire before a certain thing happens.

I cannot do `WhenAny` because I am not in control of `Run()`, I can only see it from the outside and inject a piece of code in the middle which, when executed, should void the original `launchTimeout`.

I cannot go into `Run()` because it is being used by many different classes and orchestrates high-level blocks of code that need to be executed. I could move the `launchTimeout` code inside `Run()` but some of the child classes have no clue of this concept so it would be a bit extra there.

I also don't think there's benefit in using TaskCompletionSource<bool> as I just have 1 bool flag which is only flipped to `true` and only by 1 thread. It is only read once by 1 thread too.

There cannot be any starvation as there's always the main execution thread running, 1 running with delay and checking the flag and ending and that's it.

premun: @mandel-macaque ping

mandel-macaque (approved):

PR #724, opened 2021-09-29
Allow printing ADB path using `xharness android state --adb`

premun:
Needed for https://github.com/dotnet/arcade/pull/7970 and for dotnet performance tests

Also drops "Get" from "*GetDevice" and "*GetState" commands to follow the rest of the commandset

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -35,6 +35,8 @@ namespace Microsoft.DotNet.XHarness.Android
 
         public int APIVersion => _api ?? GetAPIVersion();
 
+        public string AdbExePath => _absoluteAdbExePath;
+
         public AdbRunner(ILogger log, string adbExePath = "") : this(log, new AdbProcessManager(log), adbExePath) { }
 
         public AdbRunner(ILogger log, IAdbProcessManager processManager, string adbExePath = "")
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/Arguments/ShowAdbPathArgument.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Android
+{
+    internal class ShowAdbPathArgument : SwitchArgument
+    {
+        public ShowAdbPathArgument() : base("adb|show-adb-path", "Prints path to the adb executable XHarness is using", false)
```
MattGal, line 9: This isn't true; it also means in the current form that it skips the rest of the steps of the command that uses it?

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -51,17 +53,20 @@ public AdbRunner(ILogger log, IAdbProcessManager processManager, string adbExePa
...
+
             if (!File.Exists(_absoluteAdbExePath))
             {
                 _log.LogError($"Unable to find adb.exe");
                 throw new FileNotFoundException($"Could not find adb.exe. Either set it in the environment via {AdbEnvironmentVariableName} or call with valid path (provided:  '{adbExePath}')", adbExePath);
             }
+
             if (!_absoluteAdbExePath.Equals(adbExePath))
             {
                 _log.LogDebug($"ADBRunner using ADB.exe supplied from {adbExePath}");
```
MattGal, line 72: this line is already logging the thing this PR purports to be solving?

MattGal (commented):
Changes seem like a lot for what's happening here, especially given in the debug logging level it already logs where the exe is all the time.  (You also changed mac stuff unrelated to the title of the PR?)

Why not just expose the path on Adbrunner like you have, and just log it in https://github.com/dotnet/xharness/blob/c8c452d7746015847956618cf0c56f931ed8647a/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs#L32, without all the rest of this?

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/Arguments/ShowAdbPathArgument.cs
premun: Updated to note that it does only that then

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
premun: Yeah, but people explicitly asked to get the .exe path only not to rely on parsing some text output.

Also we will need this functionality too in Helix SDK to do some device cleaning so it's just easier to output the path only. Adding a new command seemed like an over-kill

MattGal (approved):
Approving, I get the use case now after discussing with @premun .  It may be more useful to just have a command that calls ADB and prints its output unfiltered by the log, but signing off so this can be used in the mean time.

PR #732, opened 2021-10-08
Fix race condition in SimpleFileListener

rolfbjarne:
The race condition goes like this:

* SimpleFileListener reads from one file, and writes to another.
* The reading will block until the SimpleFileListener is cancelled and
  everything has been read from the file (and written to the other file).
* When the SimpleFileListener is cancelled, the blocking read will detect
  this, and once everything has been read and written, call
  SimpleFileListener.Finished to mark that the blocking read/write was
  finished.
* The call to SimpleFileListener.Finished would notify anybody waiting for the
  listener to finish that the listener was done.
* Those waiters would now read from the listener's file output.
* The blocking read/write method would flush the output stream (because the
  call to SimpleFileListener.Finished happened inside the 'using' clause).

The fix is to move the call to SimpleFileListener.Finished to outside the
'using' clause of the output stream.


premun:
@rolfbjarne https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet-eng&package=Microsoft.DotNet.XHarness.CLI&protocolType=NuGet&version=1.0.0-prerelease.21508.1&view=overview

Issue #741, opened 2021-10-20
Apple exit code detection is flaky


premun:
For `apple run` and `apple just-run` commands, we try to detect the exit code from the system logs of the device/simulator. Sometimes the code is not detected.

The system log seems to have the line we want to get the exit code:
```
Oct 15 11:22:27 dci-mac-build-208 com.apple.CoreSimulator.SimDevice.3ACAFC62-D62A-437E-9DD8-83C910471723[886] (com.apple.security.cloudkeychainproxy3[14314]): Service exited with abnormal code: 1
Oct 15 11:22:27 dci-mac-build-208 com.apple.CoreSimulator.SimDevice.3ACAFC62-D62A-437E-9DD8-83C910471723[886] (com.apple.security.cloudkeychainproxy3): Service only ran for 0 seconds. Pushing respawn out by 10 seconds.
Oct 15 11:22:30 dci-mac-build-208 com.apple.CoreSimulator.SimDevice.3ACAFC62-D62A-437E-9DD8-83C910471723[886] (UIKitApplication:net.dot.JIT_HardwareIntrinsics[b296][rb-legacy][14312]): Service exited with abnormal code: 100
```

We look for this bit:
```
(UIKitApplication:net.dot.JIT_HardwareIntrinsics[b296][rb-legacy][14312]): Service exited with abnormal code: 100
```

But XHarness didn't find it.

premun:
This seems to have been fixed as the runtime tests are now green

PR #744, opened 2021-10-21
Attempt to fix Apple app exit code detection

premun:
When detecting exit code, the original file might be already gone but we have a copy by that time where the exit code should be found

Not sure this will resolve the issue though

#741


premun:
I think I've found the problem somewhere else

PR #757, opened 2021-11-02
Provide simulator/device flag in diagnostics data

premun:
We will need this flag later when detecting infra problems in Helix SDK

https://github.com/dotnet/arcade/issues/8109

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleAppCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleAppCommand.cs
@@ -10,6 +10,7 @@ using Microsoft.DotNet.XHarness.CLI.CommandArguments.Apple;
 using Microsoft.DotNet.XHarness.Common;
 using Microsoft.DotNet.XHarness.Common.CLI;
 using Microsoft.DotNet.XHarness.Common.Logging;
+using Microsoft.DotNet.XHarness.iOS.Shared;
 using Microsoft.DotNet.XHarness.iOS.Shared.Hardware;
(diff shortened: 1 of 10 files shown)


premun:
@greenEkatherine ping

src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleAppCommand.cs
```diff
@@ -51,6 +52,7 @@ protected sealed override async Task<ExitCode> Invoke(Extensions.Logging.ILogger
 
             var diagnosticsData = serviceProvider.GetRequiredService<IDiagnosticsData>();
             diagnosticsData.Target = Arguments.Target.Value.AsString();
+            diagnosticsData.IsDevice = !Arguments.Target.Value.Platform.IsSimulator();
```
greenEkatherine, line 55: May be it would be better to align naming with apple command - either `IsSimulator` or `IsDevice` for both

greenEkatherine (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/AppleAppCommand.cs
premun: Yeah, that would be nice, but unfortunately, I cannot change `IsSimulator` to `IsDevice` because Xamarin depends on this and I prefer to have `IsDevice` because it makes more sense for example in Apple where MacCatalyst has `IsSimulator == true` while it is not a simulator and it is not a device. So I think it would be better to have IsDevice even in this repo. But at least I make it less confusing down the line

PR #764, opened 2021-11-15
Add unit tests for Apple command implementations

### LINKED ISSUE #552: Add unit tests for apple commands

premun:
This PR adds tests to the Apple Orchestrator area which has 0% coverage right now.
The Orchestrator classes are high-level assemblies of operations with devices that implement the Apple commands.

📈 Repository coverage 50.93% → 57.47%
📈 Apple project coverage 40.86% → 70.87%

- Extracted interfaces from `AppOperation` classes to make them testable
- Injected hardcoded dependencies to `Orchestration` classes
- Added unit tests for `Orchestration` classes
- The only functional change is that we return `GENERAL_FAILURE` instead of `APP_LAUNCH_FAILURE` when we have un-caught exception from the `ExecuteApp` piece of the orchestration

Resolves #552


tests/Microsoft.DotNet.XHarness.Apple.Tests/Orchestration/InstallOrchestratorTests.cs
```diff
@@ -0,0 +1,301 @@
...
+            x => x.FindDevice(testTarget, null, It.IsAny<ILog>(), false),
+            Times.Once);
+
+        _appInstaller.Verify(
+            x => x.InstallApp(_appBundleInformation, testTarget, _device.Object, It.IsAny<CancellationToken>()),
+            Times.Once);
+    }
+
+    [Fact]
+    public async Task OrchestrateFailedDeviceInstallationTest()
```
tkapin, line 184: The test methods have identical or very similar structure and are quite copy-paste-ish. Would it be possible to refactor this?

tests/Microsoft.DotNet.XHarness.Apple.Tests/MockLogs.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Microsoft.DotNet.XHarness.Apple.Tests;
+
+#nullable disable
+public class MockLogs : List<IFileBackedLog>, ILogs
+{
+    public string Directory { get; set; } = "/tmp/logs/";
+
+    public IFileBackedLog AddFile(string path) => AddFile(path, path);
+
+    public IFileBackedLog AddFile(string path, string name)
```
tkapin, line 15: A nit: name would be better called description as name in case of file IO usually refers to file name. It's apparent from the line below, but still, would be a tiny bit more readable.

src/Microsoft.DotNet.XHarness.Apple/Orchestration/RunOrchestrator.cs
```diff
@@ -43,46 +42,41 @@ public interface IRunOrchestrator
...
             IHelpers helpers)
-            : base(processManager, deviceFinder, consoleLogger, logs, mainLog, errorKnowledgeBase, diagnosticsData, helpers)
-        {
-            _processManager = processManager ?? throw new ArgumentNullException(nameof(processManager));
-            _logger = consoleLogger ?? throw new ArgumentNullException(nameof(consoleLogger));
-            _logs = logs ?? throw new ArgumentNullException(nameof(logs));
-            _mainLog = mainLog ?? throw new ArgumentNullException(nameof(mainLog));
+            : base(appInstaller, appUninstaller, deviceFinder, consoleLogger, logs, mainLog, errorKnowledgeBase, diagnosticsData, helpers)
+        {
+            _iOSExitCodeDetector = iOSExitCodeDetector ?? throw new ArgumentNullException(nameof(iOSExitCodeDetector));
```
tkapin, line 68: Would it be possible to make the parameters non-nullable to avoid these checks?

tests/Microsoft.DotNet.XHarness.Apple.Tests/Orchestration/JustRunOrchestratorTests.cs
```diff
@@ -0,0 +1,369 @@
...
+        var testTarget = new TestTargetOs(TestTarget.Device_iOS, "14.2");
+
+        _exitCodeDetector
+            .Setup(x => x.DetectExitCode(_appBundleInformation, It.IsAny<IReadableLog>()))
+            .Returns(100)
+            .Verifiable();
+
+        var extraArguments = new[] { "--some arg1", "--some arg2" };
+
+        _appRunner
```
tkapin, line 129: Similar to above, the setup, result and verify are same or very similar to OrchestrateSimulatorJustRunTest.

tests/Microsoft.DotNet.XHarness.Apple.Tests/MockLogs.cs
premun: The parameter names come from the interface (which is shared with Xamarin). I think it makes more sense to keep them as this is a utility test class?

src/Microsoft.DotNet.XHarness.Apple/Orchestration/RunOrchestrator.cs
premun: It is non-nullable, this is a runtime check.

Afaik, it is still best practice to have them (since we also use DI).

tests/Microsoft.DotNet.XHarness.Apple.Tests/Orchestration/InstallOrchestratorTests.cs
premun: I factored out common code for Mock creation and verification of results in the test base class, such as:
```csharp
        VerifySimulatorReset(false);
        VerifySimulatorCleanUp(false);
```

Check [here](https://github.com/dotnet/xharness/blob/e74b9e0b5e69be451242a063eb2cc787a07abc42/tests/Microsoft.DotNet.XHarness.Apple.Tests/Orchestration/OrchestratorTestBase.cs) - this is common for all of these tests.

Things like `_appInstaller` setups differ in small details for every test - such as we return different the `ExitCode` or we check different parameters (`_device` vs `_simulator`), we are changing booleans and so on.
For example here we are testing the Install command - so we need different behavior of these in each setup so there is no common code to extract anymore.

Or do you have some repetitive specific code piece I didn't notice?

tests/Microsoft.DotNet.XHarness.Apple.Tests/Orchestration/JustRunOrchestratorTests.cs
premun: That setup differs in 3 parameters out of 9, we vary 6 of them in total. I could create a method for setting this up with these 9 optional params but it would be a lot of code because I would have to do it for each mocked method (we have 2 per AppRunner, 2 per AppTester..)

Not sure it's worth it?

tests/Microsoft.DotNet.XHarness.Apple.Tests/MockLogs.cs
premun: I agree Xamarin chose bad naming though

src/Microsoft.DotNet.XHarness.Apple/Orchestration/RunOrchestrator.cs
tkapin: Do you have some good read about that? I know that nullable reference types are compile time checked only, but a reasonable guidelines about when to use null guards and when not would be definitely useful for me.

tests/Microsoft.DotNet.XHarness.Apple.Tests/Orchestration/InstallOrchestratorTests.cs
tkapin: Discussed synchronously, you are probably right.

tests/Microsoft.DotNet.XHarness.Apple.Tests/Orchestration/JustRunOrchestratorTests.cs
tkapin: Same as our other conversation on this topic.

tests/Microsoft.DotNet.XHarness.Apple.Tests/Orchestration/InstallOrchestratorTests.cs
```diff
@@ -37,21 +38,8 @@ public InstallOrchestratorTests()
...
-            });
-
-        _appUninstaller
-            .Setup(x => x.UninstallSimulatorApp(_simulator.Object, s_appIdentifier, It.IsAny<CancellationToken>()))
-            .ReturnsAsync(new ProcessExecutionResult
-            {
-                ExitCode = 1, // This can fail as this is the first purge of the app before we install it
-                TimedOut = false,
-            });
+        SetupInstall(_simulator.Object);
```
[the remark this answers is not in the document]
premun: @tkapin can you approve the PR please?

tkapin (approved):

PR #766, opened 2021-11-16
Add PowerShell version of the install script

### LINKED ISSUE #761: Create XHarness bootstrap script for Windows too
### Context

There is a quick script at https://aka.ms/get-xharness which installs .NET and the tool locally.
It is very useful to have this for quick investigations or for example when DDFUN want to verify that they rebooted the device properly.

Currently, we have this for MacOS/Linux only but it would be nice to have it for Windows as well.

### Goal
Create a PowerShell / batch version and an aka.ms link to have the same story for Windows machines.

The bash version has also this quick command which can run it fast:
```
curl -L aka.ms/get-xharness | bash -
```

It would be nice to find the same alternative for Windows.

Related to #12892

premun:
Resolves #761

### DIFF
--- a/tools/Install-XHarness.ps1
+++ b/tools/Install-XHarness.ps1
@@ -25,7 +25,7 @@ Write-Host ".dotnet" -ForegroundColor Yellow
 Write-Host ".NET installed" -ForegroundColor Cyan
 Write-Host "Installing XHarness in current folder" -ForegroundColor Cyan
 
-Invoke-Expression -Command "& ./.dotnet/dotnet tool install --tool-path . --version '$xharness_version' --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json Microsoft.DotNet.XHarness.CLI"
+./.dotnet/dotnet tool install --tool-path . --version $xharness_version --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json Microsoft.DotNet.XHarness.CLI
 
 Write-Host "XHarness installed, run it using: " -ForegroundColor Cyan -NoNewline
 Write-Host ".\xharness help" -ForegroundColor Yellow
--- a/tools/Install-XHarness.ps1
+++ b/tools/Install-XHarness.ps1
@@ -1,0 +1,31 @@
+### This script is a quick way to install XHarness in a local folder.
+### .NET SDK is installed too, everything in a local folder which can be deleted afterwards.
+###
+### Use the following command to run this script from anywhere:
+### iex ((New-Object System.Net.WebClient).DownloadString('https://aka.ms/get-xharness-ps1'))
+
+param (
+    [Parameter(Mandatory = $false)]
+    [string]
+    $Version = "*"
+)
+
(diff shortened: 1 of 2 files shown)


premun (commented):
Please do not merge, I need to do it myself to fix the aka.ms links

tools/Install-XHarness.ps1
```diff
@@ -0,0 +1,31 @@
...
+
+$xharness_version = "1.0.0-prerelease.$Version"
+
+# Install .NET
+Write-Host "Getting dotnet-install.ps1.." -ForegroundColor Cyan
+Invoke-WebRequest -Uri "https://dot.net/v1/dotnet-install.ps1" -OutFile "dotnet-install.ps1"
+
+Write-Host "Installing .NET SDK locally to " -NoNewline -ForegroundColor Cyan
+Write-Host ".dotnet" -ForegroundColor Yellow
+Invoke-Expression -Command "& ./dotnet-install.ps1 -InstallDir ./.dotnet -Version 6.0.100"
```
lpatalas, line 23: You don't need `Invoke-Expression` here. You can just directly call what you have in `-Command` or even skip `&` if there are no spaces in the filename.
```suggestion
./dotnet-install.ps1 -InstallDir ./.dotnet -Version 6.0.100
```

tools/Install-XHarness.ps1
```diff
@@ -0,0 +1,31 @@
...
+
+Write-Host ".NET installed" -ForegroundColor Cyan
+Write-Host "Installing XHarness in current folder" -ForegroundColor Cyan
+
+Invoke-Expression -Command "& ./.dotnet/dotnet tool install --tool-path . --version '$xharness_version' --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json Microsoft.DotNet.XHarness.CLI"
```
lpatalas, line 28: Same as above:
```suggestion
./.dotnet/dotnet tool install --tool-path . --version $xharness_version --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json Microsoft.DotNet.XHarness.CLI
```

tools/Install-XHarness.ps1
```diff
@@ -0,0 +1,31 @@
...
+
+Write-Host "XHarness installed, run it using: " -ForegroundColor Cyan -NoNewline
+Write-Host ".\xharness help" -ForegroundColor Yellow
```
lpatalas, line 31: Nitpick: this is the only backslash in this file, all other paths use forward slash. :) Backslash is more typical on Windows but it should work either way.

premun: Good eye! I used forward slash because I hoped it would work on both and used backslash in output only because I expect people will actually only use it on Windows

tools/Install-XHarness.ps1
premun: Of course.. This line came from the copilot btw 😅

premun: I was too happy it gave me the full line that I forgot to read it

lpatalas (approved):

PR #767, opened 2021-11-18
Fix an issue with parsing test failures in NUnit v3 test files

rolfbjarne:
After finding a test failure in the xml, we need to position the xml reader at
the end of that test failure before trying to read another test failure.
Otherwise we end up rendering only the first one, because the rest of the code
will be confused.


premun:
@rolfbjarne seems like CRLF issues when running the new unit test under Windows:
https://dev.azure.com/dnceng/public/_build/results?buildId=1475567&view=ms.vss-test-web.build-test-results-tab&runId=42284730&resultId=100135&paneView=debug

rolfbjarne:
@premun it looks green now

premun (approved):

PR #768, opened 2021-11-18
Stop returning SIMULATOR_FAILURE (88) on failed uninstall

MattGal:
As requested in https://github.com/dotnet/core-eng/issues/15004 .  This, I assume, is to allow the Arcade wrapper scripts to handle it more accurately.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/ErrorKnowledgeBase.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/ErrorKnowledgeBase.cs
@@ -35,10 +35,6 @@ namespace Microsoft.DotNet.XHarness.Apple
             ["LSOpenURLsWithRole() failed with error -10825"] =
                 new("This application requires a newer version of MacOS",
                     suggestedExitCode: (int)ExitCode.GENERAL_FAILURE),
-
-            ["HE0018: Could not launch the simulator application"] =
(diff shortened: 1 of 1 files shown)


premun (approved):

premun:
@MattGal I drained the queue and the build here is now passing. I am pretty much sure this fixed the issue. Merging..

MattGal:
> @MattGal I drained the queue and the build here is now passing. I am pretty much sure this fixed the issue. Merging..

Let's check back on this after it flows to runtime, as I don't understand how this change could fix the issue until the users sending the "can't be uninstalled by XHarness" app setups would be impacted until then.

premun:
There are no "can't be uninstalled apps". We make a pre-emptive uninstall before the work item starts to start off clean (some apps actually create some data in the file system and those can then stick around and mess up the results). We do this for Android too actually.

Recently I found out that if I see a specific message/exit code in mlaunch logs when doing the pre-emptive uninstall, it is a good predictor of the simulator being in a corrupted state which would later manifest as failed app run. I did test everything on 3 broken machines with a bad simulator and this was a common feature I saw in all that disappeared after a reboot when simulators got also fixed.

I thought I would shave off 2-3 minutes of many work items, which would only detect a broken simulator by failing to run an app, by doing this early detection. This exit code 88 is currently only used in this scenario and I saw a high amount of these in the queue so I think it's the cause why we had so many retries/reboots.

PR #771, opened 2021-11-19
[wasm] write test as single line via base64

pavelsavara:


radical:
What kind of errors is this guarding against? Invalid xml only? We should catch those specific exceptions then.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
```diff
@@ -133,7 +136,14 @@ private void InvokeInternal(string message)
                     catch (Exception ex)
                     {
                         _logger.LogError($"Error while saving testResults.xml {ex}");
-                        throw;
```
maraf, line 136: Shouldn't we process this way only XML related exceptions? If we really want catch all exceptions this way, I would probably name the property differently.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
```diff
@@ -32,6 +34,7 @@ public WasmTestMessagesProcessor(string xmlResultsFilePath, string stdoutFilePat
             _stdoutFileWriter = File.CreateText(stdoutFilePath);
             _stdoutFileWriter.AutoFlush = true;
             _logger = logger;
+            InvalidXml = null;
```
maraf, line 37: Is it needed?

radical (commented):
Oops, forgot to submit!

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
```diff
@@ -133,7 +136,14 @@ private void InvokeInternal(string message)
                     catch (Exception ex)
                     {
                         _logger.LogError($"Error while saving testResults.xml {ex}");
-                        throw;
+                        // write it anyway
```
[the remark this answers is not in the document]
pavelsavara: It was trying to fail fast on corrupted xml. I abandon the idea now.

pavelsavara:
After the last [observation](https://github.com/dotnet/runtime/issues/61749#issuecomment-974501775) I realized that
- we still have multiple lines of console.log and therefore WS send with `STARTRESULTXML` and `ENDRESULTXML`.
- we still have multiple lines because xml file could contain `<![CDATA[some stuff]]>`

Therefore I made it really one line, which also simplified dealing with race condition (with lines from Chrome DevTools Protocol) on receiving side.

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
```diff
@@ -72,27 +72,16 @@ public static async Task<int> Run(string assemblyFileName, bool printXml, XunitF
                 {
                     var resultsXml = new XElement("assemblies");
                     resultsXml.Add(resultsXmlAssembly);
-                    using (var ms = new MemoryStream())
+                    int length;
+                    using (var sw = new StringWriter())
```
radical, line 76: `using var sw = new StringWriter();`

radical (dismissed):
Were you able to reproduce the issue locally? And does this fix it?

We should probably come up with a better way of doing this. Can we read files from the VFS after the app exits, but the page is still open?

radical:
The results file for `System.Memory.Tests` is ~12MB, which becomes ~16M in base64. If Larry's https://github.com/dotnet/runtime/pull/61854 works, with a chrome update from https://github.com/dotnet/runtime/pull/61856 , then we should use that, and write lines as we did earlier.

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
```diff
@@ -72,27 +72,16 @@ public static async Task<int> Run(string assemblyFileName, bool printXml, XunitF
...
-                                ms.CopyTo(stdout, 100 * 1024);
-                                twc.WriteLine();
-                                twc.WriteLine("ENDRESULTXML");
-                            }
-                        }
-                        Console.WriteLine($"Finished writing {ms.Length} bytes of RESULTXML");
+                        resultsXml.Save(sw);
+                        var bytes = System.Text.Encoding.UTF8.GetBytes(sw.ToString());
+                        length=bytes.Length;
+                        var base64 = Convert.ToBase64String(bytes, Base64FormattingOptions.None);
```
radical, line 81: this would affect non-wasm cases too

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
```diff
@@ -77,16 +80,27 @@ private void InvokeInternal(string message)
                 line = message.TrimEnd();
             }
 
-            if (_xmlResultsFileWriter == null)
+            var match = xmlRx.Match(line);
```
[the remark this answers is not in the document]
pavelsavara: So that I could capture the parts in same pass. Anyway, this is running on outside process and I'm not optimizing for speed.

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
pavelsavara: There is another usage in skiasharp which I have not noticed before. Thanks.

pavelsavara:
> The results file for `System.Memory.Tests` is ~12MB, which becomes ~16M in base64. If @lewing's [dotnet/runtime#61854](https://github.com/dotnet/runtime/pull/61854) works, with a chrome update from [dotnet/runtime#61856](https://github.com/dotnet/runtime/pull/61856) , then we should use that, and write lines as we did earlier.

I commented on 61854, I doubt it would help.

I think that 16MB should be fine. The ugly part is that emscripten stdout will do `put_char` on all of that, but we already have this problem and I only made it slightly worse.

If we get back to writing multiple lines, we need to prevent the race condition with lines from "chrome devtools protocol" by other means. My previous adding `isWebSocket` is one possible way.

pavelsavara:
> Were you able to reproduce the issue locally? 

No, not yet. Were you ?

> We should probably come up with a better way of doing this. Can we read files from the VFS after the app exits, but the page is still open?

The shell JS live V8 doesn't have this issue as far as I could tell. 

For the browser, we could replace this with HTTP POST to xharness web server and save it there. 
It would avoid all the contention and also all the buffering of stdout.
I'm not sure what would it take to bring HttpClient as new dependency of `ThreadlessXunitTestRunner`

maraf (approved):

maraf:
> For the browser, we could replace this with HTTP POST to xharness web server and save it there.
> It would avoid all the contention and also all the buffering of stdout.
> I'm not sure what would it take to bring HttpClient as new dependency of ThreadlessXunitTestRunner

I really like this HTTP idea.

radical:
> > For the browser, we could replace this with HTTP POST to xharness web server and save it there.
> > It would avoid all the contention and also all the buffering of stdout.
> > I'm not sure what would it take to bring HttpClient as new dependency of ThreadlessXunitTestRunner
> 
> I really like this HTTP idea.

we could even write it back over a different web socket, if we want to avoid using HttpClient.

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
```diff
@@ -1,28 +1,28 @@
-// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-#nullable enable
-
-using System;
-using System.Collections.Generic;
+// Licensed to the .NET Foundation under one or more agreements.
```
radical, line 1: Line endings of this file have changed, could you revert that back?

pavelsavara:
The problem is in how o get out of wasm into JS. The http is the easy part.
Currently we are abusing stdout.
Instead C# httpClient we could also recognize the pattern in JS at same place we have WS proxy and POST it from there.

I would merge this one so that we could iterrate with testing in in CI and will create another PR exploring this idea.

pavelsavara:
> we could even write it back over a different web socket, if we want to avoid using HttpClient.

I would like to avoid using C# HttpClient because I don't want to force the DLL dependency on all users of `ThreadlessXunitTestRunner`. On the other hand abusing `stdout` is no great either.

radical (approved):

pavelsavara:
I realized that we could combine multiple ideas.
@radical mentioned VFS.
We could use it to get data into JS and then we collect it via JS and upload with HTTP POST.

radical:
We could call a js function which would return the contents of the file from the VFS, or we could write that to some dom element, and read it from XHarness. We connect through selenium, so we can "talk" to the page.

PR #772, opened 2021-11-22
[wasm] fix XML header encoding

pavelsavara:
My previous change writes file with `<?xml version="1.0" encoding="utf-16"?>` in otherwise UTF8 file.


radical:
@pavelsavara if you build XHarness locally, then you can run runtime tests with `XHARNESS_CLI_PATH=/path/to/xharness/artifacts/bin/Microsoft.DotNet.XHarness.CLI/Debug/net6.0/Microsoft.DotNet.XHarness.CLI.dll` to run with your built XHarness, and catch any issues.

pavelsavara:
I know and I did, but I have not noticed the mismatch of xml header and actual encoding. @radical

PR #778, opened 2021-11-25
Add script for running E2E tests on Windows

premun:
We already have a shell script for Linux/MacOS but we could use this when working with the repo on Windows

### DIFF
--- a/tools/run-e2e-test.ps1
+++ b/tools/run-e2e-test.ps1
@@ -1,15 +1,21 @@
-### This script is a quick way to run the XHarness E2E tests.
-### These test are located in tests/integration-tests and require the Arcade and Helix SDK.
-### To run them, you need to invoke these through MSBuild which makes the process a bit cumbersome.
-### This script should make things easier.
-###
-### Usage: .\run-e2e-test.ps1 -TestProject Apple/SimulatorInstaller.Tests.proj [-SkipBuild]
+<#
+.SYNOPSIS
+  This script is a quick way to run the XHarness E2E tests.
+  These test are located in tests/integration-tests and require the Arcade and Helix SDK.
+  To run them, you need to invoke these through MSBuild which makes the process a bit cumbersome.
+  This script should make things easier.
+
+.EXAMPLE
+  .\run-e2e-test.ps1 -TestProject Apple/SimulatorInstaller.Tests.proj [-SkipBuild]
+#>
 
 param (
+    <# Path to the test project. Can be also relative to tests/integration-tests, e.g. Apple/Device.iOS.Tests.proj #>
     [Parameter(Mandatory = $true)]
     [string]
     $TestProject,
-    
+
+    <# Skip re-building the local package #>
     [switch]
     $SkipBuild = $false
 )
@@ -57,7 +63,7 @@ if ($SkipBuild) {
 
 $Env:BUILD_REASON = "pr"
 $Env:BUILD_REPOSITORY_NAME = "arcade"
-$Env:BUILD_SOURCEBRANCH = "prvysoky"
+$Env:BUILD_SOURCEBRANCH = "test"
 $Env:SYSTEM_TEAMPROJECT = "dnceng"
 $Env:SYSTEM_ACCESSTOKEN = ""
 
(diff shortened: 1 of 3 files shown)


tools/run-e2e-test.ps1
```diff
@@ -0,0 +1,66 @@
...
+
+    Remove-Item -Recurse -ErrorAction SilentlyContinue "$repoRoot\artifacts\tmp\Debug\Microsoft.DotNet.XHarness.CLI"
+    Remove-Item -Recurse -ErrorAction SilentlyContinue "$repoRoot\artifacts\artifacts\packages"
+
+    & "$repoRoot\Build.cmd" -pack -projects "$repoRoot\src\Microsoft.DotNet.XHarness.CLI\Microsoft.DotNet.XHarness.CLI.csproj"
+}
+
+$Env:BUILD_REASON = "pr"
+$Env:BUILD_REPOSITORY_NAME = "arcade"
+$Env:BUILD_SOURCEBRANCH = "prvysoky"
```
lpatalas, line 60: Are those `arcade` and `prvysoky` values correct?

tools/run-e2e-test.ps1
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    $files = Get-ChildItem -Include *Tests.proj -Recurse -Path "$repoRoot\tests\integration-tests\WASM"
+    foreach ($item in $files) {
+        '  - WASM\{0}' -f $item.Name
+    }
+}
+
+if (-not(Test-Path -Path $TestProject -PathType Leaf)) {
+    $TestProject = "$repoRoot\tests\integration-tests\$TestProject"
```
lpatalas, line 38: If this script is designed to be called manually then I think it would be more convenient if `$TestProject` was relative to current directory instead of `intergration-tests` folder because then I could use tab completion to fill the path instead of having to type or paste it manually.

tools/run-e2e-test.ps1
```diff
@@ -0,0 +1,66 @@
...
+    $SkipBuild = $false
+)
+
+$repoRoot = "$PSScriptRoot\.."
+
+function Write-Projects {
+    Write-Output "Possible options:"
+    $files = Get-ChildItem -Include *Tests.proj -Recurse -Path "$repoRoot\tests\integration-tests\Android"
+    foreach ($item in $files) {
+        '  - Android\{0}' -f $item.Name
```
lpatalas, line 23: This will show invalid path if project is in subfolder.

tools/run-e2e-test.ps1
```diff
@@ -0,0 +1,66 @@
...
+        Write-Error "The file $TestProject not found"
+        Write-Projects
+        Exit 1
+    }
+}
+
+if ($SkipBuild) {
+    Write-Host -ForegroundColor Cyan "> Skipping build"
+} else {
+    Write-Host -ForegroundColor Cyan "> Building Microsoft.DotNet.XHarness.CLI.1.0.0-dev.nupkg"
```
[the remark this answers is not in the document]
premun: Arcade makes it so that locally built packages have always this version. Packages built in AzDO then would be `1.0.0-ci`

tools/run-e2e-test.ps1
premun: That works too, just if it doesn't find it, it tries the other version (relative to integration-tests):
```
λ  .\tools\run-e2e-test.ps1 .\tests\integration-tests\Apple\Simulator.Commands.Tests.proj
```


I kept this behavior same as in the shell version of the script.

tools/run-e2e-test.ps1
premun: Yeah, I didn't know how to handle this part well in PowerShell. We have so far this structure and we actually do not want to go to subfolders.

In bash, I just do `find` with `maxDepth` and then process it. In PowerShell I didn't know how to strip the first part of the absolute path.

lpatalas: You can do similar thing in PowerShell:

```
$testRoot = "$repoRoot\tests\integration-tests\"
Get-ChildItem -Recurse -Include *Tests.proj -Path $testRoot | ForEach-Object { $_.FullName.Substring($testRoot.Length) }
```

lpatalas (approved):

PR #782, opened 2021-12-01
Add `android adb` command

### LINKED ISSUE #726: Create `android adb` command
### Context
Sometimes we want to use the bundled ADB directly. Currently, we can call `android state --adb` to get path to the ADB executable and then run that.

Follow up of #724 

### Goals
- Add the `android adb` command which would just forward the arguments to adb and pipe the output/exit code.
- The command should accept XHarness arguments still (like `-v` or `-h`) and only forward the `PassThroughArguments` which are deliminated by `--` so for example:
    ```
    xharness android adb -v -- devices list
    ```
- We should make sure we run ADB in the same way we run programs in Apple, so with a `ProcessManager`
- We need to make sure we pipe the output well and not buffer it in memory to prevent common CLI issues
- Once this is finished, we should remove the `--adb` options from the `android state` command (revert that whole [change](https://github.com/dotnet/xharness/pull/724))

premun:
Resolves #726

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
@@ -27,11 +27,7 @@ public class AdbProcessManager : IAdbProcessManager
             arguments = arguments.Prepend(DeviceSerial).Prepend("-s");
         }
 
-        var processArgs = StringUtils.FormatArguments(arguments.ToArray());
-
-        _log.LogDebug($"Executing command: '{adbExePath} {processArgs}'");
-
-        var processStartInfo = new ProcessStartInfo
+        var processStartInfo = new ProcessStartInfo()
         {
             CreateNoWindow = true,
             UseShellExecute = false,
@@ -39,10 +35,17 @@ public class AdbProcessManager : IAdbProcessManager
             RedirectStandardOutput = true,
             RedirectStandardError = true,
             FileName = adbExePath,
-            Arguments = processArgs,
         };
 
+        foreach (var arg in arguments)
+        {
+            processStartInfo.ArgumentList.Add(arg);
+        }
+
+        _log.LogDebug($"Executing command: '{adbExePath} {StringUtils.FormatArguments(processStartInfo.ArgumentList)}'");
+
         var p = new Process() { StartInfo = processStartInfo };
+
         var standardOut = new StringBuilder();
         var standardErr = new StringBuilder();
 
@@ -69,7 +72,6 @@ public class AdbProcessManager : IAdbProcessManager
         };
 
         p.Start();
-
         p.BeginOutputReadLine();
         p.BeginErrorReadLine();
 
--- a/src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
@@ -1,6 +1,8 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Text;
 using Microsoft.DotNet.XHarness.Common.Utilities;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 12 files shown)


src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
```diff
@@ -90,15 +89,15 @@ public ProcessExecutionResults Run(string adbExePath, string arguments, TimeSpan
         p.Close();
 
         lock (standardOut)
-            lock (standardErr)
+        lock (standardErr)
```
lpatalas, line 92: What are those locks needed for?

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidAdbCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        var runner = new AdbRunner(logger);
+
+        try
+        {
+            var result = runner.RunAdbCommand(string.Join(" ", PassThroughArguments), Arguments.Timeout);
+
+            Console.Write(result.StandardOutput);
+            Console.Error.Write(result.StandardError);
```
lpatalas, line 48: This will write the output in different order if stdout and stderr were originally interleaved but maybe that's not a problem in this case?

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidAdbCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+                $"    android adb --timeout 00:01:30 -- devices -l");
+
+            return Task.FromResult(ExitCode.INVALID_ARGUMENTS);
+        }
+
+        var runner = new AdbRunner(logger);
+
+        try
+        {
+            var result = runner.RunAdbCommand(string.Join(" ", PassThroughArguments), Arguments.Timeout);
```
lpatalas, line 45: Does it handle white space in the arguments correctly? For example if I call `xharness adb -- "some arg with spaces"` it should pass it as a single argument.

lpatalas (commented):
Just a few notes/questions but overall it looks good.

src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
premun: I guess they make sure we ended writing the output before returning from here.
I guess this was some auto-format change of this file

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidAdbCommand.cs
premun: Yeah, I was thinking of adding some switch that would use only one StringBuilder to merge the two in the process manager but I tried calling adb.exe manually couple of times and the output it gave seemed to look fine enough like this. I would keep it like this and improve if there is some case where this is a problem (the code change was quite considerable)

src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
lpatalas: I looked at history of this file and at the beginning they were there to solve some concurrency issues but in the meantime the code was changed a bit and now the process is closed before we even get to this section. I'm pretty sure they are not needed any more but who knows. Process API has a lot of strange edge cases.

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidAdbCommand.cs
premun: Good point. I blindly trusted the existing Android code but now I verified it and we were not doing this so I added quotation from Apple side (we unfortunately have different process managers 😥 for some reason)

src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
premun: I agree, seems like the process should be long dead when we get here

premun: Only if the Kill failed..

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidAdbCommand.cs
lpatalas: I think the most correct way to handle this would be to keep the arguments as an array and use `ProcessStartInfo.ArgumentList` instead of `Arguments`. It's more errorproof than making sure that everything is quoted correctly.

lpatalas (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidAdbCommand.cs
premun: Oooh, I was too fast and didn't even think for a second `arguments` is a `string` 😅

I see the failures in tests now:
```
Executing command: 'adb.exe  "devices -l"'
```

premun: I think the most correct way would be to switch to a single implementation in this repo which would be the one coming from Xamarin

premun: @lpatalas I switched `AdbRunner` to accept an array of arguments instead. It uncovered many places where the code wasn't escaping potential problems in paths and other places.

src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs
```diff
@@ -34,8 +39,9 @@ public ProcessExecutionResults Run(string adbExePath, string arguments, TimeSpan
             RedirectStandardOutput = true,
             RedirectStandardError = true,
             FileName = adbExePath,
-            Arguments = $"{deviceSerialArgs} {arguments}",
+            Arguments = processArgs,
```
lpatalas, line 42: Now that you have arguments in a list you should assign it to `ArgumentList` instead of `Arguments`. This will take care of escaping them properly.

premun: Ah, right. I need to then switch the Xamarin code to use `ArgumentList` too (need to talk to them first) because we could remove a lot of code then.

lpatalas (approved):

PR #784, opened 2021-12-02
Add `apple simulators reset-simulator` command

### LINKED ISSUE #777: Create `apple reset-simulator` command
### Motivation
Currently, we have the `--reset-simulator` switch which wipes the simulator clean.
For some cases (simulator is broken / its memory is full), we might want to call the reset only but it is not possible - always a part of some other command.

### Goal
Add a new command that would reset given simulator.

premun:
Resolves #777

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/Orchestration/SimulatorResetOrchestrator.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/Orchestration/SimulatorResetOrchestrator.cs
@@ -58,10 +58,10 @@ public class SimulatorResetOrchestrator : BaseOrchestrator, ISimulatorResetOrche
             return Task.FromResult(ExitCode.INVALID_ARGUMENTS);
         }
 
-        static Task<ExitCode> executeMacCatalystApp(AppBundleInformation appBundleInfo)
+        static Task<ExitCode> ExecuteMacCatalystApp(AppBundleInformation appBundleInfo)
             => throw new InvalidOperationException("reset-simulator command not available on maccatalyst");
 
-        static Task<ExitCode> executeApp(AppBundleInformation appBundleInfo, IDevice device, IDevice? companionDevice)
+        static Task<ExitCode> ExecuteApp(AppBundleInformation appBundleInfo, IDevice device, IDevice? companionDevice)
             => Task.FromResult(ExitCode.SUCCESS); // no-op
 
         return OrchestrateOperation(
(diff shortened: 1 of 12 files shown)


src/Microsoft.DotNet.XHarness.Apple/Orchestration/SimulatorResetOrchestrator.cs
```diff
@@ -0,0 +1,87 @@
...
+        if (!target.Platform.IsSimulator() || target.Platform.ToRunMode() == RunMode.MacOS)
+        {
+            _consoleLogger.LogError($"The simulator reset action requires a simulator target while {target.AsString()} specified");
+            return Task.FromResult(ExitCode.INVALID_ARGUMENTS);
+        }
+
+        static Task<ExitCode> executeMacCatalystApp(AppBundleInformation appBundleInfo)
+            => throw new InvalidOperationException("reset-simulator command not available on maccatalyst");
+
+        static Task<ExitCode> executeApp(AppBundleInformation appBundleInfo, IDevice device, IDevice? companionDevice)
```
lpatalas, line 64: Both execute methods start with lowercase. Is this intended?

tests/Microsoft.DotNet.XHarness.Apple.Tests/Orchestration/SimulatorResetOrchestratorTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        // Verify
+        Assert.Equal(ExitCode.INVALID_ARGUMENTS, result);
+
+        _deviceFinder.VerifyNoOtherCalls();
+        _appInstaller.VerifyNoOtherCalls();
+        _appUninstaller.VerifyNoOtherCalls();
+    }
+
+    [Fact]
+    public async Task TryMaccatalystResetTest()
```
lpatalas, line 87:
```suggestion
    public async Task TryMacCatalystResetTest()
```

lpatalas (approved):
I don't have enough context to review it in depth but what I see looks good :)

src/Microsoft.DotNet.XHarness.Apple/Orchestration/SimulatorResetOrchestrator.cs
premun: They used to be lambdas and I think an auto-refactoring changed them to local methods. Do you know what's the naming convention for these?

lpatalas: I have no idea. I didn't even notice that these are local functions before you wrote it. I would guess they should be uppercase.

PR #791, opened 2021-12-08
Refactor target-specific ApplicationEntryPoint subclasses to have non-xunit-specific base classes

jkoritzinsky:
Today, xharness implements the `AndroidApplicationEntryPoint`, `iOSApplicationEntryPoint` and `WasmApplicationEntryPoint` classes as xunit-specific implementations. Additionally, the `WasmApplicationEntryPoint` type didn't derive at all from `ApplicationEntryPoint` and didn't have good customization points, so it wasn't possible to replace the runner with a different runner.

For the test refactoring work that @trylek and I are doing in dotnet/runtime to make the runtime tests run more like regular tests, we're using a custom source-generated test runner. As a result, we're going to have a test runner/executor that is not the Xunit-provided runner, but we'll need to provide custom runners for each platform to provide our filtering and test execution mechanisms.

To limit the number of assemblies loaded and to ensure that we aren't depending on XUnit but still get the features that xharness uses to run tests correctly, I figured that the easiest way was to refactor the entrypoint classes to have non-Xunit-specific base implementations per-target platform. I also refactored the Wasm entry-point and runner to conform to the `TestRunner` and `ApplicationEntryPoint` base classes so Wasm test runs with our new execution model will be able to easily use the new runner.

I've done my best to keep API and ABI compatibility so as to not break consumers when they pull down the newer versions of the libraries.

cc: @fanyang-mono @naricc

### DIFF
--- a/src/Microsoft.DotNet.XHarness.TestRunners.Common/WasmApplicationEntryPointBase.cs
+++ b/src/Microsoft.DotNet.XHarness.TestRunners.Common/WasmApplicationEntryPointBase.cs
@@ -10,44 +10,16 @@ using System.Threading.Tasks;
 #nullable enable
 namespace Microsoft.DotNet.XHarness.TestRunners.Common;
 
-public abstract class WasmApplicationEntryPointBase : ApplicationEntryPoint, IDevice
+public abstract class WasmApplicationEntryPointBase : ApplicationEntryPoint
 {
-    public string BundleIdentifier => string.Empty;
-
-    public string UniqueIdentifier => string.Empty;
-
-    public string Name => string.Empty;
-
-    public string Model => string.Empty;
-
-    public string SystemName => string.Empty;
-
-    public string SystemVersion => string.Empty;
-
-    public string Locale => string.Empty;
-
     protected override int? MaxParallelThreads => 1;
 
-    protected override IDevice Device => this;
+    protected override IDevice? Device => null;
 
     public override async Task RunAsync()
     {
         var options = ApplicationOptions.Current;
-        // we generate the logs in two different ways depending if the generate xml flag was
-        // provided. If it was, we will write the xml file to the tcp writer if present, else
-        // we will write the normal console output using the LogWriter
-        var logger = new LogWriter(Device);
-        logger.MinimumLogLevel = MinimumLogLevel.Info;
-
-        var runner = await InternalRunAsync(logger);
-        WriteResults(runner, options, logger, Console.Out);
-
-        logger.Info($"Tests run: {runner.TotalTests} Passed: {runner.PassedTests} Inconclusive: {runner.InconclusiveTests} Failed: {runner.FailedTests} Ignored: {runner.FilteredTests}");
-
-        if (options.AppEndTag != null)
-        {
-            logger.Info(options.AppEndTag);
-        }
+        var runner = await InternalRunAsync(options, null, Console.Out);
 
         LastRunHadFailedTests = runner.FailedTests != 0;
     }
--- a/src/Microsoft.DotNet.XHarness.TestRunners.Common/WasmApplicationEntryPointBase.cs
+++ b/src/Microsoft.DotNet.XHarness.TestRunners.Common/WasmApplicationEntryPointBase.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.XHarness.TestRunners.Common;
+
+public abstract class WasmApplicationEntryPointBase : ApplicationEntryPoint, IDevice
+{
+
+    public string BundleIdentifier => string.Empty;
+
+    public string UniqueIdentifier => string.Empty;
+
+    public string Name => string.Empty;
+
+    public string Model => string.Empty;
+
+    public string SystemName => string.Empty;
+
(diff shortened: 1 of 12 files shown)


src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
```diff
@@ -6,97 +6,135 @@
...
 
             Console.WriteLine($"Discovering: {assemblyFileName} (method display = {discoveryOptions.GetMethodDisplayOrDefault()}, method display options = {discoveryOptions.GetMethodDisplayOptionsOrDefault()})");
-            var assembly = Assembly.LoadFrom(assemblyFileName);
-            var assemblyInfo = new global::Xunit.Sdk.ReflectionAssemblyInfo(assembly);
+            var assemblyInfo = new global::Xunit.Sdk.ReflectionAssemblyInfo(testAsmInfo.Assembly);
             var discoverer = new ThreadlessXunitDiscoverer(assemblyInfo, new NullSourceInformationProvider(), discoverySink);
 
             discoverer.FindWithoutThreads(includeSourceInformation: false, discoverySink, discoveryOptions);
-            var testCasesToRun = discoverySink.TestCases.Where(filters.Filter).ToList();
+            var testCasesToRun = discoverySink.TestCases.Where(t => _filters.IsExcluded(t)).ToList();
```
pavelsavara, line 59: The condition seems to be negated here.

src/Microsoft.DotNet.XHarness.TestRunners.Common/WasmApplicationEntryPointBase.cs
```diff
@@ -0,0 +1,54 @@
...
+        // we generate the logs in two different ways depending if the generate xml flag was
+        // provided. If it was, we will write the xml file to the tcp writer if present, else
+        // we will write the normal console output using the LogWriter
+        var logger = new LogWriter(Device);
+        logger.MinimumLogLevel = MinimumLogLevel.Info;
+
+        var runner = await InternalRunAsync(logger);
+        WriteResults(runner, options, logger, Console.Out);
+
+        logger.Info($"Tests run: {runner.TotalTests} Passed: {runner.PassedTests} Inconclusive: {runner.InconclusiveTests} Failed: {runner.FailedTests} Ignored: {runner.FilteredTests}");
```
premun, line 41: Can we extract this result line into some common property?

I already see there's a divide which we probably don't want:
image

premun (commented):
Yeah, this should work better! Just a couple of notes

jkoritzinsky:
@premun @pavelsavara I've updated the PR based on your feedback. It's ready for another review pass.

premun (approved):
@jkoritzinsky LGTM but please bare in mind that `TestRunner` is currently the only part of XHarness that has no test coverage in this repo so a green PR build doesn't mean much in this sense. We are not able to build apps in this repo so we cannot utilize it and we don't have any unit tests there neither (the code was ported from Xamarin like this).

Feel free to merge and then watch the dependency bump in runtime. However, if you could (maybe you did) test this locally by injecting this test runner into some test (WASM/Apple/Android), we would know sooner.

I personally don't mind that much breaking it if we watch the dependency PR in runtime but also considering we won't need many iterations to fix it.

pavelsavara:
We discussed this with Jeremy yesterday and I think he tested it.

The steps to test WASM with runtime repo are following:
1) compile your local xharness
2) point the environment  to it
`set XHARNESS_CLI_PATH=c:\Dev\xharness\artifacts\bin\Microsoft.DotNet.XHarness.CLI\Debug\net6.0\Microsoft.DotNet.XHarness.CLI.dll`
3) compile wasm target on runtime repo
`build.cmd -bl -os Browser -subset mono+libs+host+packs -c Debug`
4) run at least one tests
`dotnet.cmd build /bl:test.binlog /p:TargetOS=Browser /p:TargetArchitecture=wasm /p:Configuration=Debug /t:Test src/libraries/System.Private.Runtime.InteropServices.JavaScript/tests`

maraf:
> The steps to test WASM with runtime repo are following:

I don't think this is enough, as changes here touches the runner, we need to build and use nuget packages too.

premun:
I don't think this does the trick to be honest. This just uses the XHarness CLI to run the tests. What is important is to use the TestRunner DLL from locally built XHarness repo and use it when building the test applications.

The steps are something like this:
	1. Change `nuget.config` in `dotnet/runtime` and add `xharness/artifacts/packages/…/Shipping` as a nuget source
	2. Change `Versions.props` to use `1.0.0-dev` as the version of the TestRunner
	3. Run `./build.sh -build -pack` (or `Build.cmd`) in XHarness
	4. Delete `~/.nuget/packages/..xharness../1.0.0-dev`
	5. Run `./build.sh libs.pretest -os .. -arch ...`
    6. Build the test apps

jkoritzinsky:
I already started on the path of manually consuming this in runtime, so I'll do some local validation based on the steps you mentioned and report back.

jkoritzinsky:
I've validated locally that the WASM test runner from this PR works correctly with dotnet/runtime. As that was the runner that went through the most design changes in this PR, I feel comfortable merging this PR in. I'll watch for the dependency update PR in dotnet/runtime and make sure that everything is flowing smoothly.

PR #795, opened 2021-12-17
Get Android device architecture for diagnostic data

### LINKED ISSUE #794: The `android run` command doesn't include architecture in diagnostics data
The `android run` command returns `_API 29` in the `target_targetOS` diagnostics field:

image

(the `test` command shows the architecture)

This might be because we don't have it available since we are not working with the APK in this command but it should be investigated.

premun:
Currently, we do not include device's architecture in diagnostic data but rather the first of a list of all supported archs by the device. This PR adds a new ADB call that queries for the actual architecture of the device.

Secondly, this PR brings the Android and Apple `state` commands on-par and allows to export Android state information as JSON too (useful for querying devices over Helix for example).

Fixes #794

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
@@ -99,7 +99,7 @@ internal class AndroidGetStateCommand : GetStateCommand<AndroidGetStateCommandAr
         var state = runner.GetAdbState().Trim();
 
         Dictionary<string, string?> deviceAndArchList = runner.GetAttachedDevicesWithProperties("architecture");
-        var allDevices = deviceAndArchList
+        List<DeviceInfo> allDevices = deviceAndArchList
             .Select(d => new DeviceInfo(
                 Name: d.Key,
                 Architecture: runner.GetDeviceArchitecture(logger, d.Key) ?? "unknown",
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
@@ -3,6 +3,9 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.DotNet.XHarness.Android;
 using Microsoft.DotNet.XHarness.CLI.CommandArguments.Android;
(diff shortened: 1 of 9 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
```diff
@@ -25,24 +28,55 @@ public AndroidGetStateCommand() : base(TargetPlatform.Android, new ServiceCollec
...
+                {
+                    WriteIndented = true
+                };
 
-            logger.LogInformation($"List of devices:");
-            var deviceAndArchList = runner.GetAttachedDevicesWithProperties("architecture");
-            foreach (string device in deviceAndArchList.Keys)
+                JsonSerializer.Serialize(Console.OpenStandardOutput(), data, options);
+            }
+            else
```
tkapin, line 44: In such long methods, it'd be more readable to return early instead of branching to else. The difference is that with returning early, I can immediately see that the method is done, with else block I have to scroll down to check if some logic follows yet. Also reduces indentation.

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
```diff
@@ -25,24 +28,55 @@ public AndroidGetStateCommand() : base(TargetPlatform.Android, new ServiceCollec
...
             {
-                logger.LogInformation($"Device: '{device}' - Architecture: {deviceAndArchList[device]}");
+                var state = data.DeviceState switch
+                {
+                    "device" => "Device/emulator is ready",
+                    null or "" => "No device attached",
+                    _ => data.DeviceState,
+                };
+
+                void PrintDeviceInfo(DeviceInfo device)
```
tkapin, line 53: A nit (even nittier when it's a local function) - a function that only logs would be better called "Log<something>" to make clear it just logs and does nothing else.

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
```diff
@@ -53,4 +87,33 @@ protected override Task<ExitCode> InvokeInternal(ILogger logger)
...
+        var runner = new AdbRunner(logger);
+
+        logger.LogDebug("Getting state of ADB and attached Android device(s)");
+
+        var adbVersion = runner.GetAdbVersion()
+            .Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+
+        var state = runner.GetAdbState().Trim();
+
+        var deviceAndArchList = runner.GetAttachedDevicesWithProperties("architecture");
```
tkapin, line 102: Suggestion: when parsing some output, it might be useful to provide example output that is to be parsed in comments.

tkapin (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
premun: Is not `Print` better?

For me `Log` means just dump it in the log file/stream/whatever. `Print` always had the connotation of "prettify for reading in console/on the screen" (which is what is happening here).

But maybe that's just my explanation?

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
premun: Agreed, done

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
premun: Good point, I changed `var` to `Dictionary<string, string?>` so it's easier to read.

tkapin (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidGetStateCommand.cs
tkapin: For me, I see jus call into logger.LogInformation() so for me it's technically logging, but as said, no big deal.

PR #805, opened 2022-01-19
Add retries to `PlistBuddy` calls

premun:
Related to https://github.com/dotnet/runtime/issues/63572

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformationParser.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformationParser.cs
@@ -162,10 +162,14 @@ public class AppBundleInformationParser : IAppBundleInformationParser
             bundleExecutable: bundleExecutable);
     }
 
-    private async Task<string> GetPlistProperty(string plistPath, string propertyName, ILog log, CancellationToken cancellationToken = default, int attempt = 1)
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformationParser.cs
```diff
@@ -162,20 +162,35 @@ public async Task<AppBundleInformation> ParseFromAppBundle(string appPackagePath
             bundleExecutable: bundleExecutable);
     }
 
-    private async Task<string> GetPlistProperty(string plistPath, string propertyName, ILog log, CancellationToken cancellationToken = default)
+    private async Task<string> GetPlistProperty(string plistPath, string propertyName, ILog log, CancellationToken cancellationToken = default, int attempt = 1)
```
tkapin, line 165: Not a big issue, but passing attempt as an argument while having the max attempts defined in the method is a weird decomposition as the logic responsible for doing the retry loop is split in two parts. Either pass the max attempts together with the attempt and handle the retry logic in the caller or do it all here, but the logic should be at one place.

premun: Good point, I was caught bein lazy

akoeplinger (approved):

PR #808, opened 2022-01-31
Add `DOTNET_ROOT` instructions to `Install-XHarness.ps1`

premun:
This prevents the user from not being able to run things. Similar instructions are already in the Linux script

### DIFF
--- a/tools/Install-XHarness.ps1
+++ b/tools/Install-XHarness.ps1
@@ -28,6 +28,6 @@ Write-Host "Installing XHarness in current folder" -ForegroundColor Cyan
 ./.dotnet/dotnet tool install --tool-path . --version $xharness_version --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json Microsoft.DotNet.XHarness.CLI
 
 Write-Host "Run following command: " -ForegroundColor Cyan -NoNewline
-Write-Host '$Env:DOTNET_ROOT=".\.dotnet"' -ForegroundColor Yellow
+Write-Host "`$Env:DOTNET_ROOT='$pwd.dotnet'" -ForegroundColor Yellow
 Write-Host "Then run XHarness using: " -ForegroundColor Cyan -NoNewline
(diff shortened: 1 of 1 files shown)


tools/Install-XHarness.ps1
```diff
@@ -27,5 +27,7 @@ Write-Host "Installing XHarness in current folder" -ForegroundColor Cyan
 
 ./.dotnet/dotnet tool install --tool-path . --version $xharness_version --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json Microsoft.DotNet.XHarness.CLI
 
-Write-Host "XHarness installed, run it using: " -ForegroundColor Cyan -NoNewline
+Write-Host "Run following command: " -ForegroundColor Cyan -NoNewline
+Write-Host '$Env:DOTNET_ROOT=".\.dotnet"' -ForegroundColor Yellow
```
lpatalas, line 31: Shouldn't it suggest an absolute path? i.e. `$Env:DOTNET_ROOT="$pwd\.dotnet"`. With "dot" it won't work after you change the directory. Also this env var won't be persisted between powershell sessions.

premun: Yeah, this is meant for the session. The script installs xharness in the local folder, so you're already expected to work there (it also hints you to call it via `./xharness`.

 There is no way for us to set the variable from within the script that we're running afaik so we have to ask the user to do it.
 
 But maybe we could here put there `Get-Location`?

lpatalas: You can use `Get-Location` or built-in `$pwd` variable that stores current directory, e.g.:
```
Write-Host "`$Env:DOTNET_ROOT='$pwd\.dotnet'" -ForegroundColor Yellow
```

premun:
```suggestion
Write-Host "`$Env:DOTNET_ROOT='$pwd.dotnet'" -ForegroundColor Yellow
```

lpatalas (approved):

PR #813, opened 2022-02-03
Return `RETURN_CODE_NOT_SET` when Apple app's exit code not found

### LINKED ISSUE #812: Add a new exit code for failure of app's exit code detection
### Context

On all Apple platforms (iOS, tvOS, MacCatalyst), we have problems detecting app's exit code. The original (Xamarin) way of testing these platforms is always via the XHarness TestRunner that is included in the apps.

For all scenarios where we don't have the test runner (HelloiOS, functional tests), where we only fire up the app and wait for it to quit, XHarness is scanning MacOS syslogs for the event where the app terminated with non-zero code and parse the exit code from there. Unfortunately, this is not 100% reliable and sometimes we just don't see this line in the logs. We have no power over flushing of this log and there might be other reasons like MacOS cycling the log away when it reaches some size and starting a new one.

We have also tried to resolve this via DevWF test retries but those won't help as we have no failed tests.

More context: https://github.com/dotnet/runtime/pull/64452

### Goal

Since many tests rely on detection of a previously agreed non-zero exit code that the app should return in case of success, we shouldn't treat it as `GENERAL_FAILURE` when we fail to detect the exit code from logs but we should add a new exit code `NO_EXIT_CODE` that we can then retry and get a successful run.

premun:
Resolves #812

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/Orchestration/RunOrchestrator.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/Orchestration/RunOrchestrator.cs
@@ -239,6 +239,13 @@ public class RunOrchestrator : BaseOrchestrator, IRunOrchestrator
 
         if (exitCode is null)
         {
+            if (expectedExitCode != 0)
+            {
+                _logger.LogError("Application has finished but XHarness failed to determine its exit code! " +
+                    "This is a known issue, please run the app again.");
+                return ExitCode.RETURN_CODE_NOT_SET;
+            }
+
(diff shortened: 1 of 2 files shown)


MaximLipnin (approved):

premun:
FYI @fanyang-mono we will start retrying work items where we fail to detect the exit code (I believe runtime tests also hit this occasionally as they are based on the `run` command). No action for your team, just an FYI

lpatalas (approved):

PR #824, opened 2022-02-23
Refactor querying Android devices, allow specifying API version

### LINKED ISSUE #380: CLI for running Android tests should allow specifying API level
It would be nice to be able to specify some API level which will then provide a filter for selecting the emulator.

premun:
Resolves #380 

Previously, we were only querying devices and one of their properties - list of supported architectures. We were then matching this against user's wishes.
However, we also sometimes want to query by installed apps and now newly also by desired API version. I expect more device properties to come soon.

This means we needed to revamp a little bit how we query and represent Android devices - we no longer represent them as dictionaries but rather a dedicated POCO. We now also don't need to query device's architecture additionally, we can ask `AdbRunner` to get it for us right away.

The `android device` command also no longer expects app - you can specify other device properties (API version, arch).

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -16,23 +16,23 @@ namespace Microsoft.DotNet.XHarness.Android;
 
 public class AdbRunner
 {
-    #region Constructor and state variables
-
-    private static class AdbPropertyNames
+    private enum AdbProperty
     {
-        public const string BootCompletion = "sys.boot_completed";
-        public const string SupportedArchitectures = "ro.product.cpu.abilist";
-        public const string ApiVersion = "ro.build.version.sdk";
-        public const string Architecture = "ro.product.cpu.abi";
-        public const string InstalledApps = "apps";
+        Architecture,
+        ApiVersion,
+        SupportedArchitectures,
+        InstalledApps,
+        BootCompletion,
     }
 
-    private static readonly Dictionary<string, string[]> s_commandList = new()
+    #region Constructor and state variables
+
+    private static readonly Dictionary<AdbProperty, string[]> s_commandList = new()
     {
-        { AdbPropertyNames.SupportedArchitectures, new[] { "shell", "getprop", AdbPropertyNames.SupportedArchitectures } },
-        { AdbPropertyNames.ApiVersion, new[] { "shell", "getprop", AdbPropertyNames.ApiVersion } },
-        { AdbPropertyNames.Architecture, new[] { "shell", "getprop", AdbPropertyNames.Architecture } },
-        { AdbPropertyNames.InstalledApps, new[] { "shell", "pm", "list", "packages", "-3" } },
+        { AdbProperty.SupportedArchitectures, new[] { "shell", "getprop", "ro.product.cpu.abilist" } },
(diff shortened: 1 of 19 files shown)


premun:
cc @mattleibow

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -74,19 +86,31 @@ public AdbRunner(ILogger log, IAdbProcessManager processManager, string adbExePa
...
     }
 
-    private int GetAPIVersion()
+    private int GetApiVersion()
     {
-        var output = RunAdbCommand(new[] { "shell", "getprop", "ro.build.version.sdk" });
-        return int.Parse(output.StandardOutput);
+        var output = RunAdbCommand(s_commandList[AdbPropertyNames.ApiVersion]);
+        var apiVersion = int.Parse(output.StandardOutput);
+        _activeDevice = (_activeDevice ?? new AndroidDevice(null!)) with
```
lpatalas, line 108: Isn't `null!` here a mistake? I'm surprised that compiler allows this. It looks like an oxymoron. :)

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -424,19 +448,27 @@ public List<string> PullFiles(string apkPackageName, string devicePath, string l
         }
     }
 
-    public Dictionary<string, string?> GetAttachedDevicesWithProperties(string property)
+    public List<AndroidDevice> GetAttachedDevicesWithProperties()
```
lpatalas, line 451: Couldn't it be just called `GetAttachedDevices()`? Also I think it's a good practice to use immutable collections (like `IReadOnlyList<AndroidDevice>`) where possible unless we expect the called to modify the result. I'm not sure if it applies in this case.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -627,11 +792,29 @@ public ProcessExecutionResults RunApkInstrumentation(string apkName, string? ins
...
+
+            SupportedArchitectures = properties.Value.TryGetValue(AdbPropertyNames.SupportedArchitectures, out string? archs) && archs is not null
+                ? archs.Split(new char[] { ',', '\r', '\n' })
+                : null,
+        };
+
     #endregion
 
-    #region Process runner helpers
+        #region Process runner helpers
```
lpatalas, line 815: Unnecessary indentation
```suggestion
    #region Process runner helpers

```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
premun: Yeah, this is suspicious. Probably something I didn't fix.
(I wrote the code 2 weeks ago and only found it in the branch today 😅)

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
premun: Yup, makes sense.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
premun: I found out I could simplify things a lot again thanks to this.

Issue #819, opened 2022-02-09
Exit code detection failing on iOS/tvOS 15+


premun:
With the latest move of `dotnet/runtime` onto OSX 12.00, `apple run` and `apple just-run` commands started failing in 100% of cases:

![image](https://user-images.githubusercontent.com/7013027/153199265-413daa78-e834-4b01-85c8-51b205c654a9.png)

```sql
Metrics
| where EventType == "_MobileDeviceOperation" and MetricName == "ExitCode"
| join kind=inner Jobs on JobId
| where Finished > now()-1d
| extend Dimensions = parse_json(Dimensions)
| project 
    Finished,
    JobName,
    QueueName,
    WorkItemFriendlyName,
    Platform = tostring(Dimensions.platform),
    Command = tostring(Dimensions.command),
    ExitCode = MetricValue,
    Target = tostring(Dimensions.target),
    IsDevice = tobool(Dimensions.isDevice)
| where Platform == "apple"
| summarize SuccessRate=countif(ExitCode < 6)*100/count(), Count=count() by Platform, Command, QueueName
| where SuccessRate != 100
| order by Platform, SuccessRate, Command
```

premun:
Some reading for later
- https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/kqueue.2.html
- https://stackoverflow.com/a/24690619/6289970
- https://developer.apple.com/library/archive/technotes/tn2050/_index.html

premun:
Seems like the issue is Xcode 13.2 rather than OSX 12.00.

Also I found out, I might be able to use `kqueue` and `kevent` to get notified about a process' exit where the exit code should be in the data of the event:
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/kevent.2.html

```
EVFILT_PROC  Takes the process ID to monitor as the	identifier and
			 the events to watch for in fflags, and	returns	when
			 the process performs one or more of the requested
			 events.  If a process can normally see	another
			 process, it can attach	an event to it.	 The events to
			 monitor are:

			 NOTE_EXIT	   The process has exited. The exit status will be stored in data.
```

premun:
Oh noez, the sentence _"The exit status will be stored in data."_ actually only appears in FreeBSD man pages and not OSX man pages 😢

premun:
I was able to get more logs via:
```
[...]

Issue #694, opened 2021-08-20
App logs are empty on Helix Apple simulators


premun:
### Context

When we are running the Simulator app, we tell it to log the app's output into `net.dot.[APP ID].log`:

https://github.com/dotnet/xharness/blob/e9669dc84ecd668d3bbb748758103e23b394ffef/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs#L556-L558

### Problem

The file seems to be empty when run on Helix but it seems to have the output when run locally.

Might be related to this - https://github.com/xamarin/xamarin-macios/pull/12493

premun:
I have disabled a machine `dci-mac-build-056` to perform these tests on

premun:
I have re-enabled it again, will try on one of the production machines

premun:
Finished solution for this is here: https://github.com/premun/xharness/tree/prvysoky/log-stream

PR #836, opened 2022-03-17
Pull `AppBundleInfo` from Simulator if possible

premun:
### Context
For some commands, we receive path to the app bundle:
```
xharness apple test -a path/to/some.app ...
```

We parse its `Info.plist` and get full `AppBundleInformation` object.
However, for some flows, we only have `bundleIdentifier` available, e.g.:

```
xharness apple just-run -a net.dot.System.Text.Json.Tests ...
```

For the latter case, we need to try to resolve the full `AppBundleInformation` object. We will for example need the `BundleExecutable` property later when reading Simulator logs.

### Change
This PR makes it possible to query the Simulator by the `bundleIdentifier` and resolve the path where the app is installed. Then we can parse the `Info.plist` in that location and load full `AppBundleInformation`.

It is not ideal that we can only do this for Simulators but we actually only need this for simulators.

Related issues: #819 #694


src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformation.cs
```diff
@@ -35,5 +35,5 @@ public class AppBundleInformation
     }
 
     public static AppBundleInformation FromBundleId(string bundleIdentifier) =>
-        new(bundleIdentifier, bundleIdentifier, string.Empty, string.Empty, false);
+        new(bundleIdentifier, bundleIdentifier, string.Empty, string.Empty, false, bundleExecutable: bundleIdentifier);
```
dkurepa, line 38: Why are we setting properties like appName to be the same string as bundleIdentifier?

src/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/SimulatorDevice.cs
```diff
@@ -201,4 +201,28 @@ public async Task<bool> Boot(ILog log, CancellationToken cancellationToken)
...
+            "simctl",
+            new[] { "get_app_container", UDID, bundleIdentifier },
+            log,
+            output,
+            output,
+            TimeSpan.FromSeconds(30));
+
+        if (!result.Succeeded)
+        {
+            throw new Exception($"Failed to get information for '{bundleIdentifier}'. Please check the app is installed");
```
dkurepa, line 220: Maybe we should add that the app failed to get the information in 30 seconds, in case it was a time out?

src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformation.cs
premun: Good question. I will add a comment here explaining that

src/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/SimulatorDevice.cs
premun: The logs from the `_processManager` call above will actually have that info if people want to dig into the real reason (from a high-level  end user standpoint we don't care much)

dkurepa (approved):

PR #842, opened 2022-03-29
Capture app logs on Simulators via `log stream`

premun:
On iOS 15+ the logs coming from Simulator no longer contain much so we are switching to the `log stream` command that is spawned on the Simulator via `xcrun simctl` and captures the logs better


src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppRunnerBase.cs
```diff
@@ -106,6 +109,65 @@ public abstract class AppRunnerBase
...
+
+            companionLog.StartCapture();
+            _logs.Add(companionLog);
+            systemLogs.Add(companionLog);
+
+            var companionScanToken = await CaptureSimulatorLog(companionSimulator, appInformation, cancellationToken);
+            if (companionScanToken != null)
+            {
+                simulatorScanToken = CancellationTokenSource.CreateLinkedTokenSource(
+                cancellationToken,
```
dkurepa, line 157: Indentation

premun: Fixed

premun:
@fanyang-mono @steveisok fyi - this change finally makes the stdout of apps show in a file in Helix

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppRunnerBase.cs
```diff
@@ -114,17 +176,17 @@ public abstract class AppRunnerBase
     protected void AddExtraEnvVars(Dictionary<string, string> envVariables, IEnumerable<(string, string)> variables)
     {
         using (var enumerator = variables.GetEnumerator())
```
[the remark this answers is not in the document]
lpatalas: Maybe it would be better to add explicit brackets `{ .. }` around the using block? This way it would be clearer.

src/Microsoft.DotNet.XHarness.iOS.Shared/AppBundleInformationParser.cs
```diff
@@ -136,16 +136,13 @@ public async Task<AppBundleInformation> ParseFromAppBundle(string appPackagePath
...
-            }
-            catch
-            {
-                log.WriteLine("Failed to locate the bundle executable property in Info.plist");
-            }
+            bundleExecutable = await GetPlistProperty(plistPath, PListExtensions.BundleExecutablePropertyName, log, cancellationToken);
+        }
+        catch
+        {
+            log.WriteLine("Failed to locate the bundle executable property in Info.plist");
```
lpatalas, line 145: Shouldn't we log caught exception to the output?

lpatalas (approved):

dkurepa (approved):

PR #844, opened 2022-03-29
Add `apple mlaunch` command

### LINKED ISSUE #843: Add `apple mlaunch` command (same as `android adb`)
Similarly to the `android adb` command that allows you to call the bundled ADB easily, we need the same for calling `mlaunch`

premun:
Resolves #843


lpatalas (approved):
Looks good

PR #862, opened 2022-04-14
Display version when XHarness command is invoked

premun:


src/Microsoft.DotNet.XHarness.CLI/Program.cs
```diff
@@ -28,7 +30,8 @@ public static int Main(string[] args)
 
         if (shouldOutput)
         {
-            Console.WriteLine($"XHarness command issued: {string.Join(' ', args)}");
+            var version = FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location).ProductVersion;
```
tkapin, line 33: Since we are getting this string in the --version argument as well, can we refactor this call to a function so the logic is at one place?

premun: Good point, done

tkapin (approved):

PR #864, opened 2022-04-15
Initial implementation for linux-bionic (headless Android)

directhex:
We have an internal customer asking us for the ability to target Android as if it were a Linux distro with a weirdo libc, which to be fair it basically is. However, to run tests on there, we need a lot of the existing Android infra provided by XHarness (emulator starting/restarting/pushing/pulling). Hence this PR.

This is definitely quite cursed for now, but appears to do the right thing at least locally & with direct pushes to Helix (outstanding problems w/ test results not being collected appear to be Helix SDK problems, not XHarness problems)


premun (commented):
It seems like you have copied a lot of code which looks like is identical for `Android` and `AndroidHeadless`? Is it possible to re-use code as much as possible?

I understand some of the logic might differ but for example things like `Arguments` are designed to be re-usable as much as possible so that the help strings don't differ and so on.
I see many things being copied over such as the `IDiagnosticsDataExtensions` and more but if possible, it would be nice to also re-use as much of the Commands too.

I think it would be quite hard to start maintaining 2 separate Android code bases that are quite identical - mostly because we also don't have many tests for the Android namespace.

directhex:
> It seems like you have copied a lot of code which looks like is identical for `Android` and `AndroidHeadless`?

It seems that way for a reason :sweat_smile: 

I can work on reconciling the code. While I was heavily iterating on it, I wanted to work in isolation without breaking the known-good Android code.

directhex:
Okay, that's 500 lines chopped out & it still compiles. Will have another pass after lunch.

premun:
Cool! I will have a look tomorrow.

Once you get a working scenario in runtime, it would also be cool to add an E2E tests here so that there's coverage (they are easier to do and have more value for Android)

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -311,6 +379,40 @@ public int InstallApk(string apkPath)
...
+        var result = RunAdbCommand(new[] { "shell", "rm", "-fr", fullTestPath });
+
+        // See note above in install()
+        if (result.ExitCode == (int)AdbExitCodes.ADB_BROKEN_PIPE)
+        {
+            _log.LogWarning($"Hit broken pipe error; Will make one attempt to restart ADB server, and retry the uninstallation");
+
+            KillAdbServer();
+            StartAdbServer();
+            result = RunAdbCommand(new[] { "rm", "-fr", fullTestPath });
```
premun, line 401: No `shell` here?

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -311,6 +379,40 @@ public int InstallApk(string apkPath)
...
+        }
+
+        if (result.ExitCode == (int)AdbExitCodes.SUCCESS)
+        {
+            _log.LogInformation($"Successfully uninstalled {fullTestPath}");
+        }
+        else
+        {
+            _log.LogError(message: $"Error: {result}");
```
premun, line 410:
```suggestion
            _log.LogError(message: $"Failed to uninstall {fullTestPath}: {result}");

```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -364,6 +466,21 @@ public int KillApk(string apkName)
         return result.ExitCode;
     }
 
+    public int KillProcess(string testName)
+    {
+        _log.LogInformation($"Killing all running processes for '{testName}': ");
+        var result = RunAdbCommand(new[] { "shell", "pkill", testName });
+        if (result.ExitCode != (int)AdbExitCodes.SUCCESS)
+        {
+            _log.LogError($"Error:{Environment.NewLine}{result}");
```
premun, line 475:
```suggestion
            _log.LogError($"Failed to kill process by name ({testName}):{Environment.NewLine}{result}");

```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -364,6 +466,21 @@ public int KillApk(string apkName)
...
+    {
+        _log.LogInformation($"Killing all running processes for '{testName}': ");
+        var result = RunAdbCommand(new[] { "shell", "pkill", testName });
+        if (result.ExitCode != (int)AdbExitCodes.SUCCESS)
+        {
+            _log.LogError($"Error:{Environment.NewLine}{result}");
+        }
+        else
+        {
+            _log.LogDebug($"Success!{Environment.NewLine}{result.StandardOutput}");
```
premun, line 479:
```suggestion
            _log.LogDebug($"Process {testName} killed!{Environment.NewLine}{result.StandardOutput}");

```

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -789,6 +980,38 @@ public void SetActiveDevice(AndroidDevice? device)
...
+        var localRuntimePath = Path.Combine(GlobalReadWriteDirectory, "runtime", "dotnet");
+        var adbArgs = new List<string>
+        {
+            "shell",
+        };
+
+        _log.LogInformation($"Starting {testScript} from {localTestPath} (exit code 0 == success)");
+        adbArgs.Add(localTestPath);
+        adbArgs.Add("-r");
+        adbArgs.Add(localRuntimePath);
```
premun, line 995:
```suggestion
        var adbArgs = new List<string>
        {
            "shell",
            localTestPath,
            "-r",
            localRuntimePath,
        };

        _log.LogInformation($"Starting {testScript} from {localTestPath} (exit code 0 == success)");

```

premun (commented):
Just a couple of comments

premun (approved):

PR #877, opened 2022-05-05
Notify user when apk and required architectures differ

### LINKED ISSUE #481: Notify user if required arch differs from the app's arch
When required arch differs from the app's arch, user's parameters have higher priority. We will still try to run the app on the required architecture. But since it's easy to detect we may print warning about the difference.

### Goal:
Let user know that the app's architecture differs from the one they required

premun:
Resolves #481

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidInstallCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidInstallCommand.cs
@@ -59,7 +59,7 @@ Arguments:
             logger: logger,
             apkPackageName: Arguments.PackageName,
             appPackagePath: Arguments.AppPackagePath,
-            apkRequiredArchitecture: apkRequiredArchitecture,
+            requiredArchitecture: Arguments.DeviceArchitecture.Value.ToList(),
             deviceId: Arguments.DeviceId,
             apiVersion: Arguments.ApiVersion.Value,
             bootTimeoutSeconds: Arguments.LaunchTimeout,
@@ -71,7 +71,7 @@ Arguments:
         ILogger logger,
         string apkPackageName,
         string appPackagePath,
-        IEnumerable<string> apkRequiredArchitecture,
+        IReadOnlyCollection<string> requiredArchitecture,
         string? deviceId,
         int? apiVersion,
         TimeSpan bootTimeoutSeconds,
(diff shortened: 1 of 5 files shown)


premun:
Tested here: https://helixre8s23ayyeko0k025g8.blob.core.windows.net/arcade-test-f61a431e5de145e9b9/System.Numerics.Vectors.Tests-x86/1/console.4f8e1325.log?helixlogtype=result

```
[1.0.0-dev] XHarness command issued: android install --device-id=emulator-5554 --output-directory=/datadisks/disk1/work/B18F097F/w/B38709A2/uploads --package-name=net.dot.System.Numerics.Vectors.Tests --app=System.Numerics.Vectors.Tests-x86.apk --device-arch=arm64-v8a
[09:15:26] fail: The APK at /datadisks/disk1/work/B18F097F/w/B38709A2/e/System.Numerics.Vectors.Tests-x86.apk supports x86 architectures which does not match any of the specified architectures (arm64-v8a)
XHarness exit code: 3 (INVALID_ARGUMENTS)
```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidInstallCommand.cs
```diff
@@ -71,7 +71,7 @@ protected override ExitCode InvokeCommand(ILogger logger)
         ILogger logger,
         string apkPackageName,
         string appPackagePath,
-        IEnumerable<string> apkRequiredArchitecture,
+        IReadOnlyCollection<string> requiredArchitecture,
```
lpatalas, line 74: Shouldn't it be called `requiredArchitectures`?

In general I found the naming a bit confusing. There is `requiredArchitecture`, `apkRequiredArchitecture` and `apkSupportedArchitecture` which all seem similar. If I understand the logic correctly I would propose to rename them to something like:
- `requiredArchitecture` -> `requestedArchitectures`
- `apkRequiredArchitecture` -> `selectedArchitectures` / `actualArchitectures` or something like this
- `apkSupportedArchitectures` - this one is ok, `supportedArchitectures` would be fine as well

You can also extract the part where you determine `apkRequiredArchitecture` to separate method which would hide one of those variables and simplify that part to just (using proposed naming):
```
var selectedArchitectures = SelectAvailableArchitectures(requestedArchitectures, appPackahePath);
```

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidInstallCommand.cs
```diff
@@ -80,6 +80,27 @@ protected override ExitCode InvokeCommand(ILogger logger)
...
         using (logger.BeginScope("Initialization and setup of APK on device"))
         {
+            IEnumerable<string> apkRequiredArchitecture;
+            var apkSupportedArchitectures = ApkHelper.GetApkSupportedArchitectures(appPackagePath);
+
+            if (requiredArchitecture.Any())
+            {
+                if (!apkSupportedArchitectures.Intersect(requiredArchitecture).Any())
+                {
+                    logger.LogError($"The APK at {appPackagePath} supports {string.Join(", ", apkSupportedArchitectures)} architectures " +
```
lpatalas, line 90: Shouldn't you use structured logging for new code?

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidInstallCommand.cs
premun: Yeah, this naming was confusing from the start. You're right, I can change it probably a bit more since I am tending to this already

premun: I didn't end up putting it into a method because I also need to return the exit code in case of a failure (and also pass in the logger) so it was becoming too much of a hassle

PR #882, opened 2022-05-09
Add E2E tests for the scouting queue

premun:
This test project will be used when testing new Xcode. It is linked to [from the wiki](https://dev.azure.com/dnceng/internal/_wiki/wikis/DNCEng%20Services%20Wiki/870/Install-new-version-of-Xcode-on-Helix-machines)

### DIFF
--- a/tests/integration-tests/Apple/Simulator.Scouting.Tests.proj
+++ b/tests/integration-tests/Apple/Simulator.Scouting.Tests.proj
@@ -1,0 +1,34 @@
+<Project DefaultTargets="Test">
+  <Import Project="../Helix.SDK.configuration.props"/>
+
+  <ItemGroup>
+    <HelixTargetQueue Include="osx.1200.amd64.scouting.open"/>
+
+    <!-- apple test / ios-simulator-64 -->
+    <XHarnessAppleProject Include="$(MSBuildThisFileDirectory)\TestAppBundle.proj">
+      <AdditionalProperties>TestTarget=ios-simulator-64;TestAppBundleName=System.Numerics.Vectors.Tests.app</AdditionalProperties>
+    </XHarnessAppleProject>
+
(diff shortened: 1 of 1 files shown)


tests/integration-tests/Apple/Simulator.Scouting.Tests.proj
```diff
@@ -0,0 +1,34 @@
...
+    </XHarnessAppleProject>
+
+    <!-- apple run / maccatalyst -->
+    <XHarnessAppleProject Include="$(MSBuildThisFileDirectory)\TestAppBundle.proj">
+      <AdditionalProperties>TestTarget=maccatalyst;TestAppBundleName=iOS.Simulator.PInvoke.Test.app;IncludesTestRunner=false;ExpectedExitCode=42</AdditionalProperties>
+    </XHarnessAppleProject>
+
+    <!-- apple test / tvos-simulator -->
+    <XHarnessAppleProject Include="$(MSBuildThisFileDirectory)\TestAppBundle.proj">
+      <AdditionalProperties>TestTarget=tvos-simulator;TestAppBundleName=Microsoft.Extensions.Configuration.Ini.Tests.app</AdditionalProperties>
```
tkapin, line 29: LGTM, just a question - the <XHarnessAppleProject/> node differs only in the AdditionalProperties, otherwise it's the same (the Include attribute). It's not a big deal in this change, but is it possible to write this more concisely?

tkapin (approved):

tests/integration-tests/Apple/Simulator.Scouting.Tests.proj
premun: I don't think there is a better way to make this even more concise - we have to call the `TestAppBundle.proj` with different parameters

PR #893, opened 2022-05-19
Include XHarness version in diagnostic data

premun:


premun:
I will just pull it from the `xharness version` command

PR #894, opened 2022-05-20
Print version only for the version command

premun:


src/Microsoft.DotNet.XHarness.CLI/Commands/XHarnessVersionCommand.cs
```diff
@@ -12,14 +13,21 @@ internal class XHarnessVersionCommand : Command
...
+            // Print the name of the tool and the version number unix style
+            // Example:
+            // Apple clang version 11.0.3 (clang-1103.0.32.29)
+            // Target: x86_64-apple-darwin19.4.0
+            // InstalledDir: /Applications/Xcode114.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
+            Console.WriteLine($"XHarness version {version.ProductVersion} ({version.OriginalFilename})");
+            Console.WriteLine($"InstalledDir: {version.FileName}");
+        }
+
+        Console.WriteLine(version.ProductVersion);
```
andriipatsula, line 29: if Line :18 `true` we'll print `ProductVersion` 2 times: line :29, line :25

andriipatsula (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/XHarnessVersionCommand.cs
premun: Right, I'm stupid. Thanks!

PR #916, opened 2022-06-29
Give access to `kTCCServiceCamera` in simulator's TCC database

rolfbjarne:
This isn't a new permission, but for some reason it seems this becomes a problem in iOS 16 / Xcode 14 if it's not set (when it worked previously).


premun (approved):

premun:
@rolfbjarne does this mean we will have to reset all simulators before using them the first time?

rolfbjarne:
@premun unless you have tests that require camera access, I don't think you need to do anything (and even then probably not, since it should be automatic in xharness already)

PR #948, opened 2022-09-13
Fix SpecialFolder.Personal usage

eerhardt:
SpecialFolder Personal is the same value as MyDocuments. I am proposing to obsolete/deprecate SpecialFolder.Personal in https://github.com/dotnet/runtime/issues/75563. Since these two names have the same underlying value, it is more understandable/readable to use MyDocuments in code. This way readers fully understand what folder is being referenced.

Changing Personal to MyDocuments here.


premun (approved):

premun:
The failure here is unrelated, merging

PR #961, opened 2022-10-04
Add CodeQL pipeline

garath:
Add a CodeQL pipeline along the same lines as our existing pipelines. 

The new pipeline is has been designated [1184](https://dev.azure.com/dnceng/internal/_build?definitionId=1184&_a=summary), which, in my opinion, is a really great number. 

[Build 2010587](https://dev.azure.com/dnceng/internal/_build/results?buildId=2010587&view=results) is a full, successful run against my feature branch.

### DIFF
--- a/azure-pipelines-codeql.yml
+++ b/azure-pipelines-codeql.yml
@@ -1,0 +1,104 @@
+variables:
+  - name: _TeamName
+    value: DotNetCore
+  - group: SDL_Settings
+  - template: eng/common-variables.yml
+  - name: Build.Repository.Clean
+    value: true
+
+trigger: none
+
+schedules:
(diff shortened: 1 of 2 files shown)


premun (approved):

premun:
> The new pipeline is has been designated [1184](https://dev.azure.com/dnceng/internal/_build?definitionId=1184&_a=summary), which, in my opinion, is a really great number.

😂😂

@garath just one question - shouldn't code like this come from Arcade? E.g. the `ps1` file?

garath:
> shouldn't code like this come from Arcade? E.g. the ps1 file?

Turns out there is a new thing! So I switched to use that.

MattGal (approved):

PR #968, opened 2022-10-21
Arcade update + use new pool-providers.yml to pick the build pool 

MattGal:
This change updates Arcade for this branch and introduces the ability to automatically pick pool provider based on branch.  See files changed outside of the eng/common folder for what changed.

Not sure why there are such a large number of whitespace changes but this should go away in future arcade updates.

Test runs:

Main CI: https://dnceng.visualstudio.com/internal/_build/results?buildId=2026946&view=logs&s=e6b610b4-9a0b-53ae-ef5f-811cf6e05977
CodeQL:  https://dnceng.visualstudio.com/internal/_build/results?buildId=2026949&view=results


azure-pipelines-codeql.yml
```diff
@@ -38,7 +39,7 @@ jobs:
 - job: CodeQL
   timeoutInMinutes: 90
   pool: 
-    name: NetCore1ESPool-Internal
+    name: $(DncEngInternalBuildPool)
```
premun, line 42: Nice! I will use this in my new stuff I am writing for `dotnet/installer`

akoeplinger:
All of the line endings got changed to CRLF, probably because you changed it on a Windows machine. I don't think we should've merged this as is since it messes up history a bit :)

premun:
Ah, yeah, we shouldn't have. I thought this was Arcade giving us the files like that so I approved.

akoeplinger:
No big deal, we can leave it as is.

MattGal:
> No big deal, we can leave it as is.

I knew something was off but I couldn't figure it out.  I have git set as "checkout as-is / commit as-is"  and this has only happened on the Xharness repo.  Future dependency flows will overwrite it back, if they do.  Sorry!

Issue #937, opened 2022-08-10
Disable chatty log trimming on Android


akoeplinger:
We see quite often in the logcat something like this:

```
chatty  : uid=10225(net.dot.System.Net.Sockets.Tests) expire 102 lines
```

The chatty service removes log messages if a process logs too many messages in some time period. This is exactly what our tests do though since we print a line for each passed/failed test 😄 

We should disable this behavior by running `logcat -P ""` before starting the test run.

PR #972, opened 2022-11-17
Disable chatty on Android runs

steveisok:
Chatty is a service which tries to limit repeating entries in logcat output. This impacts us negatively more often than not as it will hide important entries that show what went wrong in a test.  This tries to disable chatty for all processes on the device.

Fixes https://github.com/dotnet/xharness/issues/937

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -130,7 +130,7 @@ public class AdbRunner
     {
         RunAdbCommand("logcat", "-b", "all", "-c");
 
-        // Android logs can unnecessarily hide log entires, so disable
+        // Android logs can unnecessarily hide log entries, so disable
         DisableChatty();
     }
 
@@ -1075,7 +1075,7 @@ public class AdbRunner
 
         if (!result.Succeeded)
         {
-            _log.LogWarning($"Unable to disable chatty. Logcat may hide what it finds to be repeating entires.");
+            _log.LogWarning($"Unable to disable chatty. Logcat may hide what it finds to be repeating entries.");
         }
     }
 
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -126,7 +126,7 @@ public class AdbRunner
         return result.StandardOutput;
     }
 
-    public void ClearAdbLog() => RunAdbCommand("logcat", "-c");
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -251,6 +251,9 @@ public void StartAdbServer()
         {
             throw new AdbFailureException("Failed to start the ADB server");
         }
+
+        // Android logs can unnecessarily hide log entires, so disable
+        DisableChatty();
```
akoeplinger, line 256: I don't think we should do this in `StartAdbServer()` since we don't know whether the device is connected at that point.

I'd move it directly into `ClearAdbLog()` instead.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -1063,6 +1066,16 @@ public ProcessExecutionResults RunApkInstrumentation(string apkName, string? ins
         return result;
     }
 
+    private void DisableChatty()
+    {
+        var result = RunAdbCommand(new[] { "shell", "logcat", "-P", "'\"\"'" }, TimeSpan.FromMinutes(1));
```
akoeplinger, line 1071: is the double-quotes-in-single-quotes really necessary?
you also don't need `shell` since `logcat` is (also) a comment on `adb`.

src/Microsoft.DotNet.XHarness.CLI/Commands/Android/AndroidTestCommand.cs
```diff
@@ -39,6 +39,8 @@ protected override ExitCode InvokeCommand(ILogger logger)
     {
         var runner = new AdbRunner(logger);
 
+        runner.ClearAdbLog();
```
akoeplinger, line 42: move this after `AndroidInstallCommand` so we can be sure the device is up and connected.

akoeplinger (changes requested):
Looks good, left a few comments and you also have a test failure.

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
steveisok: Yeah, it fails otherwise

akoeplinger: did you try without the `shell`? I think it might be because of that

steveisok: Yeah, it behaves the same way.

akoeplinger (approved):

PR #977, opened 2022-11-29
Add the WASI runtime test target

mkhamoyan:
Add WASI runtime as XHarness target


README.md
```diff
@@ -4,7 +4,7 @@ This repo contains the code to build the **XHarness dotnet tool** and a **TestRu
 
 ## What is XHarness
 
-XHarness is primarily a command line tool that enables running xUnit like tests on Android, Apple iOS / tvOS / WatchOS / Mac Catalyst and desktop browsers (WASM).
+XHarness is primarily a command line tool that enables running xUnit like tests on Android, Apple iOS / tvOS / WatchOS / Mac Catalyst and desktop browsers (WASM,WASI).
```
pavelsavara, line 7: WASI is not a browser, please put it in the list with the other OS

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASI/Arguments/ErrorPatternsFileArgument.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System;
+using System.IO;
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasi;
+
+internal class ErrorPatternsFileArgument : StringArgument
+{
+    public ErrorPatternsFileArgument()
+        : base("error-patterns=|p=", "File containing error patterns. Each line prefixed with '@', or '%' for a simple string, or a .net regex, respectively")
```
pavelsavara, line 13: could you please explain what does it do and why we need it ?

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASI/Arguments/LibraryArgument.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasi;
+
+internal class LibraryArgument : RequiredStringArgument
+{
+    public LibraryArgument(string defaultValue)
```
radical, line 9: This shouldn't be a separate argument. Instead, it can passed by the user as part of the arguments.
One reason is that the main dll name is a something specific to the .wasm file that we are running, and it might not be actually needed in some cases. For example, when we bundle all the assemblies, and dependencies in a single `foo.wasm` file then it has the name of the main assembly embedded, and does not need this as an argument.

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASI/Arguments/DirectoryArgument.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasi;
+
+internal class DirectoryArgument : RequiredStringArgument
```
radical, line 7: Instead of this we should use 

https://github.com/dotnet/xharness/blob/1e1eab570c2177aa9dfb2de2826dfa28aa456f8c/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/WasmTestBrowserCommandArguments.cs#L13-14

https://github.com/dotnet/xharness/blob/1e1eab570c2177aa9dfb2de2826dfa28aa456f8c/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/AppPathArgument.cs#L10-L15

- That argument is for specifying the directory that contains the application, and that's what we are trying to specify with `wasmtime --dir` too.
- Also, from the user pov it will be consistent across wasm/wasi/others.

tests/Microsoft.DotNet.XHarness.CLI.Tests/Commands/XHarnessCommandTests.cs
```diff
@@ -1,20 +1,21 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-using System;
-using System.Collections.Generic;
-using System.Linq;
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
radical, line 1: The line endings change to CRLF should be dropped.

radical (changes requested):
Good first draft 👍

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASI/WasiTestCommandArguments.cs
```diff
@@ -0,0 +1,32 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasi;
+
+internal class WasiTestCommandArguments : XHarnessCommandArguments
+{
+    public AppPathArgument AppPackagePath { get; } = new();
```
radical, line 12: Looking at this again, we don't need this `AppPathArgument` at all. This is because:

1. we could be running self-contained .wasm module which doesn't use the FS at all, in which case this isn't needed. 2. Also, `wasmtime`'s `--dir` argument can be used multiple times, and does not have any concept of "app directory".

```
        --dir <DIRECTORY>
            Grant access to the given host directory
```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASI/WasiTestCommandArguments.cs
```diff
@@ -0,0 +1,32 @@
...
+    public WasmEngineArgument Engine { get; } = new();
+    public WasmEngineLocationArgument EnginePath { get; } = new();
+    public WasmEngineArguments EngineArgs { get; } = new();
+    public WasmFileArgument WasmFile { get; } = new("dotnet.wasm");
```
radical, line 16: This is not needed. It can be specified as part of the pass through commands. For example:

`xharness wasi test -- WasmTestRunner.wasm System.Collections.Tests.dll`

All the arguments after `--` are pass-through arguments, which will be passed to `wasmtime` as-is.

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASI/WasiTestCommandArguments.cs
```diff
@@ -0,0 +1,32 @@
...
+    public AppPathArgument AppPackagePath { get; } = new();
+    public WasmEngineArgument Engine { get; } = new();
+    public WasmEngineLocationArgument EnginePath { get; } = new();
+    public WasmEngineArguments EngineArgs { get; } = new();
+    public WasmFileArgument WasmFile { get; } = new("dotnet.wasm");  
+    public ExpectedExitCodeArgument ExpectedExitCode { get; } = new((int)Common.CLI.ExitCode.SUCCESS);
+    public OutputDirectoryArgument OutputDirectory { get; } = new();
+    public TimeoutArgument Timeout { get; } = new(TimeSpan.FromMinutes(15));
+
+    public string SubCommand{ get; } = "run";
```
radical, line 21: This is not needed. For *running* the tests we will use `run` by default anyway.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+        await PrintVersionAsync(Arguments.Engine.Value.Value, engineBinary);
+
+        var cts = new CancellationTokenSource();
+        try
+        {
+            var engineArgs = new List<string>();
+            engineArgs.Add(Arguments.SubCommand);
+            engineArgs.Add("--dir");
+            engineArgs.Add(".");
+            engineArgs.Add(Arguments.WasmFile);
```
radical, line 62: These can be dropped, as they would be passed in by the user as part of the command line. If the user wants to pass any `--dir` arguments then that can be done like:

`xharness wasi test --engine-arg=--dir=/foo/bar -- ...`

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+            engineArgs.AddRange(Arguments.EngineArgs.Value);
+            engineArgs.AddRange(PassThroughArguments);
+
+            var xmlResultsFilePath = Path.Combine(Arguments.OutputDirectory, "testResults.xml");
+            File.Delete(xmlResultsFilePath);
+
+            var stdoutFilePath = Path.Combine(Arguments.OutputDirectory, "wasi-console.log");
+            File.Delete(stdoutFilePath);
```
radical, line 70: This also needs the `WasmTestMessagesProcessor` to be used. It will copy all the output to this file, and only show the failures on the console. The symbolicator arguments can be null.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            var processTask = processManager.ExecuteCommandAsync(
+                engineBinary,
+                engineArgs,
+                log: new CallbackLog(m => logger.LogInformation(m)),
```
radical, line 75: I tried running a test with the command and found that `m` already had a newline at the end, and `LogInformation` would add another unnecessary one. So, use `logger.LogInformation(m.Trim())`.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+            if (!cts.IsCancellationRequested)
+            {
+                cts.Cancel();
+            }
+        }
+
+        Task PrintVersionAsync(WasmEngine engine, string engineBinary)
+        {
+            return Task.CompletedTask;
+        }
```
radical, line 129: The implementation from wasm test command can be used here too, just with `--version` as the argument.

radical (changes requested):
This is looking much better now! Some more feedback based on trying out the new command.

radical:
It would be good to use `FindEngineInPath` in the test command (https://github.com/dotnet/xharness/pull/977#discussion_r1037413824).

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
```diff
@@ -0,0 +1,162 @@
...
+        else
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {                
+                engineBinary = FileUtils.FindFileInPath(engineBinary + ".cmd");
+            }
+        }
+      
+        logger.LogInformation($"Using wasm engine {Arguments.Engine.Value} from path {engineBinary}");
+        await PrintVersionAsync(Arguments.Engine.Value.Value, engineBinary);
```
radical, line 61: Move these inside the try/catch, so a missing engine error can be reported correctly. Do the same for `WasmTestCommand`.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
```diff
@@ -0,0 +1,162 @@
...
+                return ExitCode.GENERAL_FAILURE;
+            }
+            else
+            {
+                if (logProcessor.LineThatMatchedErrorPattern != null)
+                {
+                    logger.LogError("Application exited with the expected exit code: {result.ExitCode}."
+                                    + $" But found a line matching an error pattern: {logProcessor.LineThatMatchedErrorPattern}");
+                    return ExitCode.APP_CRASH;
+                }
```
radical, line 131: This is not needed because we are not using a error patterns scanner at all.

radical (commented):
Other than the few comments, looks good 👍

radical (approved):
LGTM 👍 great job:)

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASI/WasiTestCommandArguments.cs
pavelsavara: maybe we could try to add https://wasmedge.org/ engine to see if they have same expectations

pavelsavara (approved):
thanks!

PR #986, opened 2023-01-16
Bump to Net8

pavelsavara:
to make xharness work with [plugins in the runtime repo](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/tests/System/Net/Prerequisites/NetCoreServer/NetCoreServer.csproj) which are now Net8

### DIFF
--- a/eng/common/templates/job/job.yml
+++ b/eng/common/templates/job/job.yml
@@ -35,7 +35,7 @@ parameters:
   runAsPublic: false
 # Sbom related params
   enableSbom: true
-  PackageVersion: 7.0.0
+  PackageVersion: 8.0.0
   BuildDropPath: '$(Build.SourcesDirectory)/artifacts'
 
 jobs:
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
@@ -19,6 +19,8 @@ using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using LogLevel = Microsoft.Extensions.Logging.LogLevel;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Hosting;
 
 namespace Microsoft.DotNet.XHarness.CLI.Commands;
 
@@ -129,8 +131,8 @@ public class WebServer
             {
                 app.Use((context, next) =>
                 {
-                    context.Response.Headers.Add("Cross-Origin-Embedder-Policy", "require-corp");
-                    context.Response.Headers.Add("Cross-Origin-Opener-Policy", "same-origin");
+                    context.Response.Headers.Append("Cross-Origin-Embedder-Policy", "require-corp");
+                    context.Response.Headers.Append("Cross-Origin-Opener-Policy", "same-origin");
                     return next();
                 });
             }
--- a/src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
+++ b/src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
     <Nullable>enable</Nullable>
     <IsPackable>true</IsPackable>
     <PackAsTool>true</PackAsTool>
--- a/eng/common/templates/steps/generate-sbom.yml
+++ b/eng/common/templates/steps/generate-sbom.yml
@@ -5,7 +5,7 @@
 # IgnoreDirectories - Directories to ignore for SBOM generation. This will be passed through to the CG component detector.
 
 parameters:
-  PackageVersion: 7.0.0
+  PackageVersion: 8.0.0
   BuildDropPath: '$(Build.SourcesDirectory)/artifacts'
   PackageName: '.NET'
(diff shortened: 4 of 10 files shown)


eng/common/templates/job/job.yml
```diff
@@ -35,7 +35,7 @@ parameters:
   runAsPublic: false
 # Sbom related params
   enableSbom: true
-  PackageVersion: 7.0.0
+  PackageVersion: 8.0.0
```
pavelsavara, line 38: I'm not sure what this does and if it should be done

premun (changes requested):
Please revert all changes in `eng/common` as this folder comes from Arcade

premun (approved):

src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
```
premun, line 5: I am thinking about the other projects.. There are some dependent nuget packages such as `Microsoft.DotNet.XHarness.Common`. Maybe other projects could be updated too?

pavelsavara: That would not run in the `net6.0` process no ? (unless you mean we should add targets, not replace it)
Is `net6.0` xharness still in use ?

pavelsavara:
@premun I think that the CI failure is unrelated but it doesn't allow me to merge. Hints ?

(I don't want to do further changes you hinted at)

radical:
Don't merge yet.

radical:
How does it fail without this change?

radical:
I talked to @pavelsavara offline. The middleware project in `runtime` which this is intended to fix builds for 
https://github.com/dotnet/runtime/blob/a024a97885bc23e833a5d1d1696029181b939f05/src/libraries/Common/tests/System/Net/Prerequisites/LocalEchoServer.props#L3-L4

But `$(AspNetCoreAppCurrent)` changed to `net8.0` on Jan 12. So, we should have been seeing the wasm library tests failing, which we haven't. @pavelsavara will be investigating this more.

And we should hold merging this till then.

IOW, the reason for this change as mentioned in the PR description will be investigated. But if we want to update xharness to net8.0 outside that reason, then we can merge!

radical (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
```diff
@@ -1,25 +1,27 @@
...
-using Microsoft.AspNetCore.Routing;
-using Microsoft.AspNetCore.StaticFiles;
-using Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.FileProviders;
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
-using LogLevel = Microsoft.Extensions.Logging.LogLevel;
-
+// Licensed to the .NET Foundation under one or more agreements.
```
radical, line 1: This is changing the file to have windows line-endings.

src/Microsoft.DotNet.XHarness.CLI/Microsoft.DotNet.XHarness.CLI.csproj
lewing: net6.0 is still lts and you can target all of them from the sdk

pavelsavara:
> So, we should have been seeing the wasm library tests failing, which we haven't. @pavelsavara will be investigating this more.

Not any time soon. I have too much on my plate right now. Feel free to take over @radical

radical:
I haven't seen that error though, only you have ;)

radical:
Update: the original failure that prompted this is unknown, so this isn't blocking anything right now.

@premun if the changes here are good in general, then we can merge.

pavelsavara:
> I haven't seen that error though, only you have ;)

I got this locally few days back, and I don't have it anymore

```
Could not load file or assembly 'System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
           at System.Reflection.RuntimeModule.GetTypes(RuntimeModule module)
           at System.Reflection.Assembly.GetTypes()
           at Microsoft.DotNet.XHarness.CLI.CommandArguments.TypeFromAssemblyArgument`1.GetLoadedTypes()+MoveNext() in /_/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/TypeFromAssemblyArgument.cs:line 29
           at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Boolean& found)
           at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source)
           at Microsoft.DotNet.XHarness.CLI.Commands.Wasm.WasmTestBrowserCommand.InvokeInternal(ILogger logger) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs:line 48
           at Microsoft.DotNet.XHarness.CLI.Commands.XHarnessCommand`1.Invoke(IEnumerable`1 arguments) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/XHarnessCommand.cs:line 145
```

Speculation:
I imagine that it happened because xharness was running on random `dotnet.exe` which was on my PATH. And that dotnet didn't have Net8 runtime installed and so it run in Net7 process. And then the `RemoteLoopServer.csproj` plugin which is now Net8 could not find Net8 runtime via reflection.

Do we execute `dotnet` without full path somewhere ? Instead of explicit path to dotnet specified in `runtime\artifacts\toolset\sdk.txt` ?

PR #987, opened 2023-01-17
Handle faulty TCP connection between XHarness and Apple devices

AlitzelMendez:
Returning a new type when a TCP connection fails


premun (approved):
Thanks for this, Ali!

PR #991, opened 2023-02-01
[wasm] On windows look for .exe, .cmd, and .bat when searching for

radical:
.. binaries in PATH

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Common/Utilities/FileUtils.cs
+++ b/src/Microsoft.DotNet.XHarness.Common/Utilities/FileUtils.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Runtime.InteropServices;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Microsoft.DotNet.XHarness.Common.Utilities;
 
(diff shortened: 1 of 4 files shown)


pavelsavara (approved):

mkhamoyan (approved):

src/Microsoft.DotNet.XHarness.Common/Utilities/FileUtils.cs
```diff
@@ -3,30 +3,42 @@
...
 namespace Microsoft.DotNet.XHarness.Common.Utilities;
 
 public static class FileUtils
 {
-    public static string FindFileInPath(string engineBinary)
+    private static readonly string[] s_extensions = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                                                        ? new[] { ".exe", ".cmd", ".bat" }
+                                                        : new[] { "" };
+
+    public static (string fullPath, string? errorMessage) FindExecutableInPATH(string filename)
```
maraf, line 18: Probably just a matter of preference, but wouldn't it be more natural to throw an exception, like `FileNotFoundException`, instead of returning a tuple?

maraf (approved):

src/Microsoft.DotNet.XHarness.Common/Utilities/FileUtils.cs
radical: I wanted the control of what to do if the file is not found to be with the caller.

radical: I changed the code to follow the Try* pattern.

maraf: Yeah, I understood. What I meant was throwing and catching it later to produce correct exit code, but the Try* pattern looks good 👍

maraf (approved):

radical:
Test failure is unrelated.

PR #992, opened 2023-02-06
Run tests with requested locale.

ilonatommy:
For https://github.com/dotnet/runtime/pull/80421/ we need tests to run with specific locale set, otherwise we won't be able to test the mechanism of loading different internationalization data based on the client's locale.
I added `-locale=xx-YY` flag that will set environment variable before the test on Browser / Node is launched. Default is "en-US".

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/Arguments/LocaleArgument.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/Arguments/LocaleArgument.cs
@@ -5,9 +5,9 @@
 namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm;
 using System;
 
-internal class LocaleArgument : StringArgument
+internal class LocaleArgument : RequiredStringArgument
 {
     public LocaleArgument(string defaultValue)
-        : base("locale=", $"Sets LANG environment variable, default value {defaultValue}")
+        : base("locale=", $"Sets LANG environment variable, default value {defaultValue}", defaultValue)
     {}
 }
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/Arguments/LocaleArgument.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/Arguments/LocaleArgument.cs
@@ -8,6 +8,6 @@ using System;
 internal class LocaleArgument : RequiredStringArgument
 {
     public LocaleArgument(string defaultValue)
-        : base("locale=", $"Sets LANG environment variable, default value {defaultValue}", defaultValue)
+        : base("locale=", $"Sets the browser's/node's language through an environment variable, default value {defaultValue}", defaultValue)
     {}
 }
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/Arguments/LocaleArgument.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/Arguments/LocaleArgument.cs
@@ -1,0 +1,13 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm;
+using System;
+
+internal class LocaleArgument : StringArgument
+{
+    public LocaleArgument(string defaultValue)
+        : base("locale=", $"Sets LANG environment variable, default value {defaultValue}")
+    {}
+}
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
@@ -61,6 +61,8 @@ internal class WasmTestBrowserCommand : XHarnessCommand<WasmTestBrowserCommandAr
                             logProcessor,
                             logger);
 
+        // Browser respects LANGUAGE in the first place, only if empty it checks LANG
+        Environment.SetEnvironmentVariable("LANGUAGE", Arguments.Locale);
         (DriverService driverService, IWebDriver driver) = Arguments.Browser.Value switch
         {
             Browser.Chrome => GetChromeDriver(logger),
@@ -71,8 +73,6 @@ internal class WasmTestBrowserCommand : XHarnessCommand<WasmTestBrowserCommandAr
             // shouldn't reach here
             _ => throw new ArgumentException($"Unknown browser : {Arguments.Browser}")
         };
-        // Browser respects LANGUAGE in the first place, only if empty it checks LANG
-        Environment.SetEnvironmentVariable("BROWSER_LANGUAGE", Arguments.Locale);
 
         try
         {
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
@@ -71,6 +71,8 @@ internal class WasmTestBrowserCommand : XHarnessCommand<WasmTestBrowserCommandAr
             // shouldn't reach here
             _ => throw new ArgumentException($"Unknown browser : {Arguments.Browser}")
         };
+        // Browser respects LANGUAGE in the first place, only if empty it checks LANG
+        Environment.SetEnvironmentVariable("LANGUAGE", Arguments.Locale);
 
         try
         {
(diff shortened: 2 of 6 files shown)


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/Arguments/LocaleArgument.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm;
+using System;
+
+internal class LocaleArgument : StringArgument
+{
+    public LocaleArgument(string defaultValue)
```
premun, line 10: I think you're not using the `defaultValue` in any way? Have a look at `RequiredStringArgument` and pass it to its ctor.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -113,6 +113,8 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
             }
 
             engineArgs.AddRange(PassThroughArguments);
+            // Node respects LANG only, ignores LANGUAGE
+            Environment.SetEnvironmentVariable("LANG", Arguments.Locale);
```
premun, line 117: It's also possible to set the env just for the process you're launching with `processManager` below. Maybe that's better? Or do you need this for other processes you don't control?

ilonatommy: Only for the process below, should get changed, thanks

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/WASM/Arguments/LocaleArgument.cs
```diff
@@ -0,0 +1,13 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments.Wasm;
+using System;
+
+internal class LocaleArgument : RequiredStringArgument
+{
+    public LocaleArgument(string defaultValue)
+        : base("locale=", $"Sets LANG environment variable, default value {defaultValue}", defaultValue)
```
premun, line 11: nit: I guess this text doesn't apply for node so we could make it more generic and not spell out the concrete variable?

E.g. 
```suggestion
        : base("locale=", $"Sets the browser's/node's language through an environment variable, default value {defaultValue}", defaultValue)
```

premun: Also it's clear it's not setting XHarness' output language or similar

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -71,6 +71,8 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
             // shouldn't reach here
             _ => throw new ArgumentException($"Unknown browser : {Arguments.Browser}")
         };
+        // Browser respects LANGUAGE in the first place, only if empty it checks LANG
+        Environment.SetEnvironmentVariable("LANGUAGE", Arguments.Locale);
```
premun, line 75: This one stays?

ilonatommy: For this one I haven't figured out the way yet, reading into it

premun (approved):
LGTM! But I'd suggest also getting a review from someone on the WASM team.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -71,6 +71,8 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
             // shouldn't reach here
             _ => throw new ArgumentException($"Unknown browser : {Arguments.Browser}")
         };
+        // Browser respects LANGUAGE in the first place, only if empty it checks LANG
+        Environment.SetEnvironmentVariable("BROWSER_LANGUAGE", Arguments.Locale);
```
pavelsavara, line 75: There is `--lang` for chrome. Is that better ?

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
```diff
@@ -65,6 +66,10 @@ internal static async Task<ServerURLs> Start(IWebServerArguments arguments, stri
                     }
                 });
             })
+            .UseConfiguration(
+                new ConfigurationBuilder()
+                .AddEnvironmentVariables(prefix: "BROWSER_")
```
pavelsavara, line 71: Why we are doing it on server side?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
ilonatommy: It does not work on Linux, https://developer.chrome.com/docs/extensions/reference/i18n/#linux

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
ilonatommy: That was the only idea I came up with to pack setting the LANGUAGE var for the testing process only. I don't see any way of setting it to the BrowserDriver because adding "--lang" argument on Linux it does not have any effect. In my understanding it's the equivalent of setting the env variable for the test process, isn't it?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
radical: We have been using `--lang` with debugger tests - https://github.com/dotnet/runtime/blob/main/src/mono/wasm/debugger/DebuggerTestSuite/ChromeProvider.cs#L166-L167 . Could you also confirm that it doesn't actually work there, and update if needed?

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/JS/WasmTestCommand.cs
```diff
@@ -138,7 +138,9 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
                 log: new CallbackLog(m => logger.LogInformation(m)),
                 stdoutLog: new CallbackLog(msg => logProcessor.Invoke(msg)),
                 stderrLog: new CallbackLog(logProcessor.ProcessErrorMessage),
-                Arguments.Timeout);
+                Arguments.Timeout,
+                // Node respects LANG only, ignores LANGUAGE
+                environmentVariables: new Dictionary<string,string>() { {"LANG", Arguments.Locale} });
```
radical, line 143: `wasm test` runs for v8 also, and will be running for other js engines too.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
ilonatommy: So far it works because we are loading full ICU in debugger tests. As soon as we merge ICU sharding PR the line you quoted will be good for Windows but will fail with Linux. I already fixed it on that PR, it's in this commit: https://github.com/dotnet/runtime/pull/80421/commits/dde05ee00bf0aecacfbfe6e302f34f65e272e2d2.

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
radical: If the browser process needs this environment variable then this needs to be set when launching chrome.

ilonatommy: Maybe I will paste my test screenshots here from Linux.
Not good:
image
Good:
image

radical: Set the environment variable on the process launch then.

pavelsavara (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
ilonatommy: Credits @radical:
image

PR #1001, opened 2023-03-01
[wasm] Add a new `wasm webserver` command

radical:
- starts a webserver in the given app path
- waits till a timeout
- adds support for serving default files like `index.html`.

Prompted by need to run blazor published apps by starting a webserver for them

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/WebServerUseDefaultFiles.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/WebServerUseDefaultFiles.cs
@@ -1,0 +1,13 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments;
+
+internal class WebServerUseDefaultFilesArguments : SwitchArgument
+{
+    public WebServerUseDefaultFilesArguments()
+        : base("web-server-use-default-files", "Enable default files like index.html", false)
+    {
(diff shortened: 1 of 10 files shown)


pavelsavara (approved):
LGMT. Could you please explain where/how this would be used ? Thanks.

maraf:
Do we want to merge it with the simple server from the runtime repo?

radical:
> Do we want to merge it with the simple server from the runtime repo?

Do you mean the webserver in the app host?

pavelsavara:
> Do we want to merge it with the simple server from the runtime repo?

Do you mean use this xharness instead of simple server ? 

Pros: one codebase, consistent behavior
Cons: wait for xharness to propagate to runtime repo on any change

We also want to have blazor like server as part of the host. Should we replace simple server with that ?
Where would that live ? In SDK ?

radical:
> LGMT. Could you please explain where/how this would be used ? Thanks.

Prompted by need to run blazor published apps by starting a webserver for them. See https://github.com/dotnet/runtime/pull/82748

pavelsavara:
Another consideration is that the web server in the host, should be rather production grade. xharness is rather test utility.
Out simple server is also test utility and has helpers for our internal needs (collecting benchmark results and now also simulating errors).

radical:
> Another consideration is that the web server in the host, should be rather production grade. xharness is rather test utility.
> Our simple server is also test utility and has helpers for our internal needs (collecting benchmark results and now also simulating errors).

Yep, I don't think it is worth trying to share the webserver here, and in the apphost. And I'm looking at this as for tests, and a generic-ish one (though it has some wasm specific bits, which is fine).

radical:
Failures are unrelated.

premun:
Seems like the queue is backed up. I can try to merge myself.

PR #1004, opened 2023-03-10
Revert "Revert "Categorize crash as TCP problem...""

premun:
Reverts dotnet/xharness#1002


premun:
@akoeplinger @steveisok I understand that XHarness wasn't to blame last week. Have you managed to track down what was crashing the catalysts and we can add this back?

akoeplinger (approved):

akoeplinger:
We haven't tracked down the catalyst issues yet but we already determined it wasn't this one so let's get this back in.

PR #1008, opened 2023-03-27
Stop using 10.15 queues in repo's E2E tests

premun:
The 10.15 queue is getting deprecated soon and devices were already moved out of the queue.


premun:
@akoeplinger 10.15 is going away soon and we don't pay too much attention to it anymore, right? Any thoughts on this?

I see failures on 10.15 and that we've also removed devices from the 10.15 queues.

MilenaHristova (approved):

Issue #1010, opened 2023-03-31
Servicing branches should be created for 6.0/7.0


premun:
We should be able to make small servicing changes such as mlaunch bumps. Subscriptions should be created for the release branches to arcade and runtime

PR #1011, opened 2023-04-12
Bump mlaunch and SdkVersions

premun:
cc @rmarinho @akoeplinger

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS.Shared/SdkVersions.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS.Shared/SdkVersions.cs
@@ -5,11 +5,11 @@
 /// </summary>
 public static class SdkVersions
 {
-    public static string Xcode { get; private set; } = "13.3";
-    public static string OSX { get; private set; } = "12.00";
-    public static string iOS { get; private set; } = "15.4";
+    public static string Xcode { get; private set; } = "14.3";
+    public static string OSX { get; private set; } = "13.00";
+    public static string iOS { get; private set; } = "16.4";
     public static string WatchOS { get; private set; } = "6.2";
-    public static string TVOS { get; private set; } = "15.4";
+    public static string TVOS { get; private set; } = "16.4";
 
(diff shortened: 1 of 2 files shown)


rmarinho:
I think the fix we want is this one https://github.com/xamarin/maccore/pull/2644/files for XCode 14.3

dalexsoto (approved):

src/Microsoft.DotNet.XHarness.iOS.Shared/SdkVersions.cs
```diff
@@ -5,11 +5,11 @@
 /// </summary>
 public static class SdkVersions
 {
-    public static string Xcode { get; private set; } = "13.3";
-    public static string OSX { get; private set; } = "12.00";
-    public static string iOS { get; private set; } = "15.4";
+    public static string Xcode { get; private set; } = "14.3";
```
akoeplinger, line 8: looks like this isn't used, why do we need this?

akoeplinger: and does this mean we'd require newest versions of Xcode or iOS Simulators? just thinking since we don't have branches for xharness yet: https://github.com/dotnet/xharness/issues/1010

akoeplinger (approved):

src/Microsoft.DotNet.XHarness.iOS.Shared/SdkVersions.cs
rmarinho: Yeah I think we need to test this with Ventura/Xcode 14.3

rmarinho:
Can we merge?

src/Microsoft.DotNet.XHarness.iOS.Shared/SdkVersions.cs
premun: Actually, I don't think these matter much. I know they were used in xamarin-macios which injects this class but not here

PR #1018, opened 2023-05-02
[release/6.0] Bump mlaunch to 1.0.44

premun:
https://github.com/dotnet/xharness/issues/1010


azure-pipelines.yml
```diff
@@ -36,8 +40,8 @@ stages:
       - job: Windows_NT
         pool:
           ${{ if eq(variables._RunAsInternal, True) }}:
-            name: NetCoreInternal-Pool
-            queue: BuildPool.Server.Amd64.VS2019
+            name: NetCore1ESPool-Internal
```
akoeplinger, line 43: this will need to use the -Svc pool (probably best through the pool-providers.yml)

akoeplinger (approved):

azure-pipelines.yml
premun: Thanks, that's a good point

PR #1030, opened 2023-06-07
[iOS] Move tests to the OSX 13 queue

steveisok:
Since https://github.com/dotnet/arcade/issues/13622 was completed, we can move to the new queue.


premun:
I believe we will have to wait with this. All of our MacOS machines went dark this week

steveisok:
My mistake, the queue is available, but the devices has not yet been moved.

PR #1053, opened 2023-07-19
[iOS] Ensure that the TCC database will work with the Xcode15

mandel-macaque:
There is a bug in the iOS simulator from apple in which it does not respect the permissions set by simtcl. This is an issue because the tool adds a record with the value 4 in the auth type/mode but the simulator only respects the permission when set with value 2. To fix it we are going to be adding a trigger than when it sees a new permission added with mode 4 it will update the row with value 2 so that the simulator does respect the settings.


dalexsoto (approved):

haritha-mohan (approved):

premun:
@mandel-macaque you can find this in `8.0.0-prerelease.23370.1`

PR #1057, opened 2023-08-03
[iOS] TCCDatabase - even if we cannot create the trigger call the privacy tool

mandel-macaque:
Some bots give errors when we try to edit the db, if we fail creating the trigger, continue and hope that the privacy tool does work.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/TCCDatabase.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/TCCDatabase.cs
@@ -101,7 +101,6 @@ public class TCCDatabase : ITCCDatabase
         var format = GetTCCFormat(simRuntime);
         if (format >= 4)
         {
-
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/TCCDatabase.cs
```diff
@@ -129,45 +128,43 @@ public async Task<bool> AgreeToPromptsAsync(string simRuntime, string TCCDb, str
 
                 if (!rv.Succeeded)
                 {
-                    failure = true;
+                    // print a warning, but do not set it to failure and how that the simctl privacy command will work
```
akoeplinger, line 131:
```suggestion
                    // print a warning, but do not set it to failure and hope that the simctl privacy command will work
```

akoeplinger (approved):

dustin-wojciechowski (approved):

mandel-macaque:
Tests are a little flaky but the failure is not an issue with the Pr.

PR #1069, opened 2023-09-01
[wasm] Add telemetry for wasm commands

radical:
Wasm engine used, and version where possible.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Common/CommandDiagnostics.cs
+++ b/src/Microsoft.DotNet.XHarness.Common/CommandDiagnostics.cs
@@ -35,17 +35,6 @@ public interface IDiagnosticsData
     /// True when the target is a real HW device, false for simulators, maccatalyst..
     /// </summary>
     bool? IsDevice { get; set; }
-
-    /// <summary>
-    /// Name, or path of the wasm engine used
-    /// </summary>
-    string? WasmEngine { get; set; }
-
-    /// <summary>
-    /// Version for the wasm engine
-    /// </summary>
-    string? WasmEngineVersion { get; set; }
-
 }
 
 /// <summary>
@@ -74,12 +63,6 @@ public class CommandDiagnostics : IDiagnosticsData
 
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public bool? IsDevice { get; set; }
-
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public string? WasmEngine { get; set; }
-
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public string? WasmEngineVersion { get; set; }
 
     public int Duration => (int)Math.Round(_timer.Elapsed.TotalSeconds);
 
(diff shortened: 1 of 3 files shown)


radical:
@adamsitnik Is this enough to get the data into kusto? And the corresponding change in `dotnet/runtime` to have `EnableXHarnessTelemetry=true`.

adamsitnik:
> @adamsitnik Is this enough to get the data into kusto? And the corresponding change in `dotnet/runtime` to have `EnableXHarnessTelemetry=true`.

I am sorry, I am not familiar with kusto enough to be able to tell. Perhaps you meant to tag someone from the .NET Perf Team that recently worked on moving all data to kusto? Was it @cincuranet ?

src/Microsoft.DotNet.XHarness.Common/CommandDiagnostics.cs
```diff
@@ -64,6 +75,12 @@ public class CommandDiagnostics : IDiagnosticsData
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public bool? IsDevice { get; set; }
 
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? WasmEngine { get; set; }
```
premun, line 79: I'd suggest to not add new fields but rather put the engine into the `Target`/`TargetOS` fields as you won't have to plumb these further to Kusto.

For instance, in emulators/phones, this would have the version of the Android and architecture.

premun (commented):
@radical this won't be enough to start seeing the data in Kusto.

The way it works:
1. XHarness reads an env `XHARNESS_DIAGNOSTICS_PATH`
2. If it's set, it will output the `IDiagnosticData` object from DI into this path as JSON
3. If more commands are run, they are appended into one JSON file
4. This is all that XHarness does as it is agnostic to the environment it's used in

What happens next is that the [Helix XHarness SDK](https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Helix/Sdk/tools/xharness-runner), which we use to create XHarness jobs for Android/Apple is wrapping user jobs with custom scripts, takes the JSON file and sends it to Helix. Not sure you're using this in WASM but I think not. However, this Helix XHarness SDK is a glue layer for XHarness and Helix. It knows how to clean up a Helix machine before/after, how to install and run XHarness in between etc.

This wrapper script calls into this python script: https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Helix/Sdk/tools/xharness-runner/xharness-event-processor.py

This script reads the JSON file and (using some existing Helix python scripts on the machine) sends the data from the file to EventHub from where it later gets ingested to Kusto by Helix.

This means that you will have to:
1. Include this `.py` script in your WASM jobs
2. Set the env variable to create the JSON file
3. Call the `.py` script at the end (leave a time buffer in your job, it usually takes a second or so)
4. Do not worry about about credentials etc, this should work out of the box

The various items from `IDiagnosticData` itself get put into a JSON property bag in a Kusto column.
Have a look at an example query:
```kql
Metrics
| where EventType == "_MobileDeviceOperation" and MetricName == "ExitCode" and Timestamp > now()-1d
| extend Dimensions = parse_json(Dimensions)
| extend
    ExitCode = MetricValue,
    Command = tostring(Dimensions.command),
    Platform = tostring(Dimensions.platform),
    Target = tostring(Dimensions.target)
| where Platform == "android" and Command in ("run", "test")
| limit 100
```

radical:
@premun Thank you for the comprehensive explanation!!

premun (approved):

premun:
@radical please double check that the JSON gets created when you run XHarness with the right option

radical:
> @radical please double check that the JSON gets created when you run XHarness with the right option

I can see it being generated locally, thanks for the reminder!

PR #1071, opened 2023-09-05
Bump mlaunch to latest version

akoeplinger:
mlaunch is a .NET app now.

### DIFF
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -1,19 +1,12 @@
 <Project>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


eng/Signing.props
```diff
@@ -1,19 +1,12 @@
 <Project>
 
   <ItemGroup>
-    <FileSignInfo Include="mlaunch.exe" CertificateName="3PartySHA2" />
```
premun, line 4: Ah, nice, we should have done this some time ago I guess.

premun (approved):

PR #1073, opened 2023-09-11
Add a new enable relay argument

mandel-macaque:
This PR contains only the CLI changes to allow xharness to know that a relay is needed. The idea behind this is that we want to be able to open a relay port with the device so that we can use a protocol to interact with the tests on an application. This way we will add support to integrating the test runner in the device with an IDE.

There is no platform specific implementation in this PR, that will come in the following days, the aim is to show the new argument and how it will be passed, once we agree on that, we can move to the actual implementation.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/AndroidRunCommandArguments.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/AndroidRunCommandArguments.cs
@@ -18,6 +18,7 @@ internal class AndroidRunCommandArguments : XHarnessCommandArguments, IAndroidAp
     public InstrumentationNameArgument InstrumentationName { get; } = new();
     public InstrumentationArguments InstrumentationArguments { get; } = new();
     public ExpectedExitCodeArgument ExpectedExitCode { get; } = new((int)Common.CLI.ExitCode.SUCCESS);
+    public EnableRelayArgument EnableRelay { get; } = new();
     public DeviceOutputFolderArgument DeviceOutputFolder { get; } = new();
     public WifiArgument Wifi { get; } = new();
 
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/EnableRelayArgument.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/EnableRelayArgument.cs
@@ -1,0 +1,11 @@
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments;
+
+/// <summary>
+/// Enable xharness to function as a relay when communicating with the appliecation
+/// </summary>
+internal class EnableRelayArgument : SwitchArgument
+{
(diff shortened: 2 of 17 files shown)


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/EnableRelayArgument.cs
```diff
@@ -0,0 +1,11 @@
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments;
+
+/// <summary>
+/// Enable xharness to function as a relay when communicating with the appliecation
```
premun, line 4:
```suggestion
/// Enable XHarness to function as a relay when communicating with the application
```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Arguments/EnableRelayArgument.cs
```diff
@@ -0,0 +1,11 @@
+namespace Microsoft.DotNet.XHarness.CLI.CommandArguments;
```
premun, line 1:
```suggestion
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.
namespace Microsoft.DotNet.XHarness.CLI.CommandArguments;
```

src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Android/AndroidRunCommandArguments.cs
```diff
@@ -18,6 +18,7 @@ internal class AndroidRunCommandArguments : XHarnessCommandArguments, IAndroidAp
     public InstrumentationNameArgument InstrumentationName { get; } = new();
     public InstrumentationArguments InstrumentationArguments { get; } = new();
     public ExpectedExitCodeArgument ExpectedExitCode { get; } = new((int)Common.CLI.ExitCode.SUCCESS);
+    public EnableRelayArgument EnableRelay { get; } = new();
```
premun, line 21: Should this be added to `IAndroidAppRunArguments`?

src/Microsoft.DotNet.XHarness.iOS.Shared/Listeners/SimpleListenerFactory.cs
```diff
@@ -61,6 +64,9 @@ public class SimpleListenerFactory : ISimpleListenerFactory
 
         switch (transport)
         {
+            case ListenerTransport.Relay:
```
premun, line 67: I don't follow - is this value actually set above? Or this branch will never trigger?

premun, line 67: This seems like the `case ListenerTransport.Tcp:`. What's the actual functional effect of passing `relay = true` in here? Shouldn't we do something special with the listener then?

premun:
I am closing this, feel free to re-open if this ever becomes a real feature

PR #1081, opened 2023-10-10
[release/6.0] Use servicing queues for Android emulators

premun:
https://github.com/dotnet/dnceng/issues/952


premun:
@akoeplinger we need to do `.svc` but there was no 2204 svc variant of the android queue. I guess it's fine?

akoeplinger:
@premun I'd rather add .svc variants of the 22.04 queues, we're trying to get away from 18.04 everywhere

akoeplinger:
Btw. I think we should play with passing DncEngPublicBuildPool into msbuild so we can check whether it contains `.Svc` so we can do the svc queue automatically rather than needing to update it all the time (which we'll forget).

premun:
> Btw. I think we should play with passing DncEngPublicBuildPool into msbuild so we can check whether it contains .Svc so we can target the svc queue automatically rather than needing to update the tests all the time (which we'll forget).

I like this idea, we should do that in main though?

premun:
> I'd rather add .svc variants of the 22.04 queues, we're trying to get away from 18.04 everywhere

I will do it next week when I'm on FR

premun:
@akoeplinger I am thinking about merging this as it levels the queues used in runtime and xharness' servicing branches to the same. It will also clear an alert about us running jobs in non-svc queues.
Next week I will add the new queues - https://github.com/dotnet/dnceng/issues/993 - and we can then switch everything.

This PR also turns off the simulator installer tests which are failing as apple changed the way we get the packages but we don't really utilize these commands in 6.0/7.0

premun:
This is waiting on https://github.com/dotnet/dnceng/issues/993

premun:
@akoeplinger can you approve this one as well? I will bump the queues once they are rolled out. However this will unblock things like updating mlaunch in the 6.0 branches etc

akoeplinger (approved):

PR #1093, opened 2023-10-25
Branding, CPM, cleanup and updates

ViktorHofer:
1. Use floating TFM versions which get auto-updated via the Arcade.Sdk.
2. Enable NuGet Central Package Management and pin
   Newtonsoft.Json to 13.0.3. Use the XUnit version from Arcade (2.5.3).
3. Small clean-ups (XML stuff mostly), remove default publishing version set
4. Branding update (commit 3)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,13 +1,13 @@
-<?xml version="1.0" encoding="utf-8"?>
 <Project>
+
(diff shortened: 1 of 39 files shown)


premun (approved):
Ah, this is nice! Thanks!

PR #1114, opened 2023-12-06
Add a polling task to return control to the browser

lewing:


pavelsavara:
It goes this way

image

image

image

and suddenly it becomes `async void` on the inherited class

image

image

image

image

image

image

image

PR #1121, opened 2023-12-20
Remove unnecessary properties from integration tests project

akoeplinger:


akoeplinger:
hmm looks like those are needed :D

PR #1140, opened 2024-01-16
[iOS] Use a more modern device for XCode 15.2

rmarinho:
Based on some investigation I did with Xcode 15.2 and previous, seems the device type that xharness creates is not allowed in more recent versions. 
By looking at what simulators I was able to create and that show up on Xcode 15.2 i think we want to use the SE 3rd generation as a base for old versions. 

Screenshot 2024-01-16 at 12 59 53


akoeplinger:
Looks like iPhone SE 3rd generation is not supported by the Xcode used in CI.
The iPhone X should be fine in newer Xcodes (I checked with 15.2) but the minimum version is iPhone 6s now.

I think your issue was due to passing --target=ios-simulator which defaulted to 32bit simulator and iPhone 5. I fixed that in the PR and I think I'll just remove 32bit simulator support from main

akoeplinger (approved):

PR #1145, opened 2024-01-26
[wasm] Switch between threadless and multi-threaded runner

maraf:
- Switch between threadless and multi-threaded XUnit runner for WASM based on `IsThreadless` property
- The property will be set in the runtime based on passed flags and based on used runtime flavor
- Mutli-threaded runner extends the runner for other platforms and overrides how results are written (as single line)
- Add current thread id to STRT, PASS and FAIL messages when enabled with `XHARNESS_LOG_THREAD_ID`

Contributes to https://github.com/dotnet/runtime/issues/88763

### DIFF
--- a/src/Microsoft.DotNet.XHarness.TestRunners.Xunit/EnvironmentVariables.cs
+++ b/src/Microsoft.DotNet.XHarness.TestRunners.Xunit/EnvironmentVariables.cs
@@ -1,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.XHarness.TestRunners.Xunit;
+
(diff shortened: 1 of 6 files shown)


pavelsavara (approved):

pavelsavara:
Note that the thread ID could be misleading. The task could start on different thread than all the continuations etc ...

maraf:
> Note that the thread ID could be misleading. The task could start on different thread than all the continuations etc ...

I do undestand. When we talked about it few months, you agreed it could be usefull in sometimes.

pavelsavara:
> > Note that the thread ID could be misleading. The task could start on different thread than all the continuations etc ...
> 
> I do undestand. When we talked about it few months, you agreed it could be usefull in sometimes. I can revert it if you don't any value in it now

I learned more since. I don't mind having it behind a flag and occasionally try if it helps to resolve something.

PR #1158, opened 2024-02-27
[IGNORE] boot the device instead of trying to open it

ivanpovazan:
Do not review/merge.

Just testing xharness CLI orchestration.


premun (approved):
Cool, I always wondered what TCC stands for 😅

Issue #1171, opened 2024-03-15
XCode 15.3 support


vaind:
```shell
$ xharness apple simulators list                                                                       
[9.0.0-prerelease.24161.1+134035492ed8154fc9c5a930a4ca52c422b21afb] XHarness command issued: apple simulators list
crit: System.Exception: Failed to detect Xcode UUID!
         at Microsoft.DotNet.XHarness.CLI.Commands.Apple.Simulators.SimulatorsCommand.GetXcodeInformation() in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/Simulators/SimulatorsCommand.cs:line 401
         at Microsoft.DotNet.XHarness.CLI.Commands.Apple.Simulators.SimulatorsCommand.GetSimulatorIndexXml() in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/Simulators/SimulatorsCommand.cs:line 304
         at Microsoft.DotNet.XHarness.CLI.Commands.Apple.Simulators.SimulatorsCommand.GetAvailableSimulators() in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/Simulators/SimulatorsCommand.cs:line 95
         at Microsoft.DotNet.XHarness.CLI.Commands.Apple.Simulators.ListCommand.InvokeInternal(ILogger logger) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/Simulators/ListCommand.cs:line 33
         at Microsoft.DotNet.XHarness.CLI.Commands.XHarnessCommand`1.Invoke(IEnumerable`1 arguments) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/XHarnessCommand.cs:line 145
XHarness exit code: 71 (GENERAL_FAILURE)
```

The failing code:

https://github.com/dotnet/xharness/blob/a496e7e9e4d445b420dc818c157073ee3e6b16c1/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/Simulators/SimulatorsCommand.cs#L398-L402

And what I get when I run it manually: 

```shell
$/usr/libexec/PlistBuddy -c 'Print DVTPlugInCompatibilityUUID' /Applications/Xcode-15.3.0.app/Contents/Info.plist
Print: Entry, "DVTPlugInCompatibilityUUID", Does Not Exist
```

akoeplinger:
Thanks for the report, opened https://github.com/dotnet/xharness/pull/1172

PR #1172, opened 2024-03-16
Support Xcode 15.3 simulators

akoeplinger:
DVTPlugInCompatibilityUUID is no longer available in the Info.plist so stop reading it. We only really used it for Xcode < 14 anyway.

Fixes https://github.com/dotnet/xharness/issues/1171


src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/Simulators/SimulatorsCommand.cs
```diff
@@ -393,23 +395,33 @@ private async Task<(string XcodeVersion, string XcodeUuid)> GetXcodeInformation(
 
         xcodeVersion = xcodeVersion.Trim();
 
-        string xcodeUuid;
+        xcodeVersion = xcodeVersion.Insert(xcodeVersion.Length - 2, ".");
```
premun, line 398: Can we have a comment here explaining why?

akoeplinger: added the comment

Issue #1200, opened 2024-05-15
Shipped adb is x86_64 only


borrrden:
This means all commands on Apple silicon fail with the following (unless using rosetta and the entire dotnet runtime is x86_64 I guess?):

```
$HOME/.dotnet/tools/xharness android state
[8.0.0-prerelease.24229.2+aacfb6328fdef17e572617bbb551431bb9cb1ff2] XHarness command issued: android state
crit: Error: An error occurred trying to start process '/Users/couchbase/.dotnet/tools/.store/microsoft.dotnet.xharness.cli/8.0.0-prerelease.24229.2/microsoft.dotnet.xharness.cli/8.0.0-prerelease.24229.2/runtimes/any/native/adb/macos/adb' with working directory '/Users/couchbase/.dotnet/tools/.store/microsoft.dotnet.xharness.cli/8.0.0-prerelease.24229.2/microsoft.dotnet.xharness.cli/8.0.0-prerelease.24229.2/runtimes/any/native/adb/macos'. Bad CPU type in executable
      System.ComponentModel.Win32Exception (86): An error occurred trying to start process '/Users/couchbase/.dotnet/tools/.store/microsoft.dotnet.xharness.cli/8.0.0-prerelease.24229.2/microsoft.dotnet.xharness.cli/8.0.0-prerelease.24229.2/runtimes/any/native/adb/macos/adb' with working directory '/Users/couchbase/.dotnet/tools/.store/microsoft.dotnet.xharness.cli/8.0.0-prerelease.24229.2/microsoft.dotnet.xharness.cli/8.0.0-prerelease.24229.2/runtimes/any/native/adb/macos'. Bad CPU type in executable
         at System.Diagnostics.Process.ForkAndExecProcess(ProcessStartInfo startInfo, String resolvedFilename, String[] argv, String[] envp, String cwd, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
         at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
         at Microsoft.DotNet.XHarness.Android.Execution.AdbProcessManager.Run(String adbExePath, IEnumerable`1 arguments, TimeSpan timeOut) in /_/src/Microsoft.DotNet.XHarness.Android/Execution/AdbProcessManager.cs:line 74
[...]

PR #1201, opened 2024-05-15
[android] Bump Android platform tools version to `32.0.0`

ivanpovazan:
## Description

This PR bumps the version of Android platform tools that we ship to `32.0.0`. 

The reason for this is that on older versions `adb` for MacOS was shipped only as `x86_64` executable, which is problematic if Rosetta is disabled on arm64 Mac devices on which xharness android commands are invoked.

Version `32.0.0` of Android tools: https://developer.android.com/tools/releases/platform-tools#3200_january_2022 has changed this limitation and `adb` now ships as the universal binary for Apple M1 devices.

---
Fixes https://github.com/dotnet/xharness/issues/1200


vitek-karas (approved):
Looks good - the test failure is weird... but it does look unrelated.

PR #1211, opened 2024-06-11
[ios] Skip downloading simulators which don't have their source specified

ivanpovazan:
## Description 

When we want to install simulators through xharness (via for example: `apple simulators install ios-simulator-64_17.2`) we first fetch simulator index file from Apple server. From that file we parse the available simulators and fetch the sources for their download. 

However, it can happen that for beta simulators Apple does not provide the `source` for download which currently breaks the xharness with:
```
crit: System.Exception: Source node not found
         at Microsoft.DotNet.XHarness.CLI.Commands.Apple.Simulators.SimulatorsCommand.GetAvailableSimulators() in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/Simulators/SimulatorsCommand.cs:line 109
         at Microsoft.DotNet.XHarness.CLI.Commands.Apple.Simulators.InstallCommand.InvokeInternal(ILogger logger) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/Apple/Simulators/InstallCommand.cs:line 41
         at Microsoft.DotNet.XHarness.CLI.Commands.XHarnessCommand`1.Invoke(IEnumerable`1 arguments) in /_/src/Microsoft.DotNet.XHarness.CLI/Commands/XHarnessCommand.cs:line 145
```

The example of the problematic index file with no `source` tag:
```
<dict>
	<key>category</key>
	<string>simulator</string>
	<key>contentType</key>
	<string>cryptexDiskImage</string>
	<key>dictionaryVersion</key>
	<integer>2</integer>
	<key>downloadMethod</key>
	<string>mobileAsset</string>
	<key>fileSize</key>
	<integer>8455760175</integer>
	<key>identifier</key>
	<string>com.apple.dmg.iPhoneSimulatorSDK18_0_b1</string>
	<key>name</key>
	<string>iOS 18.0 beta Simulator Runtime</string>
	<key>platform</key>
	<string>com.apple.platform.iphoneos</string>
	<key>simulatorVersion</key>
	<dict>
		<key>buildUpdate</key>
		<string>22A5282m</string>
		<key>version</key>
		<string>18.0</string>
	</dict>
	<key>version</key>
	<string>18.0.0.1</string>
</dict>
```

## Changes

In case a source is not found for a given simulator in the index file, instead of throwing an exception we now skip the problematic simulator and log a warning message about it e.g.,:
```
[...]

PR #1214, opened 2024-06-12
[release/7.0] [ios] Skip downloading simulators which don't have their source specified

ivanpovazan:
Backport of: https://github.com/dotnet/xharness/pull/1211 to release/7.0


akoeplinger:
7.0 is out of support so we don't need this one.

PR #1221, opened 2024-06-24
Close all tabs before quitting.

ilonatommy:
Trying to fix https://github.com/dotnet/runtime/issues/101617.

Source: https://github.com/mozilla/geckodriver/issues/285#issuecomment-318791007

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
@@ -103,7 +103,18 @@ internal class WasmTestBrowserCommand : XHarnessCommand<WasmTestBrowserCommandAr
                 token.WaitHandle.WaitOne();
             }
 
-            driver.Quit();  // Firefox driver hangs if Quit is not issued.
+            // close all tabs before quit is a workaround for broken Selenium - GeckoDriver communication in Firefox
+            // https://github.com/dotnet/runtime/issues/101617
+            while (driver.WindowHandles.Count > 1)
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -103,7 +103,18 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
                 token.WaitHandle.WaitOne();
             }
 
-            driver.Quit();  // Firefox driver hangs if Quit is not issued.
+            // close all tabs before quit is a workaround for broken Selenium - GeckoDriver communication in Firefox
```
pavelsavara, line 106: maybe add logging here, so that we are sure4 it happened

pavelsavara (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -103,7 +103,19 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
+            // https://github.com/dotnet/runtime/issues/101617
+            logger.LogInformation($"Closing {driver.WindowHandles.Count} browser tabs before setting the main tab to config page and quitting.");
+            while (driver.WindowHandles.Count > 1)
```
akoeplinger, line 109: nit: it'd be nice to have some (short) timeout here so this doesn't end up being an infinite loop if closing a tab doesn't work for some reason

PR #1224, opened 2024-06-25
Prevent infinite loop when closing all tabs before quitting.

ilonatommy:
Follow up for https://github.com/dotnet/xharness/pull/1221. In case of unsuccessful closing of tabs, we would end in infinite loop. Adding 10 sec timeout.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
@@ -106,10 +106,12 @@ internal class WasmTestBrowserCommand : XHarnessCommand<WasmTestBrowserCommandAr
             // close all tabs before quit is a workaround for broken Selenium - GeckoDriver communication in Firefox
             // https://github.com/dotnet/runtime/issues/101617
             logger.LogInformation($"Closing {driver.WindowHandles.Count} browser tabs before setting the main tab to config page and quitting.");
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -106,8 +106,14 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
             // https://github.com/dotnet/runtime/issues/101617
             logger.LogInformation($"Closing {driver.WindowHandles.Count} browser tabs before setting the main tab to config page and quitting.");
+            DateTime timeoutThreshold = DateTime.Now.AddSeconds(10);
             while (driver.WindowHandles.Count > 1)
             {
+                if (DateTime.Now > timeoutThreshold)
+                {
+                    logger.LogInformation($"Timeout while trying to close tabs, {driver.WindowHandles.Count} is left open before quitting.");
+                    break;
+                }
```
akoeplinger, line 116: DateTime.Now is using local time so could be affected by time changes, a more robust approach is:

```suggestion
            var cts = new CancellationTokenSource();
            cts.CancelAfter(10000);

            while (driver.WindowHandles.Count > 1)
            {
                if (cts.IsCancellationRequested)
                {
                    logger.LogInformation($"Timeout while trying to close tabs, {driver.WindowHandles.Count} is left open before quitting.");
                    break;
                }
```

akoeplinger (approved):

PR #1228, opened 2024-07-04
Fix Environment.SpecialFolder.MyDocuments usage in BaseOrchestrator.cs

akoeplinger:
The value changed in .NET 8: https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/getfolderpath-unix#change-description

This was touched in https://github.com/dotnet/xharness/pull/948 but the change is wrong, mlaunch expects this file to be in $HOME: https://github.com/xamarin/maccore/blob/bd1f42be1995b8501542a90671aa668771f1b3a0/tools/mlaunch/Xamarin.Hosting/Xamarin.Launcher/controller-device.cs#L341


ivanpovazan (approved):
LGTM, thanks!

PR #1244, opened 2024-08-01
[WASI] HTTP server, exit code

pavelsavara:
- support HTTP server also for WASI
 - forward exit code via stderr for WASI


src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
```diff
@@ -65,8 +65,36 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
             await PrintVersionAsync(Arguments.Engine.Value.Value, engineBinary);
 
+            ServerURLs? serverURLs = null;
+            if (Arguments.WebServerMiddlewarePathsAndTypes.Value.Count > 0)
+            {
+                serverURLs = await WebServer.Start(
+                    Arguments,
+                    logger,
+                    cts.Token);
+                cts.CancelAfter(Arguments.Timeout);
```
akoeplinger, line 75: the way this is supposed to work is that you add the task into the `tasks` array and it'll trigger cancellation if the last task (the one with `Task.Delay(Arguments.Timeout)`) fires.

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
```diff
@@ -65,8 +65,36 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
+            }
+
             var engineArgs = new List<string>();
             engineArgs.AddRange(Arguments.EngineArgs.Value);
+
+            if (Arguments.WebServerMiddlewarePathsAndTypes.Value.Count > 0)
```
akoeplinger, line 81: extract `Arguments.WebServerMiddlewarePathsAndTypes.Value.Count > 0` into local `bool isWebServerEnabled`

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
```diff
@@ -224,6 +225,12 @@ private void ProcessMessage(string message, bool isError = false)
         if (line.StartsWith("WASM EXIT"))
         {
             _logger.LogDebug("Reached wasm exit");
+
+            // until WASI can work with unix exit code https://github.com/WebAssembly/wasi-cli/pull/44
+            if (line.Length > 10)
+            {
+                ForwardedExitCode = int.Parse(line.Substring(10));
```
akoeplinger, line 232: nit: add a comment that explains we have the exit code in the string and we strip the prefix

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
pavelsavara: There are multiple places where the `cts` is used here, it's not just one Task

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/WasmTestMessagesProcessor.cs
```diff
@@ -1,250 +1,259 @@
...
-        {
-            ScanMessageForErrorPatterns(line);
-            line = Symbolicate(line);
-
-            switch (logMessage?.method?.ToLowerInvariant())
-            {
-                case "console.debug": _logger.LogDebug(line); break;
-                case "console.error": _logger.LogError(line); break;
-                case "console.warn": _logger.LogWarning(line); break;
+// Licensed to the .NET Foundation under one or more agreements.
```
pavelsavara, line 1: there is CRLF fix for most of the file

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
akoeplinger: right, you don't need to add `cts.CancelAfter` here since that already happens later on

lewing (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
```diff
@@ -65,8 +65,35 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
...
+
             var engineArgs = new List<string>();
             engineArgs.AddRange(Arguments.EngineArgs.Value);
+
+            if (Arguments.IsWebServerEnabled)
+            {
+                foreach (var envVariable in Arguments.WebServerHttpEnvironmentVariables.Value)
+                {
+                    engineArgs.Add($"--env");
+                    engineArgs.Add($"{envVariable}={serverURLs!.Http}");
```
ivanpovazan, line 85: nit: Just a question:
Could these be: `engineArgs.Add($"--setenv={envVariable}={serverURLs!.Http}");` ?

ivanpovazan (approved):

ivanpovazan:
/ba-g infra issue with PACKAGE_INSTALLATION_FAILURE of debug apks

akoeplinger (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WASI/Engine/WasiTestCommand.cs
pavelsavara: nope, they need to be separate

PR #1256, opened 2024-09-02
Remove separate CodeQL pipeline

akoeplinger:
CodeQL runs in the main pipeline via 1ESPT now and can't be disabled so having a separate pipeline doesn't make sense.


ivanpovazan (approved):
Just to clarify, this is basically just a clean up as we didn't run it anyway?

akoeplinger:
We do run it (that's why I noticed) but it's pointless :D

ivanpovazan:
You mean this one: https://dev.azure.com/dnceng/internal/_build?definitionId=1184

akoeplinger:
Yes. I'll ask dnceng to delete it once this is merged.

PR #1268, opened 2024-09-13
Allow overwriting `PageLoadStrategy` in wasm tests

ilonatommy:
Trying to fix the new chrome bump: https://github.com/dotnet/runtime/pull/107504.

Till now we were running tests always with `PageLoadStrategy` with default value `Normal`. It is documented to have caused problems in testing with Selenium with older versions of chrome. This change does not change the default value, only provides an option to overwrite it. Because the failures are observed only on chrome, the change is limited to `ChromeOptions`.

https://www.selenium.dev/documentation/webdriver/drivers/options/#pageloadstrategy

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Argument.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Argument.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using OpenQA.Selenium;
 
 namespace Microsoft.DotNet.XHarness.CLI.CommandArguments;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.XHarness.CLI/CommandArguments/Argument.cs
```diff
@@ -283,6 +284,32 @@ public override void Action(string argumentValue)
...
+    {
+        if (string.IsNullOrEmpty(argumentValue))
+        {
+            Value = _defaultValue;
+        }
+        else
+        {
+            Value = argumentValue.ToLower().Equals("none") ? PageLoadStrategy.None :
+                argumentValue.ToLower().Equals("eager") ? PageLoadStrategy.Eager :
+                argumentValue.ToLower().Equals("normal") ? PageLoadStrategy.Normal :
```
akoeplinger, line 307: better to use the compare options instead of tolower+equals:

```suggestion
            Value = argumentValue.Equals("none", StringComparison.OrdinalIgnoreCase) ? PageLoadStrategy.None :
                argumentValue.Equals("eager", StringComparison.OrdinalIgnoreCase) ? PageLoadStrategy.Eager :
                argumentValue.Equals("normal", StringComparison.OrdinalIgnoreCase) ? PageLoadStrategy.Normal :
```

src/Microsoft.DotNet.XHarness.CLI/Commands/WASM/Browser/WasmTestBrowserCommand.cs
```diff
@@ -265,7 +265,10 @@ protected override async Task<ExitCode> InvokeInternal(ILogger logger)
         if (Arguments.NoQuit)
             options.LeaveBrowserRunning = true;
 
-        logger.LogInformation($"Starting {driverName} with args: {string.Join(' ', options.Arguments)}");
+        if (options is ChromeOptions chromeOptions)
+            chromeOptions.PageLoadStrategy = Arguments.PageLoadStrategy.Value;
```
akoeplinger, line 269: making this chrome-only feels like magic behavior to me that is hard to reason about. is there a problem with applying it everywhere? I guess you want to avoid needing to change calling code to specify a different value for chrome only?

ilonatommy: No, there's no problem, it's just that I was not sure what other options can be and if they have this setting while the fix requires chrome-only change.
We can add `FirefoxOptions` and `EdgeOptions` but I can't find docs on this method to exist on these two drivers, there's [chrome doc](https://www.selenium.dev/documentation/webdriver/drivers/options/#pageloadstrategy) but no similar for edge/firefox. Now when you asked I tested the method on edge/firefox and it seems not to cause any problems, so we can add the line for them as well.

akoeplinger (approved):

PR #1277, opened 2024-09-25
Support downloading and installing simulator images with Xcode 16

ivanpovazan:
### Xcode 16 change

From Xcode 16 Apple stopped providing sources for downloading newer simulator runtime images. 

If we consider `watchOS 10.5 Simulator Runtime` and `iOS 18.0 beta Simulator Runtime` from https://devimages-cdn.apple.com/downloads/xcode/simulators/index2.dvtdownloadableindex we can see that source is not provided for the latter:

```xml
<dict>
	<key>authentication</key>
	<string>virtual</string>
	<key>category</key>
	<string>simulator</string>
	<key>contentType</key>
	<string>diskImage</string>
	<key>dictionaryVersion</key>
	<integer>2</integer>
	<key>fileSize</key>
	<integer>3946536779</integer>
	<key>identifier</key>
	<string>com.apple.dmg.WatchSimulatorSDK10_5</string>
	<key>name</key>
	<string>watchOS 10.5 Simulator Runtime</string>
	<key>platform</key>
	<string>com.apple.platform.watchos</string>
	<key>simulatorVersion</key>
	<dict>
		<key>buildUpdate</key>
		<string>21T575</string>
		<key>version</key>
		<string>10.5</string>
	</dict>
	<key>source</key>
	<string>https://download.developer.apple.com/Developer_Tools/watchOS_10.5_Simulator_Runtime/watchOS_10.5_Simulator_Runtime.dmg</string>
	<key>version</key>
	<string>10.5.0.0</string>
</dict>
<dict>
	<key>category</key>
	<string>simulator</string>
	<key>contentType</key>
	<string>cryptexDiskImage</string>
	<key>dictionaryVersion</key>
	<integer>2</integer>
	<key>downloadMethod</key>
	<string>mobileAsset</string>
	<key>fileSize</key>
	<integer>8455760175</integer>
	<key>identifier</key>
	<string>com.apple.dmg.iPhoneSimulatorSDK18_0_b1</string>
	<key>name</key>
	<string>iOS 18.0 beta Simulator Runtime</string>
	<key>platform</key>
	<string>com.apple.platform.iphoneos</string>
	<key>simulatorVersion</key>
	<dict>
		<key>buildUpdate</key>
		<string>22A5282m</string>
		<key>version</key>
		<string>18.0</string>
	</dict>
	<key>version</key>
	<string>18.0.0.1</string>
</dict>
```

The new images which do not have a source for download (in the above case `iOS 18.0 beta Simulator Runtime`) are classified as cryptographically-sealed disk images a.k.a `cryptexDiskImage` via `contentType` property in the XML which we can use to differentiate between the old and the new ones.

### Recommendation

The recommended way of downloading and installing the new simulator runtimes is to use:
```
xcodebuild -downloadPlatform <platform>
```
as specified in: https://developer.apple.com/documentation/xcode/installing-additional-simulator-runtimes#Install-and-manage-Simulator-runtimes-from-the-command-line. However, this approach does not seem to allow specifying a particular version of the simulator runtime, but rather installs the latest one available for a specific Xcode version.

Furthermore, Apple states they will stop hosting images publicly:
```
The simulator runtimes are available on [Apple Developer website](https://developer.apple.com/download/all/?q=Simulator%20Runtime), but will no longer be posted through the website in future updates.
```

### Short-term solution

As a short term solution, we can use the recommended approach to download and install the simulator runtimes through `xcodebuild` if we are using Xcode 16+. 

### Long-term solution

With the recommended approach, currently there does not seem to be a way of specifying a specific version (or the build number) of the simulator runtime to be installed. To overcome this, for our internal testing and backward compatibility, we might want to consider hosting older simulator images on our own, and adapt xharness with functionality to fallback to internal feeds when a specific version is requested. This would be a "in-house" solution and would not be available for public use.

## Changes

This PR implements the `Short-term solution` solution by:
- Adapting the code to use `xcodebuild -downloadPlatform <platform>` for downloading and installing the new simulator runtimes with Xcode 16+.  
- Bumping `mlaunch` version to be compatible with Xcode 16+
- Adding description and test project for testing in scouting queues (queues used for testing new/beta versions of Xcode) 

## Testing

Xcode 16 is currently only available on scouting queue which was tested in: https://github.com/dotnet/xharness/pull/1277/commits/1fd9292b2e8b0cd489bd88c23a2a528ed8125085 and results are available [here](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-xharness-refs-pull-1277-merge-1575e564c89b429f8a/System.Numerics.Vectors.Tests/1/console.cd9d9f8c.log?helixlogtype=result).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -360,3 +360,5 @@ MigrationBackup/
 
 # some ide stuff
 .idea
+
+.vscode/
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -15,7 +15,7 @@
     <PackageVersion Include="Mono.Options" Version="6.12.0.148" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="Selenium.WebDriver" Version="4.0.0-alpha05" />
-    <PackageVersion Include="Microsoft.Tools.Mlaunch" Version="1.0.96" />
+    <PackageVersion Include="Microsoft.Tools.Mlaunch" Version="1.0.256" />
     <PackageVersion Include="NUnit" Version="3.13.0" />
     <PackageVersion Include="NUnit.Engine" Version="3.13.0" />
     <PackageVersion Include="xunit.extensibility.execution" Version="$(XUnitVersion)" />
--- a/azure-pipelines-public.yml
+++ b/azure-pipelines-public.yml
@@ -148,7 +148,7 @@ stages:
   parameters:
     name: E2E_Apple_Simulator_Commands
     displayName: Apple - Simulator Commands
-    testProject: $(Build.SourcesDirectory)/tests/integration-tests/Apple/Simulator.Scouting.Commands.Tests.proj
+    testProject: $(Build.SourcesDirectory)/tests/integration-tests/Apple/Simulator.Commands.Tests.proj
 
 - template: eng/e2e-test.yml
   parameters:
--- a/azure-pipelines-public.yml
+++ b/azure-pipelines-public.yml
@@ -148,7 +148,7 @@ stages:
   parameters:
     name: E2E_Apple_Simulator_Commands
     displayName: Apple - Simulator Commands
-    testProject: $(Build.SourcesDirectory)/tests/integration-tests/Apple/Simulator.Commands.Tests.proj
+    testProject: $(Build.SourcesDirectory)/tests/integration-tests/Apple/Simulator.Scouting.Commands.Tests.proj
 
 - template: eng/e2e-test.yml
   parameters:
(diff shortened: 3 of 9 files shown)


ivanpovazan:
This will have to be ported to net9.0 and net8.0 branches.

rolfbjarne (approved):

dalexsoto (approved):

ivanpovazan:
The `dotnet.xharness (E2E WASM Helix Tests Build_Debug)` seem to be hanging. I will force merge as these changes do not affect WASM.

dalexsoto:
@ivanpovazan should this change be back ported to the .NET 8 and 9 release branches?

ivanpovazan:
@dalexsoto it will for transparency. 

However, FWIW, both Xamarin and MAUI are subscribed only to the main branch changes (through darc) which shouldn't be an issue since all the assemblies that we build here target `NetMinimum` which is `net8.0`

PR #1285, opened 2024-10-02
`WebAssembly` tests run with `Playwright`

ilonatommy:
Following suggestion in https://github.com/microsoft/playwright/issues/32897#issuecomment-2385802923, we would like to shift the responsibility of managing chromium to the testing framework in `WebAssembly` tests. We can do it easily in `Wasm.Build.Tests` where we utilize `Playwright`. However, in library tests we are using `Selenium`. `Selenium` does not provide management over installing and locating the browser.
To keep all the WASM tests running on the same browser version, we would like to switch to `Playwright` in the library tests as well. This PR tries to achieve it, without loosing the original functionalities.


ilonatommy:
Yes, but the original reason was to fix https://github.com/dotnet/runtime/issues/107771 by allowing default chromium in WasmBuildTests. If we do it, we would like to keep browser versions in line across our tests (both `WasmBuildTest` run by `runtime` and library tests run by `xharness`).

ivanpovazan:
Can we close this PR?

PR #1304, opened 2024-10-24
List platforms and commands in docs

kotlarmilos:
## Description

This PR lists platforms and commands in docs.


ivanpovazan (approved):
Thanks!

Issue #1188, opened 2024-03-19
[mobile] [infra] Unify reporting of mobile CI testing into a single console log file


matouskozak:
### Description
Tests run on mobile platforms, don't report all outputs into the console log. This behavior renders using Build Analysis un-reliable for these scenarios because console log doesn't contain all information about the completed test. The full test output can be found inside the device logs.

E.g., console log might contain message "ERROR: WORKLOAD TIMED OUT - Killing user command" but the device log can show more specific errors for the given test.

### Possible solutions
1. Detect error messages from the device log and copy them to the console log
    - Pro: doesn't cause significant increase in console log size
    - Cons: might be difficult to detect all error messages correctly
2. Include both console and device logs in Build Analysis
    - Pro: reliability, contains all error messages from both console and device log
    - Cons: can significantly increase the strain on Build Analysis (some device log are over 300 MB in size)

### TODO:

1. Collect past log files from mobile devices runs to determine which log files are used for reporting actual test failures.
2. From the collected log files, determine pattern that can reliably match error messages. Discuss trade-off between false positives and false negatives, i.e., is it worse if a failure is un-matched or if non-error message gets included.
3. Use the matching pattern from 2. to copy error messages in to the main console log.

matouskozak:
Resolved in https://github.com/dotnet/xharness/pull/1348

PR #1348, opened 2025-01-14
Unify main and app output logs to improve the UX of failure investigation

matouskozak:
# Background
Tests run on mobile platforms, don't report all outputs into the console log. This behavior renders using Build Analysis un-reliable for these scenarios because console log doesn't contain all information about the completed test. The full test output can be found inside the app output logs.

E.g., console log might contain message "ERROR: WORKLOAD TIMED OUT - Killing user command" but the device log can show more specific errors for the given test.

# Proposed Solution
Duplicate the output from app log to the main log. This will increase the size of the console log but will allow Build Analysis to match greater variaty of patterns that are only available as part of app output logs + it speeds up the investigation experience as most of the information will be contained in the single log.

# TODO
- [x] iOS Device: [console log](https://helixr1107v0xdeko0k025g8.blob.core.windows.net/dotnet-xharness-refs-pull-1348-merge-53363fea26d1411393/ios-device-System.Buffers.Tests.app/1/console.36d682eb.log?helixlogtype=result)
- [x] MacCatalyst [console log](https://helixr1107v0xdeko0k025g8.blob.core.windows.net/dotnet-xharness-refs-pull-1348-merge-7fdc27203d984bf381/maccatalyst-System.AppContext.Tests.app/1/console.c5d21d7f.log?helixlogtype=result)
- [x] Android Device: [console log](https://helixr1107v0xdeko0k025g8.blob.core.windows.net/dotnet-xharness-refs-pull-1348-merge-89b2d0275cd8479e86/System.Buffers.Tests-arm64-v8a/1/console.390a3417.log?helixlogtype=result)
- [x] Android Emulator: [console log](https://helixr1107v0xdeko0k025g8.blob.core.windows.net/dotnet-xharness-refs-pull-1348-merge-689b78ee5378477f86/System.Buffers.Tests-x64/1/console.c2ed5b55.log?helixlogtype=result)

# Future work
Remove the need for separate app output logs for android and iOS if we are duplicating their content to the main (console) log anyways.

---

Contributes to https://github.com/dotnet/xharness/issues/1188

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
@@ -190,7 +190,8 @@ public class AppTester : AppRunnerBase, IAppTester
             using var crashLogs = new Logs(_logs.Directory);
 
             ICrashSnapshotReporter crashReporter = _snapshotReporterFactory.Create(_mainLog, crashLogs, isDevice: !isSimulator, device.Name);
-            using ITestReporter testReporter = _testReporterFactory.Create(_mainLog,
+            using ITestReporter testReporter = _testReporterFactory.Create(
+                _mainLog,
                 _mainLog,
                 _logs,
                 crashReporter,
--- a/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppRunnerBase.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppRunnerBase.cs
@@ -327,8 +327,11 @@ public abstract class AppRunnerBase
         }
         else
         {
+            // For MacCatalyst, the test output log does not propagate to the main log.
+            // Hence, we duplicate the log to the main console log to simplify the UX of failure investigation.
+            IFileBackedLog aggregatedAppOutputLog = Log.CreateReadableAggregatedLog(_mainLog, log);
             _processManager
-                .ExecuteCommandAsync("log", logArgs, _mainLog, log, log, TimeSpan.FromDays(1), cancellationToken: streamCancellation)
+                .ExecuteCommandAsync("log", logArgs, _mainLog, aggregatedAppOutputLog, aggregatedAppOutputLog, TimeSpan.FromDays(1), cancellationToken: streamCancellation)
                 .DoNotAwait();
         }
 


src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
```diff
@@ -361,11 +361,15 @@ private async Task RunDeviceTests(
             // We need to check for MT1111 (which means that mlaunch won't wait for the app to exit)
             IFileBackedLog aggregatedLog = Log.CreateReadableAggregatedLog(_mainLog, testReporter.CallbackLog);
 
+            // The app output log is not directly accessible. 
+            // Hence, we duplicate the log to the main console log to simplify the UX of failure investigation
+            IFileBackedLog aggregatedLogAppOutput = Log.CreateReadableAggregatedLog(_mainLog, appOutputLog);
```
matouskozak, line 366: Does it make sense to keep the `appOutputLog` as a separate log file if its content is duplicated to the main log (console log)? We might scratch it out completely, but it would be a bigger change to xharness then.

matouskozak:
@ivanpovazan @akoeplinger What do you think of this solution? I want to know your opinion on duplicating the app output log to the main (console) log before I proceed with finishing the other scenarios.

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
ivanpovazan: I would keep the file for now. 
I don't think it is that expensive to keep it around and not sure if there is a dependency on its existence (I might be wrong). 
As for the changes, I think they are looking good.

akoeplinger: I'd remove it eventually, it's a bit confusing having two files with the same data.

akoeplinger:
sounds good to me. the console logs aren't particularly big right?

matouskozak:
@ivanpovazan @akoeplinger I implemented the log unification for Android (emulator/device), iOS device and MacCatalyst. I don't really like that I had to do it 3 times (once for each platform) but I didn't find a single point where this could be achieved. Please let me know what you think about this.

ivanpovazan:
FWIW, from the provided logs, one thing I noticed on Android is that for emulator logs we seem to enable GC logging while on the device we don't. Not sure why it is configured differently, just thinking that GC logging could introduce a lot of noise in the logs.

ivanpovazan (approved):
LGTM

ivanpovazan:
Let's backport this, so that we have better build analysis on servicing branches

PR #1371, opened 2025-03-13
Stop building Microsoft.DotNet.XHarness.CLI for NetCurrent

akoeplinger:
We don't need to be on the latest TFM (we already set RollForward=major) and it prevents issues when xharness uses a newer SDK than runtime like in https://github.com/dotnet/runtime/pull/113381#issuecomment-2722131488


ivanpovazan (approved):
LGTM, thanks!

PR #1372, opened 2025-03-15
Do some extra work to avoid making a full size copy of the buffer when base64 encoding

lewing:
fixes: https://github.com/dotnet/runtime/issues/113474  (kinda)

### DIFF
--- a/src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
+++ b/src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
@@ -4,7 +4,10 @@
 
 using System;
 using System.IO;
+using System.Runtime.InteropServices;
 using System.Xml.Linq;
+using System.Text;
+using System.Security.Cryptography;
 
 #nullable enable
 
@@ -12,6 +15,7 @@ namespace Microsoft.DotNet.XHarness.TestRunners.Xunit;
 
 internal class WasmXmlResultWriter
 {
+#if !NET || DEBUG
     public static void WriteOnSingleLine(XElement assembliesElement)
     {
         using var ms = new MemoryStream();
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -1,24 +1,119 @@
...
+        // Ensure all data is flushed through the CryptoStream
+        xmlWriter.Flush();
+        cryptoStream.FlushFinalBlock();
+
+        ms.TryGetBuffer(out var bytes);
+        var charData = MemoryMarshal.Cast<byte,char>(bytes);
+
+        // Output the result
+        Console.WriteLine($"STARTRESULTXML {charData.Length} {charData} ENDRESULTXML");
+        Console.WriteLine($"Finished writing {charData.Length} bytes of RESULTXML");
```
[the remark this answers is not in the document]
lewing: Silly copilot I went to all that trouble to avoid making a copy of the string there, the interpolated string handler will happily consume a Span<char>

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -1,24 +1,156 @@
...
+                | 0  | 97 | 0  | 71 | 0  |107 | 0  |103 | 0  | 98 | 0  | 87 | 0  | 57 | 0  |116 |
+                +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
+
+                */
+
+                var base64Written = outputBuffer.AsSpan(base64OutputStart, base64BytesWritten);
+                var outputChars = MemoryMarshal.Cast<byte, char>(outputBuffer.AsSpan(outputOffset + totalBytesWritten, OutputBlockSize));
+                for (int i = 0; i < base64BytesWritten; i++)
+                {
+                    // Expand each acsii byte to a char write it in the same logical position
```
lewing, line 99:
```suggestion
                    // Expand each ascii byte to a char write it in the same logical position
```

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -1,24 +1,154 @@
...
+        // guaranteed to succeed with the MemoryStream() constructor
+        ms.TryGetBuffer(out var bytes);
+        // we went to a lot of trouble to put characters in the final buffer
+        // so that we can avoid a copy here and pass the span directly to the
+        // string interpolation logic.
+        Span<char> charData = MemoryMarshal.Cast<byte,char>(bytes);
+
+        // Output the result
+        Console.WriteLine($"STARTRESULTXML {charData.Length} {charData} ENDRESULTXML");
+        Console.WriteLine($"Finished writing {charData.Length} bytes of RESULTXML");
```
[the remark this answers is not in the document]
lewing: While it looks a little strange since charData is a base64 encoded string and code reading this on the other side is reading it as utf8 it will in fact be charDate.Length bytes long

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -1,24 +1,153 @@
...
+        xmlWriter.Flush();
+        cryptoStream.FlushFinalBlock();
+
+        // guaranteed to succeed with the MemoryStream() constructor
+        ms.TryGetBuffer(out var bytes);
+        // we went to a lot of trouble to put characters in the final buffer
+        // so that we can avoid a copy here and pass the span directly to the
+        // string interpolation logic.
+        Span<char> charData = MemoryMarshal.Cast<byte,char>(bytes);
+
```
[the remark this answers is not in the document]
lewing: This suggestion is both incorrect and would not compile...

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -1,24 +1,153 @@
...
+                inputBuffer, inputOffset, inputCount,
+                outputBuffer, base64OutputStart);
+
+            var base64Written = outputBuffer.AsSpan(base64OutputStart, base64BytesWritten);
+            var outputChars = outputBuffer.AsSpan(outputOffset, OutputBlockSize);
+            for (int i = 0; i < base64BytesWritten; i++)
+            {
+                // Expand each ascii byte to a char write it in the same logical position
+                // as a char in outputChars eventually filling the output buffer
+                BitConverter.TryWriteBytes(outputChars.Slice(i * 2), (char)base64Written[i]);
```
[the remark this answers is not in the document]
lewing: Copilot, I'm counting on these being Chars not a specific version of utf16.

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -1,24 +1,153 @@
...
+        xmlWriter.Flush();
+        cryptoStream.FlushFinalBlock();
+
+        // guaranteed to succeed with the MemoryStream() constructor
+        ms.TryGetBuffer(out var bytes);
+        // we went to a lot of trouble to put characters in the final buffer
+        // so that we can avoid a copy here and pass the span directly to the
+        // string interpolation logic.
+        Span<char> charData = MemoryMarshal.Cast<byte,char>(bytes);
+
```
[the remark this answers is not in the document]
lewing: There is an implicit cast operator

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -1,24 +1,153 @@
...
```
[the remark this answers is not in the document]
lewing: this isn't needed but meh

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -1,24 +1,160 @@
...
+internal class WasmXmlResultWriter
+{
+#if DEBUG
+    public static void WriteOnSingleLine(XElement assembliesElement)
+    {
+        using var ms = new MemoryStream();
+        assembliesElement.Save(ms);
+        ms.TryGetBuffer(out var bytes);
+        var base64 = Convert.ToBase64String(bytes, Base64FormattingOptions.None);
```
pavelsavara, line 24: I think that base64 could contain `\n` in the middle right ? If so, you can insert some which would make the buffer in emscripten smaller. Because `stdout` emulator would build buffer until newline and only then write it into browser `console.log`.

 @maraf noted that maybe we had some issues with message order WRT `WASM EXIT` ... so I don't know if multi-line is good idea or not.

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -1,24 +1,160 @@
...
+
+        // guaranteed to succeed with the MemoryStream() constructor
+        ms.TryGetBuffer(out var bytes);
+        // we went to a lot of trouble to put characters in the final buffer
+        // so that we can avoid a copy here and pass the span directly to the
+        // string interpolation logic.
+        Span<char> charData = MemoryMarshal.Cast<byte, char>(bytes.AsSpan());
+
+        // Output the result and the the ascii length of the data
+        Console.WriteLine($"STARTRESULTXML {charData.Length} {charData} ENDRESULTXML");
```
lewing, line 156:
```suggestion
        Console.Write($"STARTRESULTXML {charData.Length} ");
        Console.Write(charData)
        Console.WriteLine(" ENDRESULTXML");
```

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
lewing: We could in theory write each block as we transform it rather than create a single buffer. We should really review the emscripten line buffering.

pavelsavara: Alternatively, this could be `fetch('POST', xml)` in browser and nodejs. 
Remaining is `v8` shell which doesn't have file write. So we may need to keep this contraption.

pavelsavara: I really thing we should implement it as POST. Do we run V8 in CI at all ?

akoeplinger: > Do we run V8 in CI at all ?

yes, we noticed it in https://github.com/dotnet/runtime/pull/102069#issuecomment-2107048126. it'd be nice to fix that though.

pavelsavara: https://github.com/dotnet/xharness/pull/1388

Issue #1382, opened 2025-03-24
[apple mobile] [infra] Unify Apple mobile device logs for dotnet/runtime runtime tests logs


matouskozak:
The https://github.com/dotnet/xharness/pull/1348 was supposed to unify app and console logs to improve UX experience of failure investigation. However, currently on iOS/tvOS devices the app test logs for dotnet/runtime runtime test logs are not included in the main console log and instead remain in the `net.dot.<test_app_name>.log`.

matouskozak:
Looking at the dotnet/runtime logs:

Runtime test cmd:
```
+ sudo launchctl asuser 504 sudo -u helix-runner sh ./xharness-runner.apple.sh --target tvos-device --command-timeout 11280 --timeout 03:00:00 --launch-timeout 00:05:00 --includes-test-runner --expected-exit-code 0 --app /tmp/helix/working/A8FA09D2/w/9FA008F4/e/Exceptions.app --output-directory /tmp/helix/working/A8FA09D2/w/9FA008F4/uploads
```

Libraries test cmd:
```
+ sudo launchctl asuser 504 sudo -u helix-runner sh ./xharness-runner.apple.sh --target tvos-device --command-timeout 4980 --timeout 01:15:00 --launch-timeout 00:05:00 --includes-test-runner --expected-exit-code 0 --app /tmp/helix/working/AC4408EF/w/9CE508C0/e/System.Collections.Immutable.Tests.app --output-directory 
```

Look identical...

matouskozak:
I'm wondering if https://github.com/dotnet/runtime/pull/107118 might have changed the output mechanism

matouskozak:
The work in xharness was done, remaining work is needed in dotnet/runtime to switch to `apple test` again.

PR #1383, opened 2025-03-24
[apple] [infra] Refactor unification of app and main logs for iOS devices and MacCatalyst

matouskozak:
Refactor unification of app and main logs for iOS devices and MacCatalyst :
- Move the logic of unifying the app and main logs to the TestOrchestrator.
- Copy system and application logs to the main log after the app has finished.

Example of refactored reporting: https://helixr1107v0xdeko0k025g8.blob.core.windows.net/dotnet-xharness-refs-pull-1383-merge-253e93d337aa4abdbb/tvos-device-System.Buffers.Tests.app/1/console.f4e82d5e.log?helixlogtype=result

---

Possibly fixing https://github.com/dotnet/xharness/issues/1382


src/Microsoft.DotNet.XHarness.Apple/Orchestration/TestOrchestrator.cs
```diff
@@ -364,4 +373,42 @@ ExitCode LogProblem(string message, ExitCode defaultExitCode)
...
+                {
+                    while (!reader.EndOfStream)
+                    {
+                        logContent = reader.ReadLine()!;
+                        if (logContent != null)
+                        {
+                            _mainLog.WriteLine(logContent);
+                        }
+                    }
+                }
```
ivanpovazan, line 403: nit:
```suggestion
                using var reader = log.GetReader();
                while (!reader.EndOfStream)
                {
                    var logContent = reader.ReadLine();
                    if (logContent is null)
                        continue;
                    _mainLog.WriteLine(logContent);
                }
```

ivanpovazan (approved):
I think this is OK and thanks for the fix.

One thing to consider is putting the log merging behind a CLI parameter which is not enabled by default, but rather from our testing infra. 
Yes, that is more work, but what I found "annoying" is that now when you run a test/app from command line on Android the terminal gets polluted with the whole adb logcat output. From UX perspective we don't want that and should at least provide a way to turn it off.

/cc: @akoeplinger for thoughts

matouskozak:
> One thing to consider is putting the log merging behind a CLI parameter which is not enabled by default, but rather from our testing infra. Yes, that is more work, but what I found "annoying" is that now when you run a test/app from command line on Android the terminal gets polluted with the whole adb logcat output. From UX perspective we don't want that and should at least provide a way to turn it off.

That's a very good point I one that I haven't considered much before. I think we could add it as `apple test` / `android test` argument. I'll open a follow-up issue to address this as I would like to get this change in asap to verify that it fixes the runtime test logging issue.

matouskozak:
fyi: moved the log merging after the `ParseResult` call to not affect the logs before they get parsed.

akoeplinger (approved):
LGTM. I agree we should add the CLI option for opting-in

Issue #1376, opened 2025-03-17
MacCatalyst tests apps don't run on macOS 15, possibly missing some required entitlements


matouskozak:
When testing possibly of migrating to newer macOS queue https://github.com/dotnet/xharness/pull/1368#issuecomment-2715086406, we encountered an issue where the precompiled MacCatalyst `System.AppContext.Tests.app` and `iOS.Simulator.PInvoke.Test.app` don't run on macOS 15 with error message:

```
_LSOpenURLsWithCompletionHandler() failed with error -54.
```

This could mean that the app is missing some required entitlements or might have to be rebuilt with newer XCode to work properly on newer macOS.

The queue seem to be configured properly as the dotnet/runtime MacCatalyst jobs are running as expected https://dev.azure.com/dnceng-public/public/_build/results?buildId=982758&view=results.

matouskozak:
fyi: @rolfbjarne @ivanpovazan

akoeplinger:
Hmm do you only see this on arm64 runs? the precompiled app is x64 only so the new mac might be missing Rosetta.

matouskozak:
> Hmm do you only see this on arm64 runs? the precompiled app is x64 only so the new mac might be missing Rosetta.

Both x64 and arm64 have the same issue.

rolfbjarne:
What happens if you launch the executable directly, instead of using `open ...`?

So instead of this:

> open -n -W /tmp/helix/working/99EA0909/w/A68008D5/e/iOS.Simulator.PInvoke.Test.app

do this:

> /tmp/helix/working/99EA0909/w/A68008D5/e/iOS.Simulator.PInvoke.Test.app/Contents/MacOS/iOS.Simulator.PInvoke.Test

Also I'm assuming the name of the test app ("iOS.Simulator") doesn't mean anything?

matouskozak:
> What happens if you launch the executable directly, instead of using `open ...`?
> 
> So instead of this:
> 
> > open -n -W /tmp/helix/working/99EA0909/w/A68008D5/e/iOS.Simulator.PInvoke.Test.app
> 
> do this:
> 
> > /tmp/helix/working/99EA0909/w/A68008D5/e/iOS.Simulator.PInvoke.Test.app/Contents/MacOS/iOS.Simulator.PInvoke.Test

This works and the app is executed as expected (verified locally).

> Also I'm assuming the name of the test app ("iOS.Simulator") doesn't mean anything?

Yes, it is just a name I believe.

[...]

PR #1384, opened 2025-03-24
[maccatalyst] New signed test app for MacCatalyst (`System.Collections.NonGeneric.Tests`)

matouskozak:
On newer macOS version, MacCatalyst apps must be signed to be able tu run. 

In this PR, we add a new signed test app, `System.Collections.NonGeneric.Tests.app` taken from dotnet/runtime CI and compressed using `tar -czf` to create a gzip archive. The compressed archive is uploaded as `System.Collections.NonGeneric.Tests.app.zip` to the remote storage.

Removing the old app `maccatalyst-System.AppContext.Tests.app` as there isn't 1:1 signed replacement which we could use after we migrate to newer osx queues.

---

https://github.com/dotnet/xharness/issues/1376

### DIFF
--- a/tests/integration-tests/Apple/Simulator.Scouting.Tests.proj
+++ b/tests/integration-tests/Apple/Simulator.Scouting.Tests.proj
@@ -15,7 +15,7 @@
 
     <!-- apple test / maccatalyst -->
     <XHarnessAppleProject Include="TestAppBundle.proj">
-      <AdditionalProperties>TestTarget=maccatalyst;TestAppBundleName=System.AppContext.Tests.app</AdditionalProperties>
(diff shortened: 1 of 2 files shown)


matouskozak:
> Can we just remove the `maccatalyst-System.AppContext.Tests.app` in this PR, as it seems like a replacement? 

Sure, I can remove it here if that's preferred.

> Additionally, what about the other: `iOS.Simulator.PInvoke.Test.app` is that one not causing signing issues?

It does as well, that one is going to be replaced with the same app but signed. The replacement happens at the storage level so no need for change in xharness codebase.

matouskozak:
Closing in favor of https://github.com/dotnet/runtime/pull/113313

PR #1388, opened 2025-03-31
[browser] HTTP POST test results when possible

pavelsavara:
- use HTTP POST when in available
- always single line otherwise
- `WebServerUploadResults` option, default true
- async `WriteResults`

Contributes to https://github.com/dotnet/runtime/issues/113474

### DIFF
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
@@ -184,7 +184,7 @@ public class WebServer
                     {
                         var xmlResultsFilePath = Path.Combine(options.OutputDirectory ?? Directory.CreateTempSubdirectory().FullName, "testResults.xml");
                         await using var fileStream = File.Create(xmlResultsFilePath);
-                        await request.BodyReader.CopyToAsync(fileStream);
+                        await context.Request.BodyReader.CopyToAsync(fileStream);
                         _logger.LogInformation($"Stored {xmlResultsFilePath} results {fileStream.Position} bytes");
                     });
                 });
--- a/src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
+++ b/src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
@@ -182,7 +182,7 @@ public class WebServer
                 {
                     router.MapPost("/test-results", async context =>
                     {
-                        var xmlResultsFilePath = Path.Combine(options.OutputDirectory, "testResults.xml");
+                        var xmlResultsFilePath = Path.Combine(options.OutputDirectory!, "testResults.xml");
 
                         await using var fileStream = File.Create(xmlResultsFilePath);
                         await context.Request.BodyReader.CopyToAsync(fileStream);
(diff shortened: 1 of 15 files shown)


pavelsavara:
image
image

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
```diff
@@ -175,6 +176,20 @@ public void Configure(IApplicationBuilder app, IOptionsMonitor<TestWebServerOpti
...
+            if (options.WebServerUploadResults)
+            {
+                app.UseRouter(router =>
+                {
+                    router.MapPost("/test-results", async context =>
+                    {
+                        var xmlResultsFilePath = Path.Combine(options.OutputDirectory ?? Directory.CreateTempSubdirectory().FullName, "testResults.xml");
+                        using var fileStream = new FileStream(xmlResultsFilePath, FileMode.Create);
+                        await context.Request.Body.CopyToAsync(fileStream);
+                        await fileStream.FlushAsync();
```
akoeplinger, line 188: nit: [docs](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-9.0#read-the-request-body) suggest this:
```suggestion
                        await using var fileStream = File.Create(xmlResultsFilePath);
                        await request.BodyReader.CopyToAsync(fileStream);
```

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/WasmXmlResultWriter.cs
```diff
@@ -12,10 +16,31 @@ namespace Microsoft.DotNet.XHarness.TestRunners.Xunit;
...
+            req.Content.Headers.ContentType = new MediaTypeHeaderValue("application/xml");
+            req.Content.Headers.ContentLength = ms.Length;
+
+            using var httpClient = new HttpClient();
+            using var response = await httpClient.SendAsync(req);
+            if (response.IsSuccessStatusCode)
+            {
+                Console.WriteLine($"Finished uploading {ms.Length} bytes of RESULTXML");
+                return;
+            }
```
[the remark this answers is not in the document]
lewing: in what cases will we fall back?

lewing (approved):

src/Microsoft.DotNet.XHarness.CLI/Commands/WebServer.cs
```diff
@@ -175,6 +176,19 @@ public void Configure(IApplicationBuilder app, IOptionsMonitor<TestWebServerOpti
                     });
                 });
             }
+            if (options.WebServerUploadResults)
+            {
+                app.UseRouter(router =>
+                {
+                    router.MapPost("/test-results", async context =>
+                    {
+                        var xmlResultsFilePath = Path.Combine(options.OutputDirectory ?? Directory.CreateTempSubdirectory().FullName, "testResults.xml");
```
akoeplinger, line 185: we shouldn't need to handle OutputDirectory being null since it's a required CLI option
```suggestion
                        var xmlResultsFilePath = Path.Combine(options.OutputDirectory, "testResults.xml");

```

PR #1445, opened 2025-07-23
Skip TCP launch timeout if result files are copied directly

kotlarmilos:
## Description

On iOS 18 and later (devices and simulators), result files are copied directly from the app container instead of being transmitted over TCP. In these cases, the TCP launch timeout logic is skipped to avoid terminating tests when no TCP connection is established.

## Changes

 - Skip TCP launch timeout logic for Apple mobile platforms on iOS 18+
 - Extend resultFileHandler with an IsVersionSupported method
 - Update resultFileHandler to throw an exception if the version format is invalid

Contributes to https://github.com/dotnet/runtime/issues/117714

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -53,6 +53,8 @@ extends:
                 name: Build
                 displayName: Build
                 condition: succeeded()
+
+              - template: /eng/pipelines/apiscan-steps.yml
             templateContext:
               outputs:
               - output: pipelineArtifact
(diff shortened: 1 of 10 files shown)


rolfbjarne:
> > Process mlaunch exited with 42
> 
> That test is supposed to be exiting with 42 so it looks like mlaunch is somehow using that as its own exit code

Yes, that's correct, mlaunch will forward the exit code if it can.

PR #1477, opened 2025-10-07
Add network isolation policy to pipeline settings

kotlarmilos:
## Description

Update network isolation policy to block access to all public package feed endpoints.


kotlarmilos:
This blocks Android platform tools download, let me reach out to the corresponding team.

akoeplinger:
note that we don't need to set this explicitly. you'll get opted in to CFSClean automatically after 7 days of clean runs as per the docs

Issue #1135, opened 2024-01-12
[enhancement] Coverage Collection


amirvenus:
Hi,

I have been using this fantastic library with xunit however, I have noted that the test coverage results are not collected/published.

I would be grateful if consideration is given to collecting test coverage data as part of the test execution so that they could be used in a CI/CD run. 

Thanks!

akoeplinger:
I assume you're looking at getting some .xml or other file produced by the coverage tooling from the device right?
Which tool are you using? And on which platform?

amirvenus:
> I assume you're looking at getting some .xml or other file produced by the coverage tooling from the device right? Which tool are you using? And on which platform?

The idea is something like this: 
https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-code-coverage?tabs=linux
or dotCover that in addition to the actual results of the tests (and stacktraces where applicable) it will also collect the code coverage i.e. covered/uncovered percentage/lines/statements etc. 

I am using this tool on a macOS azdo self-hosted agent for a CI/CD pipeline. 

Thanks!

akoeplinger:
Yeah so there are essentially two steps: 1) plug the coverage collector into the test run 2) collect the coverage results xml/json.

The former is probably something you can do already in your runner class which is derived from iOSApplicationEntryPoint.
The latter is a bit more complicated since there's no good way to get a file off of the device, we currently stream the nunit/xunit test results over the network: https://github.com/dotnet/xharness/blob/main/src/Microsoft.DotNet.XHarness.TestRunners.Common/iOSApplicationEntryPointBase.cs#L19-L21

amirvenus:
Actually I am using the TestRunners from the Maui repo with some slight modifications used with xharness in a CI/CD pipeline. 

I only use net6.0-android (not maui) for my project so I have deleted anything non-android (iOS, macCatalyst, etc.) 

I can see that they have this:

 
```
internal class HeadlessTestRunner : AndroidApplicationEntryPoint
{
    private readonly HeadlessRunnerOptions _runnerOptions;
    private readonly TestOptions _options;

    public HeadlessTestRunner(HeadlessRunnerOptions runnerOptions, TestOptions options)
    {
        _runnerOptions = runnerOptions;
        _options = options;

        var cache = Application.Context.CacheDir!.AbsolutePath;
        TestsResultsFinalPath = Path.Combine(cache, _runnerOptions.TestResultsFilename);
    }

    protected override bool LogExcludedTests => true;

[...]

PR #1565, opened 2026-03-15
Add --enable-coverage flag for code coverage collection from device tests

amirvenus:
## Summary

Adds an optional `--enable-coverage` CLI flag across all platforms that enables code coverage collection during XHarness device test execution. When enabled, XHarness generates a Cobertura XML coverage report and places it in the output directory alongside test results.

**No external coverage tools (coverlet, etc.) are required.** CoverageManager generates method-level coverage by reflecting over the loaded test assemblies. Standard tools like `coverlet.msbuild` cannot instrument assemblies for device builds (APK/app bundles) since they only hook into the `dotnet test` pipeline — CoverageManager solves this by providing built-in coverage that works on all platforms.

If an external tool has already produced a coverage file at the expected path, CoverageManager uses that file instead.

Addresses #1135.

### What's included

- **`--enable-coverage` CLI flag** on all platform commands:
  - `android test`, `android run`
  - `apple test`, `apple just-test`
  - `wasm test`, `wasm test-browser`

- **Test Runner integration** (`TestRunners.Common`):
  - `NUNIT_ENABLE_COVERAGE` / `NUNIT_COVERAGE_OUTPUT_PATH` environment variable support in `ApplicationOptions`
  - `CoverageManager` that generates method-level Cobertura XML via reflection over loaded test assemblies (no external dependencies)
  - Coverage lifecycle integrated into `ApplicationEntryPoint.InternalRunAsync()`

- **Per-platform transport**:
  - **Android**: Coverage flag passed as instrumentation argument → test app bridges to env var → `CoverageManager` generates report → path reported via `INSTRUMENTATION_RESULT: coverage-results-path=<path>` → `InstrumentationRunner` pulls file from device via `adb`
  - **Apple (iOS/tvOS/macOS)**: `NUNIT_ENABLE_COVERAGE` injected as env var via the orchestrator → coverage file written to app's Documents directory → pulled alongside test results
  - **WASM**: Coverage data emitted on stdout using `STARTCOVERAGEXML <len> <base64> ENDCOVERAGEXML` markers → `WasmTestMessagesProcessor` decodes and writes to output directory

- **`ThreadlessXunitTestRunner` made public**: Enables Android test apps to use the runner that works with .NET Android's assembly store (in-memory test discovery via `ReflectionAssemblyInfo`, no `.dll` file on disk required)

### Usage

```bash
# Android
xharness android test --app=test.apk --package-name=com.example --enable-coverage --output-directory=./out

# iOS
xharness apple test --app=TestApp.app --target=ios-simulator-64 --enable-coverage --output-directory=./out

# WASM
xharness wasm test --enable-coverage --output-directory=./out -- test.js
```

After the run, `coverage.cobertura.xml` appears in the output directory. The file uses standard Cobertura format consumable by Azure DevOps, Codecov, ReportGenerator, etc.

## Test plan

- [x] Unit tests for `CoverageManager` (8 tests) — constructor, path resolution, env var setting, reflection coverage generation, fallback behavior
- [x] All existing unit tests pass (97 TestRunners + 21 Android = 118 tests)
- [x] **Android end-to-end**: Built test APK with `Android.OS.Build` API tests → deployed to emulator via locally-built XHarness CLI with `--enable-coverage` → 5/5 tests passed → `coverage.cobertura.xml` (2,741 bytes) pulled from device to output directory
- [x] **iOS end-to-end**: Built iOS app with `UIKit.UIDevice` API tests → ran on iPhone 11 Pro simulator (iOS 26.2) via locally-built XHarness CLI with `--enable-coverage` → 6/6 tests passed → `coverage.cobertura.xml` generated in app container
- [x] CLI builds successfully for all platforms

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vitek-karas:
I looked through this and there are two parts, the first one is copying a specific XML file from the device. That one seems useful, although it might be beneficial to look into support some more generic way of doing this. But if we start with just this file I guess it's OK as well.
The second part is weird, it seems to generate a coverage report from just assembly metadata - so it's not really coverage. I'm, curious what this is for - could you please explain the purpose of it?

vitek-karas:
I think that's the right thing to do to start with - let's just add the ability to download the coverage file from the device and include it in the output. Also please take a look at https://github.com/dotnet/xharness/pull/1579 and ideally integrate the code coverage with it - by including the downloaded file in the file collection reported by the new structured log.

src/Microsoft.DotNet.XHarness.TestRunners.Common/ApplicationEntryPoint.cs
```diff
@@ -206,11 +211,36 @@ private async Task<TestRunner> InternalRunAsync(LogWriter logger)
...
+            logger.Info("Code coverage enabled. Coverage output path: " + coverageManager.OutputPath);
+        }
+
         // notify the clients we are starting
         TestsStarted?.Invoke(this, new EventArgs());
 
         await runner.Run(testAssemblies).ConfigureAwait(false);
 
+        // Check for coverage results if enabled and all tests passed
+        if (coverageManager != null && runner.FailedTests == 0)
```
vitek-karas: `CoverageResultPath` is only assigned inside the `runner.FailedTests == 0` branch. On a failing run, the Android/WASM handoff paths never see a coverage file path even if the file already exists on disk.

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
```diff
@@ -372,6 +372,13 @@ await resultFileHandler.CopyCrashReportAsync(
                         isSimulator: true);
                 }
             }
+
+            // Try to copy coverage results (non-fatal if not present)
+            var coverageDest = Path.Combine(_logs.Directory, "coverage.cobertura.xml");
+            if (await resultFileHandler.CopyCoverageResultsAsync(runMode, true, simulator.OSVersion, simulator.UDID, appInformation.BundleIdentifier, coverageDest))
```
vitek-karas: This copy attempt runs on every Apple test run, not only when `--enable-coverage` was passed. If `coverage.cobertura.xml` is present, it will still be copied and later reported as a produced file regardless of the CLI flag.

src/Microsoft.DotNet.XHarness.iOS.Shared/ResultFileHandler.cs
```diff
@@ -133,6 +133,46 @@ await _processManager.ExecuteCommandAsync(
...
+        RunMode runMode,
+        bool isSimulator,
+        string osVersion,
+        string udid,
+        string bundleIdentifier,
+        string hostDestinationPath)
+    {
+        // Coverage file path mirrors the value set in AppleTestCommand/AppleJustTestCommand
+        string sourcePath = runMode == RunMode.iOS
+            ? "/Documents/coverage.cobertura.xml"
```
vitek-karas: The source path is hard-coded here, while the app side is configured with the relative value `coverage.cobertura.xml` via `NUNIT_COVERAGE_OUTPUT_PATH`. The writer path and the pull path are resolved independently, so they are not guaranteed to match.

src/Microsoft.DotNet.XHarness.TestRunners.Xunit/ThreadlessXunitTestRunner.cs
```diff
@@ -15,7 +15,7 @@
 #nullable enable
 namespace Microsoft.DotNet.XHarness.TestRunners.Xunit;
 
-internal class ThreadlessXunitTestRunner : CustomXunitTestRunner
+public class ThreadlessXunitTestRunner : CustomXunitTestRunner
```
vitek-karas: This changes the type from `internal` to `public`, which widens the public API surface. None of the coverage-file collection changes in this PR add a new in-repo reference that depends on this visibility change.

vitek-karas (approved):

PR #1584, opened 2026-04-07
Add WASM app verbosity compatibility switch

vitek-karas:
Add --no-app-verbosity to 'wasm test-browser' so callers can keep xharness logging verbosity without forwarding '-verbosity <level>' into the app arguments.

This restores a backward-compatible option for the .NET 8.0 runtime tests, where the forwarded verbosity arguments break older WASM browser tests that validate their command-line args strictly.

This is to unblock https://github.com/dotnet/runtime/pull/126322 - which brings latest xharness to the 8.0 servicing branch.
My thinking was that it's simpler to add a backward compat switch to xharness than to backport relatively large test change in the WASM space.


kotlarmilos:
A faster way to validate and fix this in dotnet/runtime would be to add a case like 
```
case "-verbositry": break; // no-op
```
here: https://github.com/dotnet/runtime/blob/07260fe3bc96732ed7f2923ce4852c36a972f112/src/libraries/Common/tests/WasmTestRunner/WasmTestRunner.cs#L52-L53

PR #1605, opened 2026-06-01
Add ci-scan agentic workflow

kotlarmilos:
Adds a gh-aw workflow that runs every 6 hours, walks AzDO definition 72 (dotnet.xharness) on main, classifies failures across the Apple/Android/WASM helix matrix and build legs, applies a follow-up-build presence gate to filter transient regressions, dedups against open Known Build Error issues, and files structured KBEs with literal and regex substrings that Build Analysis can match against PR CI. Agent runs read-only; writes go through capped, allowlisted safe-outputs (max 3 issues per run; only Known Build Error and ancillary platform labels allowed).


kotlarmilos:
Superseded by #1606 which consolidates all three workflows into one PR.

PR #1612, opened 2026-06-10
[runtime-failure-observer] Inline curl calls and require fetched evidence before opening PRs

kotlarmilos:
## Description

The `runtime-failure-observer` agent ran non-deterministically because the copilot harness authorizes a shell command only by its first token, yet the prompt told the agent to pre-bind URLs and loop with `for`. Those forms were denied with `Permission denied` even though the firewall already allows `.dev.azure.com`/`.helix.dot.net`. 

This PR rewrites rule 11 to require every command to start with an allow-listed program and inline each URL into a single `curl ... -o file`.

### DIFF
--- a/.github/workflows/runtime-failure-observer.agent.md
+++ b/.github/workflows/runtime-failure-observer.agent.md
@@ -80,13 +80,12 @@ The agent reads `dotnet/runtime` and the failing build logs. It never writes to 
 3. **Every PR title starts with `[runtime-observer] `.** PRs are opened as drafts.
 4. **Small-fix bounds for complete autofix PRs.** A *complete* fix PR must satisfy all of: `<=` 30 changed lines total, `<=` 2 files (one source + one test), no new public API, no protocol change, no native code change. If the fix needs more, do not silently truncate it: open a clearly-marked best-effort/diagnosability **draft** PR (Step 5) that a human finishes. Best-effort and diagnosability draft PRs may exceed these bounds but must be marked work-in-progress and must still avoid new public API, protocol changes, and native code.
 5. **Don't propose fixes for runtime test bugs.** If the failure is in the test binary itself (assertion in the test code, missing mock, runtime API regression), record `skipped: runtime-side issue` and emit nothing.
-6. **Never assume.** Cite the runtime build URL, the Helix work item URL, the xharness command line, and the exact stderr / exit code in every PR body.
-6b. **No PR without fetched evidence.** Every PR must be backed by a build timeline and a Helix console that you actually downloaded *this run*. If any required fetch (build list, build timeline, Helix work-item list, console log) failed, was empty, or was denied, emit nothing for that candidate. Never reconstruct a build id, AzDO URL, Helix job GUID, console URL, exit code, or stderr excerpt from memory, from the prompt, or by inference. If you did not download it this run, you may not cite it.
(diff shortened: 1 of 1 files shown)


vitek-karas (approved):

.github/workflows/runtime-failure-observer.agent.md
```diff
@@ -118,42 +118,49 @@ These exit codes from `src/Microsoft.DotNet.XHarness.Common/CLI/ExitCode.cs` are
...
+```bash
+curl -s "https://dev.azure.com/dnceng-public/public/_apis/build/builds?definitions=154&branchName=refs/heads/main&statusFilter=completed&resultFilter=failed,partiallySucceeded&%24top=1&api-version=7.1" -o /tmp/gh-aw/agent/preflight.json
+jq -r '.count' /tmp/gh-aw/agent/preflight.json
+```
+
+Valid JSON: continue. If `curl` itself is denied, that is the harness rejecting the command form, not the firewall (`.dev.azure.com` and `.helix.dot.net` are allow-listed). If an inlined first-token `curl` still fails, record `skipped: harness denied inlined curl to dev.azure.com; firewall already allows it` and stop. Never blame the firewall allowlist and never open a PR.
+
 ## Step 1. Set up
 
+Run one inlined `curl` per definition id in `154 223 224 225 226 228 260 261 265`, substituting the id in the URL and the `-o` path:
```
matouskozak, line 134: Just curious, what's the motivation for moving from `for` cycle to just verbal definition? 

We could probably do something like `printf '%s\n' 154 223 224 225 226 228 260 261 265 | xargs -I{} curl -s "https://dev.azure.com/dnceng-public/public/_apis/build/builds?definitions={}&branchName=refs/heads/main&statusFilter=completed&resultFilter=failed,partiallySucceeded&%24top=10&api-version=7.1" -o "/tmp/gh-aw/agent/builds-{}.json"` to keep the loop in instructions. 

I don't have much experience with writting the agentic workflows so feel free to disregard the suggestion.

vitek-karas: The problem is that the check looks only at the first command (basically first word) - if it's not in the allowed list it will fail. We could add 'for' as allowed, but it would defeat the purpose of the list - we don't want the agent to runs scripts which are not allowed...and with for in the allowed list it would be easy to run "anything".
It is not a perfect system at all, but we can at least try :-)

matouskozak: Ok, so if we would allow `for` that would automatically allow also all the inner commands inside that loop, irrespective of what we have in the allowed tools, correct?

If that is true, that would also make the combination of `printf` and `xargs`, both allowed tools, potentially very powerful (dangerous).

btw. is there any documentation on how exactly the checking for allowed tools work? I only found https://github.github.com/gh-aw/reference/tools/ which doesn't give much insight.

kotlarmilos: Yes, the check only looks at the first word of each command, so  for or xargs can run anything inside them. That's why I replaced the loop with explicit  curl  calls, and an additional safety net is the network allow-list plus  read-all  permissions.

kotlarmilos:
re-approval needed

vitek-karas (approved):

PR #1614, opened 2026-06-10
Bump mlaunch to 1.1.127 to support Xcode 27 simulators

dalexsoto:
Xcode 27 removed the standalone Simulator.app; DeviceHub.app is its replacement as the simulator UI host. XHarness boots and launches apps on simulators via 'mlaunch --launchsimulator' / '--launchsimbundleid', which in older mlaunch opens Simulator.app and waits for the 'com.apple.iphonesimulator.ready' notification before booting the device. On Xcode 27 that app no longer exists, so the device never boots and simulator runs fail before any test executes.

mlaunch 1.1.127 adds DeviceHub support (gated on Xcode >= 27), so it boots simulators through DeviceHub on Xcode 27 while keeping the existing Simulator.app behavior on older Xcode versions.


vitek-karas (approved):

akoeplinger:
/ba-g android failures can't be related as this only changes apple assets

PR #1622, opened 2026-06-18
Isolate agentic workflow authentication

kotlarmilos:
Updates both agentic workflows to authenticate via a dedicated, isolated environment rather than the shared default token, tightening the security boundary around agentic workflows.
**Changes**
- Both `*.agent.md`: scoped to the isolated environment with a dedicated token override.
- Regenerated both `*.agent.lock.yml`.

### DIFF
--- a/.github/workflows/runtime-failure-observer.agent.md
+++ b/.github/workflows/runtime-failure-observer.agent.md
@@ -13,6 +13,13 @@ on:
   workflow_dispatch:
   roles: [admin, maintain, write]
 
+environment: copilot-pat-pool
+
+engine:
+  id: copilot
+  env:
+    COPILOT_GITHUB_TOKEN: ${{ secrets.COPILOT_PAT }}
+
 if: github.repository == 'dotnet/xharness'
 
 timeout-minutes: 60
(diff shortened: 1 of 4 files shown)


vitek-karas (approved):

.github/workflows/runtime-failure-observer.agent.md
```diff
@@ -118,42 +125,49 @@ These exit codes from `src/Microsoft.DotNet.XHarness.Common/CLI/ExitCode.cs` are
...
+```bash
+curl -s "https://dev.azure.com/dnceng-public/public/_apis/build/builds?definitions=154&branchName=refs/heads/main&statusFilter=completed&resultFilter=failed,partiallySucceeded&%24top=1&api-version=7.1" -o /tmp/gh-aw/agent/preflight.json
+jq -r '.count' /tmp/gh-aw/agent/preflight.json
+```
+
+Valid JSON: continue. If `curl` itself is denied, that is the harness rejecting the command form, not the firewall (`.dev.azure.com` and `.helix.dot.net` are allow-listed). If an inlined first-token `curl` still fails, emit `report_incomplete` with reason `harness denied inlined curl to dev.azure.com; firewall already allows it` and stop. Never blame the firewall allowlist and never open a PR.
+
 ## Step 1. Set up
 
+Run one inlined `curl` per definition id in `154 223 224 225 226 228 260 261 265`, substituting the id in the URL and the `-o` path:
```
matouskozak: Is it intentional that this PR contains changes from https://github.com/dotnet/xharness/pull/1612 which is not yet merged.

kotlarmilos: Good catch, thanks! I accidentally rebased on the previous PR. Dropped

vitek-karas (approved):

matouskozak (approved):

PR #1630, opened 2026-07-05
Report wall-clock device log diagnostic as warning

vitek-karas:
## Summary

This updates Apple device log capture so the known `Wall Clock adjustment detected - results might be strange while using --end` diagnostic is reported as a warning when it is the only stderr output from log reading.

The main reason for the change is that this message is non-fatal, but the current `Errors while reading device logs:` wording makes it easy for people and tooling consuming the logs to treat it as the failure reason.

The change keeps the existing error path for any other stderr content and adds tests to cover:

- warning-only stderr
- mixed warning + error stderr
- warning text embedded in another line, which must still remain an error

## Testing

- `dotnet test tests/Microsoft.DotNet.XHarness.iOS.Shared.Tests/Microsoft.DotNet.XHarness.iOS.Shared.Tests.csproj`

> [!NOTE]
> This PR description was created by GitHub Copilot.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.iOS.Shared/Logging/DeviceLogCapturer.cs
+++ b/src/Microsoft.DotNet.XHarness.iOS.Shared/Logging/DeviceLogCapturer.cs
@@ -19,6 +19,8 @@ public interface IDeviceLogCapturer : IDisposable
 
 public class DeviceLogCapturer : IDeviceLogCapturer
 {
+    internal const string WallClockAdjustmentWarning = "Wall Clock adjustment detected - results might be strange while using --end";
+
     private readonly ILog _mainLog;
     private readonly ILog _deviceLog;
     private readonly string _deviceUdid;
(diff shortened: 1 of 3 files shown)


vitek-karas:
This is relation to https://github.com/dotnet/runtime/issues/128707 - that KBE seems to be effectively hiding real test failures or infra issues during test execution itself (notice in the log that mlaunch failed before we tried to read the logs).

According to documentation this message is effectively a warning that the retrieved logs could be "corrupted" (or rather incomplete). So far I didn't find a good way to workaround if this happens, but I didn't spend too much time on that one. Regardless we should not be reporting this as an "error" since it will easily catch attention of both agents and humans and will lead to wrong assumptions about the nature of the failure.

matouskozak (approved):
LGTM!

src/Microsoft.DotNet.XHarness.iOS.Shared/Logging/DeviceLogCapturer.cs
```diff
@@ -19,6 +19,8 @@ public interface IDeviceLogCapturer : IDisposable
 
 public class DeviceLogCapturer : IDeviceLogCapturer
 {
+    internal const string WallClockAdjustmentWarning = "Wall Clock adjustment detected - results might be strange while using --end";
```
kotlarmilos, line 23: Nit: Since we will likely add more known warnings over time, I suggest implementing KnownWarningDiagnostics as HashSet<string> instead of having string.Equals calls in IsKnownWarningDiagnostic

```csharp
private static readonly HashSet<string> KnownWarningDiagnostics = new(StringComparer.Ordinal)
{
    WallClockAdjustmentWarning,
};

internal static bool IsKnownWarningDiagnostic(string diagnostic)
    => KnownWarningDiagnostics.Contains(diagnostic);
```

kotlarmilos (approved):

src/Microsoft.DotNet.XHarness.iOS.Shared/Logging/DeviceLogCapturer.cs
vitek-karas: Applied, thanks. `IsKnownWarningDiagnostic` now checks a `KnownWarningDiagnostics` `HashSet<string>` so adding future known warnings stays centralized.

kotlarmilos (approved):

PR #1636, opened 2026-07-16
Improve Android emulator recovery in Helix tests

### LINKED ISSUE #1623: E2E Android - Manual Commands leg fails: emulator stays offline
## Summary

The **E2E Android - Manual Commands** Helix leg (`Build_Debug`) is consistently failing in the PR pipeline. The failure is unrelated to PRs changes, the Android emulator never comes online, so the XHarness manual-commands run times out with no results.

The sibling legs E2E Android - Devices and E2E Android - Simulators pass on the same build, which points at this specific Manual Commands leg / emulator queue rather than a product regression.

## Tasks
- Investigate why `emulator-5554` reports `offline` for the full retry window on the `ubuntu.2204.amd64.android.29.open` queue / Manual Commands leg.

vitek-karas:
## Summary

- Continue emulator recovery when `adb devices` still reports the emulator as `offline`
- Apply emulator recovery to the `xharness android device` command
- Give the Android Manual Commands work item the same 30-minute timeout used by runtime Android work items
- Add regression coverage for an emulator that remains offline after the ADB server reset

## Rationale

The Manual Commands work item could spend most of its 12-minute timeout polling an offline emulator. Helix then terminated the work item with exit code 143 before XHarness could complete recovery or return `DEVICE_NOT_FOUND`, preventing Arcade's diagnostics processor from requesting its normal retry and machine reboot.

The updated recovery path only considers an ADB `device` state successful. An emulator that remains `offline` now proceeds to the emulator restart step, and the longer timeout allows XHarness to finish recovery or return a classifiable exit code.

Fixes #1623

## Testing

- `Build.cmd -configuration Debug -test -projects tests\Microsoft.DotNet.XHarness.Android.Tests\Microsoft.DotNet.XHarness.Android.Tests.csproj`
- `Build.cmd -configuration Debug -projects src\Microsoft.DotNet.XHarness.CLI\Microsoft.DotNet.XHarness.CLI.csproj`
- Validated the Manual Commands shell script with `bash -n`

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
+++ b/src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
@@ -444,14 +444,13 @@ public class AdbRunner
             _log.LogWarning($"Error restarting ADB server during recovery: {e.Message}");
         }
 
-        // Step 3: Re-check for devices after ADB reset - if a device reappeared, we are done
+        // Step 3: Re-check for devices after ADB reset - if an online device reappeared, we are done
         var recheckResult = RunAdbCommand(new[] { "devices", "-l" }, TimeSpan.FromSeconds(30));
         _log.LogInformation($"Devices after ADB server reset:{Environment.NewLine}{recheckResult.StandardOutput}");
 
-        var recheckLines = recheckResult.StandardOutput.Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries);
-        if (recheckLines.Length >= 2)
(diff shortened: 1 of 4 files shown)


vitek-karas:
The real fix for the original failure is https://github.com/dotnet/xharness/pull/1637, but this one is still good to have as it makes the infra more robust. (on its own this actually fixes the problem, but improving the robustness, 1637 addresses the root cause)

kotlarmilos (approved):

PR #1637, opened 2026-07-16
Avoid retrying non-transient Android property failures

vitek-karas:
## Summary

- retry Android device property reads only for transient ADB connection failures
- avoid five-minute retries when optional CPU-frequency files do not exist
- add regression coverage and enable debug output for the affected Manual Commands leg

## Root cause

The repeated ten-minute delays were not emulator boot waits. XHarness tried these optional CPU-frequency paths while collecting device environment information:

- `/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq`
- `/sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq`

Both paths are absent on the Helix API 29 emulator. Each `adb shell cat` therefore returned exit code 1 with `No such file or directory`, but XHarness treated every nonzero property command as an offline device and retried it for five minutes. The generated `android test` path collects the environment once; separate `android install` and `android run` commands collect it twice, producing the consistent twenty-minute delay.

## CI confirmation

[Build 1513688](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1513688) confirmed the diagnosis:

- the Manual Commands stage completed in **2m 52s**, down from **39m 37s** in the comparison run
- both device boot checks completed in **0 seconds**
- both missing CPU-frequency paths failed once and immediately fell back to `/proc/cpuinfo`
- install completed in about **3 seconds** and run preparation in about **2 seconds**
- the affected [Helix work item](https://helix.dot.net/api/2019-06-17/jobs/c0996ada-1a18-439d-ba86-8490d8636e04/workitems/System.Numerics.Vectors.Tests-x86/console) passed within the original 12-minute timeout

The overall build failure is unrelated: the Apple Simulator management leg reached its three-hour agent limit.

Related to #1623. Independent of #1636.

## Testing

- `Microsoft.DotNet.XHarness.Android.Tests`: 36 passed
- XHarness CLI build: succeeded for all target frameworks


src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
```diff
@@ -1304,6 +1313,19 @@ private IReadOnlyCollection<AndroidDevice> GetDevices(params AdbProperty[] prope
...
+    private static bool IsTransientDeviceConnectionFailure(ProcessExecutionResults result)
+    {
+        if (result.TimedOut)
+        {
+            return true;
+        }
+
+        return result.StandardError.Contains("device offline", StringComparison.OrdinalIgnoreCase)
+            || result.StandardError.Contains("no devices/emulators found", StringComparison.OrdinalIgnoreCase)
+            || (result.StandardError.Contains("device", StringComparison.OrdinalIgnoreCase)
```
kotlarmilos: This does not match "device not found" as a single phrase but any combination of words, which could be too broad

kotlarmilos (approved):

src/Microsoft.DotNet.XHarness.Android/AdbRunner.cs
vitek-karas: Fixed in `42d67a0` by requiring either the exact `device not found` phrase or ADB's quoted `device '<serial>' not found` form, instead of matching `device` and `not found` independently.

kotlarmilos (approved):
Thanks!

PR #1606, opened 2026-06-01
Add agentic workflows for runtime failure observer and Maestro auto-merge

kotlarmilos:
## Description

Adds two gh-aw agentic workflows for xharness.

`runtime-failure-observer` runs every 12h. Scans nine `dotnet/runtime` pipelines that drive xharness (Apple, Android, WASM, NativeAOT outer loop), walks AzDO timelines and Helix work items, filters to the eight xharness exit codes from `ExitCode.cs` that map to small fixes, and opens a draft PR on xharness with a bounded improvement (improved diagnostic, narrow retry, exit-code propagation). When a complete small-bounds fix does not fit, it opens a best-effort draft PR (partial fix plus analysis) or a diagnosability draft PR that adds logging, falling back to an issue only when neither fits. Caps at 2 PRs and 2 issues per run.

`maestro-auto-merge` is a reviewer plus a finalize companion that auto-merges Maestro dependency PRs after CI passes and the configured wait elapses. The reviewer (gh-aw) gates on author, diff scope, channel coherency, nuget.org availability, no-downgrade, CI, and version magnitude, then labels safe PRs `auto-merge-candidate`. The finalize job (plain Actions, scheduled) squash-merges a labeled PR once required CI is green and the PR has been idle for 4 hours, pinning the merge to the validated head commit.

The `pr-failure-scan` skill under `.github/skills/` remains available for on-demand PR CI triage via the Copilot CLI.

PR #1673, opened 2026-08-14
Update gh-aw to v0.86.2

vitek-karas:
Updates the runtime failure observer's generated workflow artifacts to gh-aw v0.86.2.

- Regenerates the workflow lock with v0.86.2.
- Refreshes the action and container pins emitted by the compiler.

Validated with `gh-aw v0.86.2 compile runtime-failure-observer.agent.md --schedule-seed dotnet/xharness --no-check-update --no-emit --validate`.


.gitattributes
```diff
@@ -1,4 +1,4 @@
 # Apply LF to shell scripts automatically
 *.sh text eol=lf
 
-.github/workflows/*.lock.yml linguist-generated=true merge=ours
\ No newline at end of file
+.github/workflows/*.lock.yml linguist-generated=true
```
davidnguyen-tech, line 4: What was the original purpose of `merge=ours`?

.github/workflows/runtime-failure-observer.agent.lock.yml
```diff
@@ -135,27 +136,28 @@ jobs:
...
       - name: Generate agentic run info
         id: generate_aw_info
         env:
           GH_AW_INFO_ENGINE_ID: "copilot"
           GH_AW_INFO_ENGINE_NAME: "GitHub Copilot CLI"
-          GH_AW_INFO_MODEL: ${{ vars.GH_AW_MODEL_AGENT_COPILOT || vars.GH_AW_DEFAULT_MODEL_COPILOT || 'claude-sonnet-4.6' }}
-          GH_AW_INFO_VERSION: "1.0.70"
-          GH_AW_INFO_AGENT_VERSION: "1.0.70"
-          GH_AW_INFO_CLI_VERSION: "v0.82.9"
+          GH_AW_INFO_MODEL: ${{ vars.GH_AW_MODEL_AGENT_COPILOT || vars.GH_AW_DEFAULT_MODEL_COPILOT || 'auto' }}
```
davidnguyen-tech: The default model changed from `claude-sonnet-4.6` to `auto`. I'm aware it's autogenerated, but still wanted to ask - do we want to keep this?

davidnguyen-tech (approved):

.gitattributes
vitek-karas: AI Response: It was intended to avoid manual merge conflicts in compiler-generated `.lock.yml` files: an `ours` merge driver would retain the version already on the merge target instead of combining independently regenerated locks. It was introduced with the first gh-aw workflow in #1606.

It is not used at workflow runtime, and this repo does not define an `ours` merge driver. gh-aw v0.85.4 also stopped emitting the attribute, so this update keeps `linguist-generated=true` while removing the obsolete merge directive.

davidnguyen-tech (approved):
LGTM, thanks! Just curious - why `gpt-5.6-terra` in the end?

vitek-karas:
> why gpt-5.6-terra in the end?

Prefer GPT models (due to cost) - I don't think we need Sol for this (I've been using Terra locally and it works pretty well on most things), Luna doesn't feel strong enough (locally it gives up too easily).

PR #1696, opened 2026-08-24
Improve diagnostics for Apple apps that exit before test startup

### LINKED ISSUE #1685: Improve diagnostics for Apple apps that exit before test startup
## Problem

Apple physical-device runs can report `APP_CRASH` even when the available evidence cannot establish that the launched app crashed. This makes recurring failures difficult to triage and can associate a run with an unrelated old `.ips` report.

Example: dotnet/runtime dependency-flow PR [dotnet/runtime#132383](https://github.com/dotnet/runtime/pull/132383) updated XHarness to `11.0.0-prerelease.26413.1`. Its iOS and tvOS `System.Runtime.Tests` runs installed and launched the app, but `devicectl` reported that it exited with code 1 before emitting test output or creating `test-results.xml`:

- [iOS console](https://helix.dot.net/api/2019-06-17/jobs/e47bf191-9c81-439f-a73f-9b60a7433311/workitems/System.Runtime.Tests/console): app launched, then exited 1 in about 0.30 seconds.
- [tvOS console](https://helix.dot.net/api/2019-06-17/jobs/bb25d003-0f81-4d15-add7-9bf78e17ef51/workitems/System.Runtime.Tests/console): app launched, then exited 1 in about 0.71 seconds.

Both runs then failed all four result-file copy attempts because the app had not created the file. XHarness reported `APP_CRASH` (exit 80), but it did not find a crash report created for this run. The reports it selected were stale:

- The iOS report was from July 27, while the run was on August 17.
- The tvOS report was from July 20 and belonged to a different bundle identifier.

The same pre-test exit pattern occurred in earlier `release/8.0-staging` builds, including [1548187](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1548187), [1549804](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1549804), and [1549996](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1549996). This is therefore not caused by the dependency update itself.

[XHarness #1630](https://github.com/dotnet/xharness/pull/1630) correctly changed the unrelated wall-clock log-reader diagnostic into a warning. These runs demonstrate the remaining gap: distinguishing a confirmed app crash from an app that exits before the test protocol begins when no contemporaneous crash report is available.

## Proposed improvement

1. Snapshot available crash reports before launch. After the app exits, only select reports that:
   - were created during the current run window;
   - match the launched bundle identifier; and
   - if available, match the launched process identity.

2. Add structured launch diagnostics to `xharness-result.json`, including launcher command exit code, app exit code, test-start and test-end signal state, expected result-file path, copy-attempt results, and matched crash-report metadata.

3. Add a distinct outcome for an app that exits before the test protocol starts and has no matched, contemporaneous crash report. Avoid labelling this condition `APP_CRASH`; retain `APP_CRASH` for a matched crash report or other confirmed crash evidence.

4. Preserve and upload targeted post-exit diagnostics: the runner's device log, `devicectl` launch output, the crash-report inventory before and after launch, and the selected report metadata when a report is matched.

## Example output

Console output could make the classification explicit:

```text
Apple launch outcome:
  Bundle ID: net.dot.System.Runtime.Tests
  Device: DNCENGOSX118
  Launcher exit code: 0
  App exit code: 1
  Test protocol started: no
  Test result file: /Documents/test-results.xml (not created)
  Crash reports before launch: 12
  Crash reports created during run: 0
  Matching crash report: none

XHarness exit code: 81 (APP_EXITED_BEFORE_TEST_START)
```

And the result artifact could preserve the machine-readable evidence:

```json
{
  "exitCode": 81,
  "exitCodeName": "APP_EXITED_BEFORE_TEST_START",
  "appleLaunch": {
    "bundleId": "net.dot.System.Runtime.Tests",
    "appExitCode": 1,
    "testProtocolStarted": false,
    "testResultFile": {
      "path": "/Documents/test-results.xml",
      "copyAttempts": 4,
      "exists": false
    },
    "crashReport": {
      "reportsBeforeLaunch": 12,
      "reportsCreatedDuringRun": 0,
      "matchedReport": null
    }
  }
}
```

This would make the next occurrence actionable: it would show whether investigation should start with the app/device launch path, a genuine crash report, or result-file transport, without relying on stale device state.

> [!NOTE]
> This issue was drafted with GitHub Copilot.

vitek-karas:
Fixes #1685

## Why

Apple test runs could report `APP_CRASH` even when there was no evidence that the launched app crashed. XHarness selected the latest available `.ips` report, which could be stale, belong to another application, or predate the current run.

This made CI failures misleading and obscured the more actionable condition: the app launched and exited before producing test results, with no matching crash report.

## What changed

- Snapshot crash reports before launch and compare them after exit.
- Consider only reports created during the current run.
- Match reports using the bundle identifier or executable/process name.
- Parse both modern `.ips` JSON and legacy Apple crash reports.
- Remove the unsafe fallback that selected the latest report.
- Preserve crash-report inventories and matched reports as diagnostics.
- Record launcher and application exit codes separately.
- Record whether a test protocol was expected and whether it connected, avoiding the previous ambiguity between listener connectivity and actual test execution.
- Record result-file copy attempts, expected path, end-signal state, and crash-report evidence.
- Add an `appleLaunch` section to `xharness-result.json`.
- Add `APP_EXITED_BEFORE_TEST_START` with exit code `93`; known TCP failures retain `TCP_CONNECTION_FAILED` (`92`).
- Add `TEST_RESULTS_MISSING` with exit code `94` for runs where the protocol connected but no results were produced.
- Reserve `APP_CRASH` for runs with a matching crash report.

The Apple diagnostics model remains in the Apple-specific assembly. Common only exposes a generic mechanism for platforms to append structured result sections, allowing Android and other targets to add their own diagnostics without depending on Apple concepts.

## Why removing `GetCrashReason` is not a loss of functionality

The previous crash-reason path looked as though it extracted an OS termination reason, but it was not connected to the reports collected during a test run:

- `TestReporter` enumerated a private `_crashLogs` collection that was never populated by `CrashSnapshotReporter`; the reporter wrote to a different `Logs` instance.
- Consequently, the `GetCrashReason` loop and its exception callback were not reached for downloaded crash reports.
- `GetCrashReason` expected an older JSON process-list shape (`processes/item/reason`), not the modern `.ips` or legacy crash-report formats handled by the crash reporter.

Removing this plumbing therefore does not remove diagnostics that were working in practice. The new path still inventories, downloads, preserves, matches, and symbolicates reports produced during the run. Extracting termination or exception reasons from actual modern and legacy crash-report fields can be added separately without reviving the disconnected parser.

## Result

Failures now distinguish between:

- a matching Apple crash report;
- a known TCP connection failure;
- an app that exits without test-startup evidence or matching crash evidence; and
- a connected test protocol that never produces a results file.

This prevents stale reports from being attributed to the current run and gives CI triage enough structured evidence to identify the appropriate investigation path.

## Testing

- Apple orchestration and app-launch diagnostics
- Modern and legacy crash-report matching
- Missing, stale, unrelated, and BOM-prefixed crash reports
- Early-exit, missing-results, and confirmed-crash classification
- Protocol-connected crash-report precedence
- TCP failure precedence
- Pre-iOS 18 protocol and iOS 18 container-copy diagnostics
- Android summary compatibility

Repository build completed with 0 warnings and 0 errors. All 64 focused tests pass.

### DIFF
--- a/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
+++ b/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
@@ -6,9 +6,11 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DotNet.XHarness.Common;
+using Microsoft.DotNet.XHarness.Common.Execution;
 using Microsoft.DotNet.XHarness.Common.Logging;
 using Microsoft.DotNet.XHarness.Common.Utilities;
 using Microsoft.DotNet.XHarness.iOS.Shared;
@@ -23,6 +25,7 @@ namespace Microsoft.DotNet.XHarness.Apple;
 public interface IAppTester
 {
     bool ListenerConnected { get; }
+    AppleLaunchDiagnostics? LaunchDiagnostics { get; }
 
     Task<(TestExecutingResult Result, string ResultMessage)> TestApp(
         AppBundleInformation appInformation,
@@ -58,6 +61,15 @@ public interface IAppTester
 /// </summary>
 public class AppTester : AppRunnerBase, IAppTester
 {
+    private static readonly Regex[] s_appExitCodeRegexes =
+    {
+        new(@"The app '[^']+' exited with exit code (?<exitCode>-?\d+)", RegexOptions.Compiled),
+        new(@"The app terminated with the exit code (?<exitCode>-?\d+)\.", RegexOptions.Compiled),
+        new(@"terminated \(with exit code '(?<exitCode>-?\d+)' and/or crashing signal", RegexOptions.Compiled),
+        new(@"DOTNET\.APP_EXIT_CODE:\s*(?<exitCode>-?\d+)\s*$", RegexOptions.Compiled),
+        new(@"Service exited with abnormal code:\s*(?<exitCode>-?\d+)\s*$", RegexOptions.Compiled),
+    };
+
     private readonly IMlaunchProcessManager _processManager;
     private readonly ISimpleListenerFactory _listenerFactory;
     private readonly ICrashSnapshotReporterFactory _snapshotReporterFactory;
@@ -73,6 +85,7 @@ public class AppTester : AppRunnerBase, IAppTester
     /// This is used later to determine if a cause for a failed run is a failing TCP connection.
     /// </summary>
     public bool ListenerConnected { get; private set; }
+    public AppleLaunchDiagnostics? LaunchDiagnostics { get; private set; }
 
     public AppTester(
         IMlaunchProcessManager processManager,
@@ -111,6 +124,8 @@ public class AppTester : AppRunnerBase, IAppTester
         string[]? skippedTestClasses = null,
         CancellationToken cancellationToken = default)
     {
+        LaunchDiagnostics = CreateLaunchDiagnostics(appInformation);
+
         var testLog = _logs.Create($"test-{TestTarget.MacCatalyst.AsString()}-{_helpers.Timestamp}.log", LogType.TestLog.ToString(), timestamp: false);
         var appOutputLog = _logs.Create(appInformation.BundleIdentifier + ".log", LogType.ApplicationLog.ToString(), timestamp: true);
 
@@ -169,6 +184,7 @@ public class AppTester : AppRunnerBase, IAppTester
     {
         var runMode = target.Platform.ToRunMode();
         var isSimulator = target.Platform.IsSimulator();
+        LaunchDiagnostics = CreateLaunchDiagnostics(appInformation);
 
         var testLog = _logs.Create($"test-{target.AsString()}-{_helpers.Timestamp}.log", LogType.TestLog.ToString(), timestamp: false);
 
@@ -187,9 +203,7 @@ public class AppTester : AppRunnerBase, IAppTester
             var deviceListenerPort = deviceListener.InitializeAndGetPort();
             deviceListener.StartAsync();
 
-            using var crashLogs = new Logs(_logs.Directory);
-
-            ICrashSnapshotReporter crashReporter = _snapshotReporterFactory.Create(_mainLog, crashLogs, isDevice: !isSimulator, device.Name);
+            ICrashSnapshotReporter crashReporter = _snapshotReporterFactory.Create(_mainLog, _logs, isDevice: !isSimulator, device.Name, appInformation);
             using ITestReporter testReporter = _testReporterFactory.Create(
                 _mainLog,
                 _mainLog,
@@ -315,7 +329,9 @@ public class AppTester : AppRunnerBase, IAppTester
             }
 
             // Check the final status, copy all the required data
-            return await testReporter.ParseResult();
+            var result = await testReporter.ParseResult();
+            CompleteLaunchDiagnostics(testReporter, crashReporter, deviceListener);
+            return result;
         }
     }
 
@@ -344,6 +360,7 @@ public class AppTester : AppRunnerBase, IAppTester
             waitForExit: true,
             cancellationToken);
 
+        RecordLaunchResult(result, SimulatorApplicationLog);
         await testReporter.CollectSimulatorResult(result);
 
         // On iOS 18 and later, transferring results over a TCP tunnel isn’t supported.
@@ -358,6 +375,7 @@ public class AppTester : AppRunnerBase, IAppTester
                 simulator.UDID,
                 appInformation.BundleIdentifier,
                 deviceListener.TestLog.FullPath);
+            RecordResultFileCopy(resultFileHandler, runMode, resultsCopied);
 
             // If results weren't copied, it likely means the app crashed before tests could run
             // Try to retrieve the crash report, but only if the test run didn't already complete.
@@ -371,13 +389,7 @@ public class AppTester : AppRunnerBase, IAppTester
                 }
                 else
                 {
-                    _mainLog.WriteLine("Test results file not found, app may have crashed before tests started.");
-                    await resultFileHandler.CopyCrashReportAsync(
-                        simulator.UDID,
-                        simulator.Name,
-                        appInformation,
-                        _mainLog,
-                        isSimulator: true);
+                    _mainLog.WriteLine("Test results file not found. Checking the pre-launch crash report snapshot for a matching report.");
                 }
             }
 
@@ -448,6 +460,7 @@ public class AppTester : AppRunnerBase, IAppTester
                 envVars,
                 cancellationToken: cancellationToken));
 
+            RecordLaunchResult(result, appOutputLog as IReadableLog);
             await testReporter.CollectDeviceResult(result);
         }
         finally
@@ -480,6 +493,7 @@ public class AppTester : AppRunnerBase, IAppTester
                 device.UDID,
                 appInformation.BundleIdentifier,
                 deviceListener.TestLog.FullPath);
+            RecordResultFileCopy(resultFileHandler, runMode, resultsCopied);
 
             // If results weren't copied, it likely means the app crashed before tests could run
             // Try to retrieve the crash report, but only if the test run didn't already complete.
@@ -493,13 +507,7 @@ public class AppTester : AppRunnerBase, IAppTester
                 }
                 else
                 {
-                    _mainLog.WriteLine("Test results file not found, app may have crashed before tests started.");
-                    await resultFileHandler.CopyCrashReportAsync(
-                        device.UDID,
-                        device.Name,
-                        appInformation,
-                        _mainLog,
-                        isSimulator: false);
+                    _mainLog.WriteLine("Test results file not found. Checking the pre-launch crash report snapshot for a matching report.");
                 }
             }
 
@@ -539,9 +547,7 @@ public class AppTester : AppRunnerBase, IAppTester
         deviceListener.StartAsync();
         var (enableCoverage, coverageFileName) = GetCoverageSettings(extraEnvVariables);
 
-        using var crashLogs = new Logs(_logs.Directory);
-
-        ICrashSnapshotReporter crashReporter = _snapshotReporterFactory.Create(_mainLog, crashLogs, isDevice: false, null);
+        ICrashSnapshotReporter crashReporter = _snapshotReporterFactory.Create(_mainLog, _logs, isDevice: false, null, appInformation);
         IResultFileHandler resultFileHandler = new ResultFileHandler(_processManager, _mainLog);
         using ITestReporter testReporter = _testReporterFactory.Create(
             _mainLog,
@@ -602,6 +608,7 @@ public class AppTester : AppRunnerBase, IAppTester
             await crashReporter.StartCaptureAsync();
 
             var result = await RunMacCatalystApp(appInformation, appOutputLog, timeout, waitForExit: true, extraAppArguments, envVariables, combinedCancellationToken.Token);
+            RecordLaunchResult(result, appOutputLog as IReadableLog);
             await testReporter.CollectSimulatorResult(result);
 
             if (enableCoverage)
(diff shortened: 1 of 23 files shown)


rolfbjarne (approved):
LGTM

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
```diff
@@ -626,7 +633,113 @@ await resultFileHandler.CopyCrashReportAsync(
...
+        ISimpleListener deviceListener)
+    {
+        if (LaunchDiagnostics is null)
+        {
+            return;
+        }
+
+        LaunchDiagnostics.TestProtocolStarted = LaunchDiagnostics.TestResultFile.CopyAttempts > 0
+            ? null
+            : testReporter.TestExecutionStarted;
```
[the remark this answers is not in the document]
vitek-karas: Fixed in `417acde`. A known `true` value now takes precedence; the field is only `null` when no protocol start was observed and direct result-copy attempts mean the TCP signal is not applicable.

src/Microsoft.DotNet.XHarness.iOS.Shared/CrashSnapshotReporter.cs
```diff
@@ -166,12 +176,176 @@ private async Task<IFileBackedLog> ProcessCrash(string crashFile)
...
+            string processName = null;
+            string bugType = null;
+            string timestamp = null;
+            int? processId = null;
+
+            if (!string.IsNullOrWhiteSpace(firstLine) &&
+                firstLine.TrimStart('\uFEFF', ' ', '\t').StartsWith("{", StringComparison.Ordinal))
+            {
+                using (JsonDocument header = JsonDocument.Parse(firstLine))
+                {
```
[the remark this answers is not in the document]
vitek-karas: Fixed in `417acde`. The header line is normalized once and the BOM-trimmed value is passed to `JsonDocument.Parse`. I also added coverage for a BOM-prefixed `.ips` header.

tests/Microsoft.DotNet.XHarness.iOS.Shared.Tests/TestReporterTests.cs
```diff
@@ -505,4 +507,48 @@ public async Task ParseResult_WhenTestCompletedButResultsUnavailable_ReturnsSucc
         // Crash reporter should be called with 0 delay since tests succeeded
         _crashReporter.Verify(c => c.EndCaptureAsync(It.Is<TimeSpan>(t => t.TotalSeconds == 0)), Times.Once);
     }
+
+    [Fact]
+    public async Task ParseResult_WhenAppExitsBeforeProtocolWithoutCrashReport_ReturnsEarlyExit()
```
davidnguyen-tech: Why do we call it `BeforeProtocol`?

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
```diff
@@ -60,8 +56,6 @@ public class TestReporter : ITestReporter
     /// <summary>
     /// Callback needed for the Xamarin.Xharness project that does extra logging in case of a crash.
     /// </summary>
```
davidnguyen-tech: This comment belonged to the now-removed `private readonly ExceptionLogger? _exceptionLogger;`

I suppose we can remove this comment, too?

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
davidnguyen-tech: This should be redundant now (see https://github.com/dotnet/xharness/pull/1696/changes#r3852291826)

davidnguyen-tech: This should be redundant now (see https://github.com/dotnet/xharness/pull/1696/changes#r3852291826)

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
```diff
@@ -132,7 +124,12 @@ private async Task<int> GetPidFromRunLog()
     {
         int pid = -1;
 
-        using var reader = _runLog.GetReader(); // diposed at the end of the method, which is what we want
+        using var reader = _runLog.GetReader();
+        if (reader is null)
```
davidnguyen-tech: Can the reader ever be `null`? If so, I think we should update the `GetReader` definition. Right now, it is:

```cs
StreamReader GetReader();
```

davidnguyen-tech: Have we considered setting some flags in the if-branch below? E.g. `_launchFailure` when `reader is null`

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
```diff
@@ -632,13 +582,32 @@ private async Task GenerateXmlFailures(string failure, bool crashed, string? cra
...
+            AppleCrashReportDiagnostics? crashDiagnostics = _crashReporter.CaptureDiagnostics;
+            if (!TestExecutionStarted && crashDiagnostics?.MatchedReport is null)
+            {
+                appExitedBeforeTestStart = true;
+                result.ResultMessage = "App exited before the test protocol started and no matching crash report was found";
+                _mainLog.WriteLine(result.ResultMessage);
+            }
+            else
+            {
+                crashed = true;
```
davidnguyen-tech: Why does "test execution started" && "no matching crash report" imply that the test crashed?

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
```diff
@@ -76,7 +70,7 @@ public class TestReporter : ITestReporter
 
     public bool ResultsUseXml => _xmlJargon != XmlResultJargon.Missing;
 
-    private bool TestExecutionStarted => _listener.ConnectedTask.IsCompletedSuccessfully && _listener.ConnectedTask.Result;
+    public bool TestExecutionStarted => _listener.ConnectedTask.IsCompletedSuccessfully && _listener.ConnectedTask.Result;
```
davidnguyen-tech: I know this wasn't introduced in this PR, but is it possible that the `_listener` connects successfully without the test actually starting?

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
```diff
@@ -667,63 +640,21 @@ private async Task GenerateXmlFailures(string failure, bool crashed, string? cra
...
-            }
-            else if (_launchFailure)
+            if (_launchFailure)
             {
-                // same as with a crash
                 result.ResultMessage = $"Launch failure";
             }
 
-            await GenerateXmlFailures(result.ResultMessage, crashed, crashReason);
+            await GenerateXmlFailures(result.ResultMessage, crashed, crashReason: null);
```
davidnguyen-tech: It looks like this is the only callsite of the method.

Are we keeping the `crashReason` argument just in case we want to use it in the future?

src/Microsoft.DotNet.XHarness.Apple/Orchestration/TestOrchestrator.cs
```diff
@@ -337,7 +339,7 @@ ExitCode LogProblem(string message, ExitCode defaultExitCode)
 
             // TCP errors are encounter all the time but they are not always the cause of the failure
             // If the app crashed, TCP_CONNECTION_FAILED and there was not other exit code we will return TCP_CONNECTION_FAILED
-            if (defaultExitCode == ExitCode.APP_CRASH && tcpErrorFound)
+            if ((defaultExitCode == ExitCode.APP_CRASH || defaultExitCode == ExitCode.APP_EXITED_BEFORE_TEST_START) && tcpErrorFound)
```
davidnguyen-tech: When there is a genuinely matched crash report, should a TCP log line still override `APP_CRASH`?

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
```diff
@@ -371,13 +389,7 @@ private async Task RunSimulatorTests(
...
                 else
                 {
-                    _mainLog.WriteLine("Test results file not found, app may have crashed before tests started.");
-                    await resultFileHandler.CopyCrashReportAsync(
-                        simulator.UDID,
-                        simulator.Name,
-                        appInformation,
-                        _mainLog,
-                        isSimulator: true);
+                    _mainLog.WriteLine("Test results file not found. Checking the pre-launch crash report snapshot for a matching report.");
```
davidnguyen-tech: Where does the matching for a crash report happen? The log message makes it sound like it should be happening somewhere around here.

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
```diff
@@ -493,13 +507,7 @@ private async Task RunDeviceTests(
...
                 else
                 {
-                    _mainLog.WriteLine("Test results file not found, app may have crashed before tests started.");
-                    await resultFileHandler.CopyCrashReportAsync(
-                        device.UDID,
-                        device.Name,
-                        appInformation,
-                        _mainLog,
-                        isSimulator: false);
+                    _mainLog.WriteLine("Test results file not found. Checking the pre-launch crash report snapshot for a matching report.");
```
davidnguyen-tech: Where does the matching for a crash report happen? The log message makes it sound like it should be happening somewhere around here.

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
```diff
@@ -626,7 +633,113 @@ await resultFileHandler.CopyCrashReportAsync(
...
+
+    private static AppleLaunchDiagnostics CreateLaunchDiagnostics(AppBundleInformation appInformation)
+        => new()
+        {
+            BundleId = appInformation.BundleIdentifier,
+        };
+
+    private void RecordLaunchResult(ProcessExecutionResult result, IReadableLog? appOutputLog)
+    {
+        if (LaunchDiagnostics is null)
```
davidnguyen-tech: Why do we silently return in this case? (repeated in methods below too)

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
```diff
@@ -626,7 +633,113 @@ await resultFileHandler.CopyCrashReportAsync(
...
+        ISimpleListener deviceListener)
+    {
+        if (LaunchDiagnostics is null)
+        {
+            return;
+        }
+
+        LaunchDiagnostics.TestProtocolStarted = testReporter.TestExecutionStarted
+            ? true
+            : LaunchDiagnostics.TestResultFile.CopyAttempts > 0 ? null : false;
```
davidnguyen-tech: What's the semantics of `TestProtocolStarted == null`?

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
vitek-karas: Removed the null check in `183ad03`. `GetReader()` has a non-null contract, so tests now provide a valid reader rather than adding an impossible production state or setting launch flags for it.

tests/Microsoft.DotNet.XHarness.iOS.Shared.Tests/TestReporterTests.cs
vitek-karas: Fixed in `183ad03`. The test is now named `ParseResult_WhenAppExitsWithoutTestStart_ReturnsEarlyExit`, which describes the evidence rather than conflating listener connectivity with actual execution.

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
vitek-karas: Updated the device-path message in `183ad03` to describe the actual lifecycle: matching happens later while finalizing the run result in `CrashSnapshotReporter.EndCaptureAsync()`.

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
vitek-karas: Replaced the silent returns in `183ad03` with `GetLaunchDiagnostics()`, which throws an explicit `InvalidOperationException` if the initialization invariant is violated.

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
vitek-karas: Fixed in `183ad03`. Missing results are classified as `Crashed` only when a crash report matches. If the protocol connected without producing results, the run now returns `TestResultsMissing` / `TEST_RESULTS_MISSING` (94).

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
vitek-karas: Removed the unused `exceptionLogger` parameter end-to-end from `TestReporter`, `ITestReporterFactory`, the implementation, and both call sites in `183ad03`.

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
vitek-karas: Removed the orphaned comment in `183ad03`.

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
vitek-karas: Updated the message in `183ad03` to state that new reports are matched when the run result is finalized. Matching occurs in `CrashSnapshotReporter.EndCaptureAsync()` through `MatchCrashReport()`.

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
vitek-karas: Removed the unused `ExceptionLogger` alias in `183ad03`.

src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs
vitek-karas: Removed the nullable state in `183ad03`. Diagnostics now expose `TestProtocolExpected` and `TestProtocolConnected` as separate booleans, so iOS 18 container-copy runs are explicit without a tri-state value.

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
vitek-karas: Removed the always-null `crashReason` argument and its unreachable branch in `183ad03`.

src/Microsoft.DotNet.XHarness.iOS.Shared/TestReporter.cs
vitek-karas: Yes, connectivity alone does not prove test execution. In `183ad03` the property is renamed to `TestProtocolConnected`, and the structured diagnostics now separately report whether the protocol was expected and whether it connected.

davidnguyen-tech (approved):
LGTM, thanks a lot!

PR #1722, opened 2026-09-11
Use GPT-5.6 Luna for PAT validation

vitek-karas:
## Summary

- Pin PAT validation to `gpt-5.6-luna` with low reasoning effort.
- Keep the validation prompt lightweight while using a model that supports the configured effort level.

## Validation

- `git diff --check`
- YAML parse check passed.
- Live PAT validation was not run locally because the workflow's Actions credentials are not available in the local environment.


vitek-karas:
The review comment is about the fact that the agents we use typically run on Terra but this is testing Luna - in theory the PAT could be restricted to only certain models which would then cause breaks. This is not important right now, as we don't restrict the PATs this way.

davidnguyen-tech (approved):

